PR #40, opened 2014-12-03
catchable/capture for name

tpolecat:
It's not clear that this is useful at all. Probably will let this sit here for a while.


tpolecat:
I think this is the closest we can get to an `Id` target monad so I'm going to go ahead and call it good, even if nobody ever uses it. I'll add some tests though.

Issue #41, opened 2014-12-22
HikariTransactor


tpolecat:

prange:
Any quick workaround for using HickaryCP?

tpolecat:
Hi @prange, I want to build an algebra and free monad so you can do all the configuration, etc., in an API that fits with the rest of doobie, but yeah for now just hack something up based on that example.

FYI at this point we're really close to v0.2 so your best bet is to build and publish-local and use the snapshot.

prange:
Yes, i have cloned the 0.2-SNAPSHOT and will use that for the moment...

PR #51, opened 2015-01-18
Hikari transactor and example

tpolecat:
This implementation is slightly weak because it gives you direct access to the `HikariDataSource` if you ask to configure, but it's good enough for now.


tpolecat:
This fixes #41

PR #72, opened 2015-02-02
Use 'data type' instead of 'monad'.

gseitz:
Based on [IO Monad considered harmful](http://blog.jle.im/entry/io-monad-considered-harmful) I find it more appropriate to talk about data types and only refer to the the monad in the context of the monadic properties of these data types (eg. when `flatMap`, for-comprehension, etc. come into play).


tpolecat:
Yes, much better.

PR #95, opened 2015-02-15
PostGIS support

### LINKED ISSUE #82: Add PostGIS support
Should be easy, just need to add the dependency and mappings in `pgtypes.scala`.

/cc @aloiscochard

tpolecat:
This adds support for the main PostGIS types:
- `PGgeometry`
- `PGbox2d`
- `PGbox3d`

As well as the following abstract and fine-grained types carried by `PGgeometry`:
- `Geometry`
- `ComposedGeom`
- `GeometryCollection`
- `MultiLineString`
- `MultiPolygon`
- `PointComposedGeom`
- `LineString`
- `MultiPoint`
- `Polygon`
- `Point`

fixes #82 

cc @aloiscochard @icassina

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ addons:
   postgresql: "9.3"
 
 before_script:
+  - psql -c "create extension postgis" -U postgres
   - psql -c 'create database world;' -U postgres
   - psql -c '\i world.sql' -d world -U postgres
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,9 @@ addons:
   postgresql: "9.3"
 
 before_script:
-  - psql -c "create extension postgis" -U postgres
   - psql -c 'create database world;' -U postgres
   - psql -c '\i world.sql' -d world -U postgres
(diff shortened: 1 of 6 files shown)


tpolecat:
Note that these are the only types supported by the `postgis-jdbc` extension. Seems like it's kind of minimal.

aloiscochard:
@tpolecat wow, that's awesome :-)

I think there is all the types we need, I should be able to finish my current task this week and move on helping @Icassina, and hopefully one of us will try to migrate or at least just do a few tests with that!

Thanks for all

tpolecat:
Cool, I will merge this later this week after I write a bit of documentation. You can publish-local for testing. I hope to have 0.2.1 out around 5-Mar.

tpolecat:
Also I'm pleased that it took very little code to make this work: https://github.com/tpolecat/doobie/blob/postgis/contrib/postgresql/src/main/scala/doobie/contrib/postgresql/pgtypes.scala#L120-L143

tpolecat:
Will leave #82 open pending doc.

Issue #75, opened 2015-02-02
Make HList work for params/rows


tpolecat:
I expected to get this for free with `ProductTypeclass` but it doesn't seem like I do. Probably a simple fix.

tpolecat:
Miles says upgrading to 2.1.0 will make this easier, so linking with #88 and moving back to 0.3.0.

tpolecat:
Open pending doc.

Issue #99, opened 2015-02-17
Support for query fragments


tpolecat:
@wedens suggests concatenable query fragments that build up a set of parameters along with the sql string. Need `HList` support to make this convenient.

wedens:
my proposed data structure is something like this:
```scala
case class QueryFragment[L <: HList: Composite](query: Cord, params: L) {
  def ++[P <: HList, POut <: HList](other: QueryFragment[P])
    (implicit p: Prepend.Aux[L, P, POut], mtout: Composite[POut]): QueryFragment[POut] =
      QueryFragment(query ++ other.query, params ++ other.params)
}
```

tpolecat:
I like it.

kryptt:
I've been running into some issues with this idea... specifically, I can't make a semigroup out of this, since what we want is precisely to be able to append arbitrary QueryFragments together...

tpolecat:
It can never be a semigroup because the type argument changes, but it should definitely be possible to append fragments (that's the whole point). Sorry I haven't had time to focus on this; I'm working on some less interesting things.

PR #100, opened 2015-02-17
Composite instances for HList and HNil

wedens:
small step towards #99 :) This will make possible to derive Composite for HList

### DIFF
--- a/core/src/main/scala/doobie/util/composite.scala
+++ b/core/src/main/scala/doobie/util/composite.scala
@@ -18,13 +18,13 @@ import shapeless._
 
 import java.sql.ParameterMetaData
 
-/** 
+/**
  * Module defining a typeclass for composite database types (those that can map to multiple columns).
  */
 object composite {
 
   @implicitNotFound("Could not find or construct Composite[${A}].")
-  trait Composite[A] { c =>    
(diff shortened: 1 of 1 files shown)


wedens:
also closes #75 even without shapeless 2.1.0

tpolecat:
Oh cool, I'll check it out tonight.

tpolecat:
Hm I can't get this to work. What do I need to do?

```scala
scala> Composite[(Int,Int)]
res12: doobie.util.composite.Composite[(Int, Int)] = doobie.util.composite$LowerPriorityComposite$$anon$4$$anon$5@5a1166b2

scala> Composite[Int :: Int :: HNil]
<console>:17: error: Could not find or construct Composite[shapeless.::[Int,shapeless.::[Int,shapeless.HNil]]].
              Composite[Int :: Int :: HNil]
                       ^
```

wedens:
Some conflict in implicit resolution. HList Composite should be higher priority (it's really weird as I have almost the sami hierarchy in my playground and it resolves without any ambiguity)

tpolecat:
I'll look at it again when I'm more rested but I think we're good. :+1:  Thanks!

PR #102, opened 2015-02-20
enum support

### LINKED ISSUE #96: Add support for using Postgres enum types
This is simply `Meta[String].nxmap(...)`. Add an example in the doc.

Recommended in #94 

**EDIT** actually it's much more complicated than just invariant mapping of `String`; it's platform-specific and very twitchy.

tpolecat:
Support for constructing `Meta` instances for Scala and Java enumerated types.

fixes #96

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ before_script:
   - psql -c 'create database world;' -U postgres
   - psql -c '\i world.sql' -d world -U postgres
   - psql -d world -c "create extension postgis" -U postgres
+  - psql -d world -c "create type myenum as enum ('foo', 'bar')" -U postgres
 
 notifications:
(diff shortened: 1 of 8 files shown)


tpolecat:
Hm guess I should add a test innit.

tpolecat:
:-1: turns out you need `setObject` with a `PGobject` to set, and there seems to be no way at all to set `NULL`. Investigating...

tpolecat:
Ok, removed the generic enum stuff in favor of something Postgres-specific. It will do for now. Of course I broke the tests. :-1:

tpolecat:
Ok well that was a bit of a slog. Need to leave #96 open for documentation.

tpolecat:
12 commits later...

PR #104, opened 2015-02-24
DataSourceTransactor

### LINKED ISSUE #101: Transactor for existing DataSource
It will be useful when some pool other than hikari is used or DataSource already exists (e.g. when using doobie with plain jdbc).

tpolecat:
fixes #101


tpolecat:
Need to think about this. is there any reason for `dataSource` to be `M[DataSource]`? Doesn't seem like it.

PR #108, opened 2015-02-28
remove 22-param limit

tpolecat:
THIS IS RIDICULOUS

I'll ask @milessabin if I'm screwed or not, but I think I am. The invocations come from the string interpolator rewrite `sql"... $a ... $b ... "` to `new SqlInterpolator(ctx).sql(a, b, ...)` so I fear shapeless can't save me. Macro time maybe? Or call it good since fifty parameters should be plenty for anyone, heh-heh.


milessabin:
You are somewhat screwed.

Interpolated strings are rewritten syntactically so, in principle, you could rely on auto-tupling to handle argument lists of up to 22 elements with a single method (see [here](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/hlists.scala#L65) for an example). But there are no tuple types beyond `Tuple22` so this mechanism runs out of steam at that point.

To go beyond that you'll need a method which takes an `Any*` argument with a macro implementation which pulls out the actual argument types to construct the `HList` you want ... the macro implementation will take a `Tree*` argument list, and each `Tree` will have useful static type. I don't have an example of exactly this, but the shapeless `Record` constructor [uses related techniques](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/records.scala#L60).

Give me a shout if you need a hand.

tpolecat:
Looks fine and examples run, but should review book output as a sanity check. We'll call this good for 0.2.1 and revisit with #109, in which @tpolecat will learn how to macro. Note the date and time.

PR #120, opened 2015-03-06
remove length from composite. make update/set/get stateful computations

### LINKED ISSUE #118: Remove length from composite
Refactor composite `get/set/update` to be stateful computations that pass an offset rather than computing length statically. This will make composite coproducts possible, among other things.

wedens:
fixes #118

### DIFF
--- a/core/src/main/scala/doobie/hi/preparedstatement.scala
+++ b/core/src/main/scala/doobie/hi/preparedstatement.scala
@@ -60,7 +60,7 @@ object preparedstatement {
   /** @group Execution */
   def process[A: Composite]: Process[PreparedStatementIO, A] =
     resource(PS.executeQuery)(rs =>
-             PS.liftResultSet(rs, RS.close))(rs => 
+             PS.liftResultSet(rs, RS.close))(rs =>
              PS.liftResultSet(rs, resultset.getNext[A]))
 
   /**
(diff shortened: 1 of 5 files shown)


tpolecat:
After thinking about this I am less convinced that it makes sense for a `Composite` to have a non-constant width; composition depends on the ability to know that the `Composite` on the right will have the proper offset; if the width is determined dynamically then this doesn't work. Even a coproduct `A :+: B :+: C` with a discriminant will map to some fixed number of columns, although each member may use a different subset of them. We can for instance define a combinator for disjoint coproducts with a leading discriminant:

```scala
type ABC = A :+: B :+: C :+: CNil

def c3[A: Composite, B: Composite, C: Composite]: Composite[ABC] = 
  new Composite[ABC] {
    def get = n => HRS.get[Int](n) flatMap {
      case 0 => HRS.get[A](n + 1).map(Coproduct[ABC])
      case 1 => HRS.get[B](n + 1 + Composite[A].length).map(Coproduct[ABC])
      case 2 => HRS.get[C](n + 1 + Composite[A].length + Composite[B].length).map(Coproduct[ABC])
    }
    ...
  }
```

So I think maybe I'm missing the point. Do you have an example of how this would work?

wedens:
It's simple. Try to define `length` property on this `Composite` instance :)

wedens:
You will have the same problem with `flatMap` definition for `Composite` (imagine it's read-only for now). To define `length` property you'll need to know `length` of `Composite[B]`, and obviously you don't know it before `f` is applied

wedens:
You're right. I haven't thought about it.

PR #125, opened 2015-03-13
Composite split in read/write typeclasses

wedens:


tpolecat:
This looks good. I'm tagging master for releases right now so I'm going to sit on this until I tag 0.2.1, but this should be ok to merge after that. So give me a few days.

tpolecat:
I'm working on a more aggressive version of this for 0.3 so closing this.

PR #128, opened 2015-03-14
add driver class to hikari ctor

### LINKED ISSUE #127: Working Hikari Example
Be sure the `HikariTransactor` example works ... it seems like you need to do the `Class.forName` thing.

tpolecat:
As reported by @ppurang in the gitter channel, the Hikari transactor doesn't really work unless you load the driver class by hand. So this PR adds it as a ctor parameter and restores the example. This is a **breaking change** and will need doc. Fixes #127 pending doc.


ppurang:
+1 Thanks

PR #149, opened 2015-04-08
Unapply for FreeCoyo

### LINKED ISSUE #57: SqlStateOps instance not inferrable for non-aliased free monad
The `SqlStateOps` syntax in the PostgreSQL add-on doesn't work when the unaliased `Free` type is inferred. Unclear how to make this work. 

```scala
scala> 1.point[ConnectionIO].onStringDataRightTruncationClass01(null)
res1: doobie.imports.ConnectionIO[Int] = Gosub()

scala> 1.point[ConnectionIO].map(_ + 1).onStringDataRightTruncationClass01(null)
<console>:21: error: value onStringDataRightTruncationClass01 is not a member of scalaz.Free[[x]scalaz.Coyoneda[doobie.free.connection.ConnectionOp,x],Int]
              1.point[ConnectionIO].map(_ + 1).onStringDataRightTruncationClass01(null)
                                               ^

scala> (1.point[ConnectionIO].map(_ + 1) : ConnectionIO[Int]).onStringDataRightTruncationClass01(null)
res2: doobie.imports.ConnectionIO[Int] = Gosub()
```

### LINKED ISSUE #124: Add Kind Projector
Type lambdas are kind of a pain. Switch to KP.

tpolecat:
This adds an unapplied implicit constructor for `Monad[Free[Coyoneda[F, ?], ?]]` and an `Unapply` instance matching this shape. This prevents you from losing the `Monad` instance and associated syntax for things like `ConnectionIO` after the alias has been expanded.

Also adds [Kind Projector](https://github.com/non/kind-projector).

This fixes #124 and fixes #57.

Many thanks to Stephen Compall for helping me figure this out.

### DIFF
--- a/contrib/postgresql/src/main/scala/doobie/contrib/postgresql/syntax.scala
+++ b/contrib/postgresql/src/main/scala/doobie/contrib/postgresql/syntax.scala
@@ -1,6 +1,6 @@
 package doobie.contrib.postgresql
 
-import scalaz.{ Monad, Catchable }
+import scalaz.{ Monad, Catchable, Unapply }
 
 import doobie.contrib.postgresql.sqlstate._
 import doobie.util.catchsql.exceptSomeSqlState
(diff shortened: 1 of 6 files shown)


contrib/postgresql/src/main/scala/doobie/contrib/postgresql/syntax.scala
```diff
@@ -8,7 +8,14 @@ import doobie.util.catchsql.exceptSomeSqlState
 /** Module of recovery combinators for PostgreSQL-specific SQL states. */
 object syntax {
 
-  implicit class SqlStateOps[M[_]: Monad: Catchable, A](ma: M[A]) {
+  implicit def toSqlStateOps[M[_]: Monad: Catchable, A](ma: M[A]) =
+    new SqlStateOps(ma)
+
+  implicit def toSqlStateOpsUnapply[FA](v: FA)
+    (implicit F0: Unapply[Monad, FA], F1: Unapply[Catchable, FA]) =
+    new SqlStateOps[F0.M, F0.A](F0(v))(F0.TC, F1.TC.asInstanceOf[Catchable[F0.M]])
```
tpolecat: We know that `F0.M =:= F1.M` so this cast is fine, but it would be nice if we could prove it here. But we can't ask for implicit evidence here because the dependent method type rules don't allow it. I guess we may need another auxiliary type or something. Anyway, it works.

PR #171, opened 2015-06-03
Unify build files and bump sbt and specs2 versions

mdmoss:
Everything seems to work...
* Some tests fail, but they fail on master too
* I didn't touch `dash.sbt` because I'm not sure what's going on there
* I can't test publishing to bintray, so maybe throw a quick eye over that to make sure it looks alright


tpolecat:
This looks great, thank you! I will give it a closer look tonight.

Btw if you run `test` from the top level you'll see failures because the `examples` project includes some intentionally broken tests. The CI script only runs the "real" tests (see `.travis.yml`). Confusing, sorry.

tpolecat:
Really great contribution, thank you! :+1:

PR #183, opened 2015-07-05
more generalized function for collecting result into some data structure

wedens:
If you want to left untouched more performant function specialized to List, I can just add it as a new function without touching anything List related :)


tpolecat:
We can add `.accumulate` and use it for `.vector` but it's quadratic for `List`, so please restore the linear `.list` logic.

wedens:
opened new PR (accidentally committed to master)

PR #184, opened 2015-07-06
more generalized function for collecting result into some data structure

wedens:
#183


tpolecat:
Looks good, thanks. :+1:

PR #187, opened 2015-07-12
Add unsafeGet to meta/atom/composite

### LINKED ISSUE #161: Processing large result sets
First, thank you for making this library available. I think its quite ingenious, and I really like code I end up with when using it. One thing holding me back from testing it more broadly is performance on some of my common use cases. Maybe its early in doobie's lifetime for performance enhancements, but I though I'd toss this over in case there's an easy workaround (I didn't find one).

Very often I have queries that will return a few million rows back. Below are some benchmark results comparing JDBC, Slick, and Doobie, returning various numbers of rows back from a four column table (connection time is included). Doobie struggles on large result sets.

One bottleneck seems to be the match statement in resultset kleisliTrans: https://github.com/tpolecat/doobie/blob/master/core/src/main/scala/doobie/free/resultset.scala#L1583. On a four column table, its called nine times per row. Given that this work is redundant (the transformation is the same for each row), is there a way to avoid this?

```
[info] Benchmark                 (n)  Mode  Cnt   Score    Error  Units
[info] PerfTest.test_doobie       10  avgt    5   0.015 ±  0.001   s/op
[info] PerfTest.test_doobie     1000  avgt    5   0.061 ±  0.007   s/op
[info] PerfTest.test_doobie   100000  avgt    5   4.355 ±  0.355   s/op
[info] PerfTest.test_doobie  1000000  avgt    5  44.889 ±  3.233   s/op
[info] PerfTest.test_jdbc         10  avgt    5   0.012 ±  0.001   s/op
[info] PerfTest.test_jdbc       1000  avgt    5   0.015 ±  0.001   s/op
[info] PerfTest.test_jdbc     100000  avgt    5   0.278 ±  0.010   s/op
[info] PerfTest.test_jdbc    1000000  avgt    5   2.707 ±  0.371   s/op
[info] PerfTest.test_slick        10  avgt    5   0.015 ±  0.002   s/op
[info] PerfTest.test_slick      1000  avgt    5   0.019 ±  0.003   s/op
[info] PerfTest.test_slick    100000  avgt    5   0.286 ±  0.013   s/op
[info] PerfTest.test_slick   1000000  avgt    5   2.770 ±  0.210   s/op
```

Again, brilliant work!

tpolecat:
Added `unsafeGet` to `Meta/Atom/Composite`, which allows optimization for column vector reads that make `RS.list` and similar collectors indistinguishable from hand-written JDBC. Resolves #161 

Other changes:
- Benchmark updated.
- Added optimized `HRS.ilist`.
- New constructors `HRS.build` and `HRS.buildMap` for fast accumulation via `CanBuildFrom`.
- Updated `Query/Query0` to support `to[F[_]]` via `CanBuildFrom`.


tpolecat:
:-1: for now, need to add generalized `CBF` for `Query` since the one-off operations are getting out of hand.

tpolecat:
:-1: problem with `Query0.map` when constructed via `Query.toQuery0` … fix in progress.

PR #188, opened 2015-07-13
Exclude stale postgres:postgres dependency from org.postgis:postgis-jdbc (fixes #181)

### LINKED ISSUE #181: postgis dependency pulls in ancient version of postgresql
the postgresql extensions for doobie 0.2.2 depend on postgis which in turn pulls in an old version of postgresql. 

I believe we should make the postgis extension optional and additionally mark the dependency as provided in the sbt config

kryptt:


tpolecat:
Oh awesome, I didn't know how to do this. :+1: pending CI

PR #197, opened 2015-07-18
bintray to sonatype

### LINKED ISSUE #194: publish snapshots to sonatype and releases to maven central
blocked on https://github.com/scalaz/scalaz-stream/issues/258

and I'm happy to help.

http://central.sonatype.org/pages/ossrh-guide.html

For now, just create an account and raise a ticket asking for permission to publish to the entire `org.tpolecat` organisation. You'll probably want to have some way of validating that you own `tpolecat.org` or that you're the only `tpolecat` worth their consideration.

tpolecat:
Ok this is an attempt to set up sonatype publishing. A snapshot publish from this branch seemed to work, so we'll see if it shows up on central. Assuming this works, it resolves #194

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,7 @@ bin/
 
 # Ctags
 .tags
+
+# ENSIME
+.ensime
+.ensime_cache/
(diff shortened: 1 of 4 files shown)


tpolecat:
Ok this is good enough for snapshots. May run into trouble publishing for reals but there's no way to know until I get there.

PR #208, opened 2015-08-07
ignore useless NoNulls in analysis

### LINKED ISSUE #164: specs2 check fails inappropriately under JOIN circumstances

tpolecat:
This PR removes the check for needlessly lifted types (`Option` when no `NULL` values are expected) because JDBC's `NoNulls` metadata is useless ("might not be nullable"). 

Resolves #164

### DIFF
--- a/core/src/main/scala/doobie/util/analysis.scala
+++ b/core/src/main/scala/doobie/util/analysis.scala
@@ -66,8 +66,9 @@ object analysis {
   case class NullabilityMisalignment(index: Int, name: String, st: Meta[_], jdk: NullabilityKnown, jdbc: NullabilityKnown) extends AlignmentError {
     val tag = "C"
     def msg = this match {
-      case NullabilityMisalignment(i, name, st, NoNulls, Nullable) => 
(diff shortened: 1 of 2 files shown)


kryptt:
It would be nice if the information wasn't completely discarded. There is some benefit in letting the user know there *might* be an unnecessary `Option`

PR #213, opened 2015-08-16
Initial commit, add support for reading from an informix SmartBlob

freskog:
Initial support for Informix, requires a bit more work which I can add if the approach looks ok.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -85,8 +85,8 @@ lazy val doobie = project.in(file("."))
   .settings(noPublishSettings)
   .settings(unidocSettings)
   .settings(unidocProjectFilter in (ScalaUnidoc, unidoc) := inAnyProject -- inProjects(example, bench, docs))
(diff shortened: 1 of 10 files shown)


tpolecat:
Yep, this looks like a reasonable approach. Looks like you're following the pattern from contrib-postgres, which is perfect.

We'll need to figure out a strategy for testing ... is there a free version of Informix?

Thanks for doing this! Let me know if you need any guidance.

freskog:
As far as I know there's no repository that allows free access to the jdbc driver. This complicates things a bit. My idea is to basically just provide a stub (the interface) in the db_driver_stubs package, and that is something that can be used as a provided dependency in doobie. It could also be used as a test dependency for projects that need to mock out the actual driver in their project (that's what I'm currently doing).

As for testing the actual implementation, any ideas?

tpolecat:
Closing this; if you end up working on it more please open a new PR; this one is against a stale branch.

PR #219, opened 2015-09-07
param typeclass, in-clause support

### LINKED ISSUE #80: Support for IN (...) clauses
Need a better solution than this:

```scala
def mkSql(names: List[String]): String =
  s"select id, name, age from person where name in (${names.map(_ => "?").mkString(", ")})"
 
def mkSet(names: List[String]): PreparedStatementIO[Unit] =
  names.zipWithIndex.foldRight(().point[PreparedStatementIO]) { case ((s, i), p) =>
    HPS.set(i + 1, s) *> p
  }
 
val names = List("Alice", "Bob")
 
scala> HC.process[(Int, String, Option[String])](mkSql(names), mkSet(names)).quick.run
  (1,Bob,None)
  (3,Alice,Some(42))
```

### LINKED ISSUE #202: atom constraint on sql params went away
The change to `ProductArgs` means you can use any `Composite` as an interpolated parameter because everything gets flattened into an `HList`. The failing example in [Chapter 10](http://tpolecat.github.io/doobie-0.2.2/10-Custom-Mappings.html) actually compiles. You can't actually use the resulting query because the arity will be messed up.

tpolecat:
This adds a `Param` typeclass that's used only by the `sql` interpolator. This ensures that we can't use arbitrary `Composite` instances as parameters and thus resolves #202. It also provides a way to do `IN` clauses and thus resolves #80.

### DIFF
--- a/doc/src/main/tut/10-Custom-Mappings.md
+++ b/doc/src/main/tut/10-Custom-Mappings.md
@@ -89,7 +89,7 @@ Composite[PersonId].length
 
 However if we try to use this type for a *single* column value (i.e., as a query parameter, which requires an `Atom` instance), it doesn't compile.
 
-```tut:nofail
+```tut:fail
 sql"select * from person where id = $pid"
 ```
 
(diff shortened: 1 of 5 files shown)


doc/src/main/tut/10-Custom-Mappings.md
```diff
@@ -89,7 +89,7 @@ Composite[PersonId].length
 
 However if we try to use this type for a *single* column value (i.e., as a query parameter, which requires an `Atom` instance), it doesn't compile.
 
-```tut:nofail
+```tut:fail
 sql"select * from person where id = $pid"
```
tpolecat: This was actually compiling in `0.2.2` which was a bug.

Issue #228, opened 2015-09-26
parameterized mysql query checks fail more spectacularly than expected


refried:
```text
// check(myQuery(0))
[error]   x SQL Compiles and Typechecks
[error]    x Parameter metadata not available for the given statement (specs2.scala:61)
```

I'm don't know what's going on behind the scenes, so I don't have much info to contribute here.

Some people have claimed that the `?generateSimpleParameterMetadata=true` url parameter avoids this exception for some situations — this doesn't seem to be one of them though.

* Is the url parameter really being passed through?
* If it's really a MySQL bug, would it be possible to just run the query; ignore the result set, and alert whether or not the server reports it was good SQL?
* Is it possible that it's a doobie bug and not a MySQL bug?  Or is there some special mysql initialization that needs to be done before it will work properly?  Is it possible to dump the doobie program that is being executed as JDBC commands, for external debugging?

Thanks!

refried:
Last night we saw in the `com.mysql.jdbc.MysqlParameterMetadata` source that MySQL's `generateSimpleParameterMetadata` flag just switches between throwing an exception on all parameter metadata checks, vs throwing on some (e.g. nullability check) and returning junk data (e.g. `varchar`, `0`, `false`) on other checks; and should probably be called `generateGarbageParameterMetadata`!

Something like @miketkelly's result-set-only check 
https://github.com/miketkelly/doobie/commit/81907dc362a69c1714f45ab25d1e1b38a7d17aa3
would be a big improvement over the current test failure when parameter metadata is not available (i.e. when `java.sql.ParameterMetaData`'s methods throw `SQLException`).  `doobie.hi.preparedstatement#getParameterJdbcMeta` would probably have to return some signal.

I could try a PR, but how would you want to thread the warning message through — that "btw, parameter metadata not available, but here is the result set type check"?

refried:
I tried implementing his check by extending `AnalysisSpec`, but then the tests needed to look up `Composite` instead of using the instance already captured by `Query0.apply`.

PR #230, opened 2015-09-27
trying to support metadata analysis of query output (without parameters)

refried:
Related to #228, no functional differences to @miketkelly's branch.  I'm currently building against this version.

### DIFF
--- a/contrib/specs2/src/main/scala/doobie/contrib/specs2/specs2.scala
+++ b/contrib/specs2/src/main/scala/doobie/contrib/specs2/specs2.scala
@@ -49,6 +49,9 @@ object analysisspec {
     def check[A](q: Query0[A])(implicit A: TypeTag[A]) =
       checkAnalysis(s"Query0[${typeName(A)}]", q.stackFrame, q.sql, q.analysis)
 
+    def checkOutput[A](q: Query0[A])(implicit A: TypeTag[A]) =
+      checkAnalysis(s"Query0[${typeName(A)}]", q.stackFrame, q.sql, q.outputAnalysis)
+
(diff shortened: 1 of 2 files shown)


tpolecat:
I'm going to sit on this because I want to square this with the MS-SQL metadata issue first, but I think this is probably a good approach for now. Thanks!

refried:
Anything I can do to help?  I can grab a SQL Server trial.

tpolecat:
@refried thanks, I'll let you know. That might be helpful.

tpolecat:
I tried a few things and I think this approach might be the best option. I was trying to make it emit a warning :warning: and continue if it can't figure out the parameter types, but there's no way to distinguish this case from a syntax error in the SQL, which should be a failure :x: ... they both show up as vendor-specific exceptions that I can't distinguish in a generic way.

With your approach you can just say ok I just won't look at the parameters. Which is lame but superior in that it works. So will probably merge unless I am inspired otherwise in the next few hours. :+1:

refried:
I'm pretty happy with this for now, but to play devil's advocate: 
Does bad SQL let you get as far as `doobie.hi.preparedstatement#getParameterJdbcMeta`? And if not, how big of a mess would it make if you put a `catch` in there and changed the type to `PreparedStatementIO[Option[List[ParameterMeta]]]` or whatever you use for errors if not `Option`.

tpolecat:
@refried Yeah I tried that. The problem is that `prepareStatement` doesn't do anything immediately (at least for PostgreSQL) so problems in the SQL don't appear until you ask for the metadata, at which point you can't distinguish bad SQL from failure fetching metadata. So you don't know if it should be a failure or a warning.

tpolecat:
Closed via #235

PR #235, opened 2015-10-06
metadata analysis of query output

### LINKED ISSUE #222: MS-SQL Driver failure when requesting metadata for CTE queries
Forked from #218. The failure is reported as a problem with the SQL, but the failure actually occurs when requesting column metadata. The analysis should report the failure as a warning that says something like "due to driver limitations column metadata is unavailable for this query".

> I tested some queries with the snapshot and it seems to work so far. However the trivial one still fails as well as queries containing CTEs here are some failing examples:
> 
>     val trivial = sql"""SELECT 42""".query[Int]
> 
>     val cte =
>           sql"""WITH cte_test AS (
>                 SELECT 42::int AS "FOO", 43::int AS "BAR"
>               ) SELECT * FROM cte_test WHERE "FOO" = 42
>            """.query[(Int, Int)]
> 
>     check(trivial)
>     check(cte)
> 
> Fails like so:
> 
>     [info] Query0[Int] defined at DBTests.scala:18
>     [info]   SELECT 42
>     [error]   x SQL Compiles and Typechecks
>     [error]    x com.microsoft.sqlserver.jdbc.SQLServerException: Unable to identify the table
>     [error]      SELECT 42 for the metadata. (specs2.scala:61)
>     [info]
>     [info] Query0[(Int, Int)] defined at DBTests.scala:21
>     [info]   WITH cte_test AS (
>     [info]   SELECT 42::int AS "FOO", 43::int AS "BAR"
>     [info]   ) SELECT * FROM cte_test WHERE "FOO" = 42
>     [info]
>     [error]   x SQL Compiles and Typechecks
>     [error]    x com.microsoft.sqlserver.jdbc.SQLServerException: Unable to identify the table
>     [error]      WITH cte_test AS (
>     [error]     SELECT 42::int AS "FOO", 43::int AS "BAR"
>     [error]     ) SELECT * FROM
>     [error]      cte_test WHERE "FOO" = 42
>     [error]     for the metadata. (specs2.scala:61)
> 

/cc @raichoo

### LINKED ISSUE #228: parameterized mysql query checks fail more spectacularly than expected

### LINKED ISSUE #231: MS-SQL driver failure when requesting parameter metadata
Likely a duplicate of #222. The Microsoft JDBC driver's implementation of `getParameterMetadata` parses the statement text to generate a new query that selects the parameters as columns. As an example, for the purposes of parameter metadata this query:

```
select code from country where population > ?
```

is sent to the server as:

```
SET FMTONLY ON SELECT population FROM country WHERE 1 = 2
```

The parser inside the driver that attempts this is extremely rudimentary, and often fails on valid sql.

This works:
```
scala> def q(n: Int) = sql"select code from country where population > $n".query[String]
 
scala> q(0).check.run

  select code from country where population > ?

  ✓ SQL Compiles and Typechecks
  ✓ P01 Int  →  INTEGER (int)
  ✓ C01 code CHAR (char) NOT NULL  →  String
```

But if you add some whitespace it fails:
```
scala> def q(n: Int) = sql"""
     select code
     from country
     where population > $n
   """.query[String]

      select code
      from country
      where population > ?

scala> q(0).check.run


  ✕ SQL Compiles and Typechecks
    - com.microsoft.sqlserver.jdbc.SQLServerException: Unable to identify the table

```

And some things like between clauses will never work:

```
scala> def q(minPop: Int, maxPop: Int) = sql"select code from country where population between $minPop and $maxPop".query[String]

scala> q(0, 0).check.run

  select code from country where population between ? and ?

  ✕ SQL Compiles and Typechecks
    - com.microsoft.sqlserver.jdbc.SQLServerException: Incorrect syntax near the
      keyword 'between'.

// actual query sent to sql server:
// SET FMTONLY ON SELECT between, and FROM country WHERE 1 = 2
```

tpolecat:
Forked from #230 ... added `checkOutput` to `Yolo`.

(Also rebased, not that it matters.)

In a lame way this
- resolves #231
- resolves #228 
- resolves #222

### DIFF
--- a/contrib/specs2/src/main/scala/doobie/contrib/specs2/specs2.scala
+++ b/contrib/specs2/src/main/scala/doobie/contrib/specs2/specs2.scala
@@ -49,6 +49,9 @@ object analysisspec {
     def check[A](q: Query0[A])(implicit A: TypeTag[A]) =
       checkAnalysis(s"Query0[${typeName(A)}]", q.stackFrame, q.sql, q.analysis)
 
+    def checkOutput[A](q: Query0[A])(implicit A: TypeTag[A]) =
+      checkAnalysis(s"Query0[${typeName(A)}]", q.stackFrame, q.sql, q.outputAnalysis)
+
     def check[A](q: Update[A])(implicit A: TypeTag[A]) =
       checkAnalysis(s"Update[${typeName(A)}]", q.stackFrame, q.sql, q.analysis)
 
--- a/core/src/main/scala/doobie/util/query.scala
+++ b/core/src/main/scala/doobie/util/query.scala
@@ -49,6 +49,12 @@ object query {
      */
     def analysis: ConnectionIO[Analysis] =
       HC.prepareQueryAnalysis[I, O](sql)
+    /**
+     * Program to construct an analysis of this query's SQL statement and result set column types.
+     * @group Diagnostics
+     */
+    def outputAnalysis: ConnectionIO[Analysis] =
+      HC.prepareQueryAnalysis0[O](sql)
 
     /**
      * Apply the argument `a` to construct a `Process` with effect type 
@@ -130,6 +136,7 @@ object query {
         def sql = outer.sql
         def stackFrame = outer.stackFrame
         def analysis = outer.analysis
+        def outputAnalysis = outer.outputAnalysis
         def process = outer.process(a)
         def to[F[_]](implicit cbf: CanBuildFrom[Nothing, B, F[B]]) = outer.to[F](a)
         def accumulate[F[_]: MonadPlus] = outer.accumulate[F](a)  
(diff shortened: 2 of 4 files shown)


tpolecat:
:-1: for now, updating doc...

PR #238, opened 2015-10-07
Meta derivation for unary products

### LINKED ISSUE #234: no Atom/Param instance for 1-member case classes
> I have a query with a parameter that is an `Int` in the database.  In my code, it's represented by `case class MeetingId(int: Int)`.  I do automatically get a `Composite[MeetingId]`, but I get the error `Could not find or construct Param[MeetingId :: HNil]` and a hint saying, "make sure you can summon an `Atom`", which I can't.  Should I be able to get one automatically from something like `MeetingId` which is isomorphic to `Int :: HNil`?

.

> @refried hm yeah that's a corner case. for now you can work around it by constructing a `Meta`
`implicit val MeetingIdMeta: Meta[MeetingId] = Meta[Int].xmap(MeetingId, _.int)` Can you open an issue for that? That's irritating.

wedens:
fixes #234

### DIFF
--- a/core/src/main/scala/doobie/util/meta.scala
+++ b/core/src/main/scala/doobie/util/meta.scala
@@ -12,6 +12,8 @@ import scala.reflect.ClassTag
 import scala.Predef._
 
 import scalaz._, Scalaz._
+import shapeless._
+import shapeless.ops.hlist.IsHCons
 
 /** Module defining the lowest level of column mapping. */
 object meta {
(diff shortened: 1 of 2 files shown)


refried:
Check that the `Atom`s exist?

wedens:
@refried sorry, what `Atom`?

tpolecat:
:+1: lgtm. Any final words?

core/src/main/scala/doobie/util/meta.scala
```diff
@@ -295,6 +297,48 @@ object meta {
...
+      // representation (L) for type A
+      implicit gen: Generic.Aux[A, L],
+      // head (H) and tail (T) type of representation (L)
+      c: IsHCons.Aux[L, H, T],
+      // Meta instance for the head (the only element in representation)
+      hmeta: Lazy[Meta[H]],
+      // provide evidence that representation (L) and singleton hlist with
+      // the only element of type H are the same type
+      ev: =:=[H :: HNil, L],
+      ev2: Null <:< H
```
refried: Do you guys prefer this over `H >: Null` in the typearg list?

refried:
@wedens I was worried because I thought in testing a solution like this, I was getting ambiguity due to multiple derivation paths when I actually needed to materialize `Atom` or `Composite` instances.  It looks good here in the quick tests I did, but I can try in the big app tomorrow if you guys want me to.

What does `Lazy` / resolving `Meta` lazily do?  Doesn't it get used immediately, or how does that work?  Thanks!

wedens:
@refried `Lazy` prevents some diverging implicits (for more info see comments in shapeless). it's used only when derived `Meta` is used.

core/src/main/scala/doobie/util/meta.scala
wedens: dunno. I'm bad at nullity/subtyping. just copied it from `nxmap`

refried: Ah ok yeah.  I guess `Meta#nxmap` does it that way because its `A` is declared outside of `nxmap`.

refried:
@wedens Ah, thanks sorry.  I didn't see any comments in lazy.scala but I see some in the Feature Overview.  Maybe `Lazy` is what I was missing from my implementation.

Thanks a lot for pulling this together.

wedens:
@refried https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/lazy.scala#L25

refried:
@wedens Thank you!  My IDE was showing me this instead: https://github.com/milessabin/shapeless/blob/shapeless-2.2.2/core/src/main/scala/shapeless/lazy.scala

tpolecat:
Ok you guys tell me when you're happy with the state of things. There was a NPE in one of the 4 CI builds in `compositespec` so I restarted it ... maybe a threading thing with specs.

wedens:
@tpolecat :+1:

refried:
Just got a chance to drop it into my bigger project and it seems to work great.  Thanks guys.

Issue #237, opened 2015-10-06
Upgrade to scalaz-stream 0.8


tpolecat:
:warning: This will require changing the resource combinator. See the [0.8 release notes](https://github.com/scalaz/scalaz-stream/wiki/0.8) for details.

PR #254, opened 2015-11-21
#237 Upgrade to scalaz-stream 0.8

guersam:
Use `bracket` instead of `eval` with inner `onComplete`.

One concern is that ` onHalt { _.asHalt }` is [generally recommended against](https://github.com/functional-streams-for-scala/fs2/wiki/0.8#terminated-removal), but I just took it for now as this one is also used in [scalaz-stream's own resource combinator](https://github.com/functional-streams-for-scala/fs2/blob/release/0.8/src/main/scala/scalaz/stream/io.scala#L152).

### DIFF
--- a/core/src/main/scala/doobie/free/blob.scala
+++ b/core/src/main/scala/doobie/free/blob.scala
@@ -46,7 +46,7 @@ import resultset.ResultSetIO
  *
  * `BlobIO` is a free monad that must be run via an interpreter, most commonly via
  * natural transformation of its underlying algebra `BlobOp` to another monad via
- * `Free.runFC`. 
+ * `Free#foldMap`.
  *
  * The library provides a natural transformation to `Kleisli[M, Blob, A]` for any
  * exception-trapping (`Catchable`) and effect-capturing (`Capture`) monad `M`. Such evidence is 
--- a/core/src/main/scala/doobie/free/blob.scala
+++ b/core/src/main/scala/doobie/free/blob.scala
@@ -1,6 +1,6 @@
 package doobie.free
 
-import scalaz.{ Catchable, Coyoneda, Free => F, Kleisli, Monad, ~>, \/ }
+import scalaz.{ Catchable, Free => F, Kleisli, Monad, ~>, \/ }
 import scalaz.concurrent.Task
 
 import doobie.util.capture._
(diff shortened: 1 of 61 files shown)


tpolecat:
This looks great, thanks! I'm going to get one more set of eyes on this just in case, because I'm not confident that I understand the changes in semantics from `0.7x` to `0.8`. But :+1: from me.

tpolecat:
@timperrett confirms that the changes in resource handling for streams 0.8 breaks doobie 0.2.3, which is problematic because sbt will happily evict 0.7. The symptom is programs crashing with `Terminated(End)` and no other information. So we need to confirm that the changes here fix that problem. I suspect they do.

guersam:
Updated to 0.8a as well as Scalaz 7.2.0.

core/src/main/scala/doobie/hi/preparedstatement.scala
```diff
@@ -72,7 +69,12 @@ object preparedstatement {
 
   /** @group Batching */
   val executeBatch: PreparedStatementIO[List[Int]] =
-    PS.executeBatch.map(_.toList)
+    PS.executeBatch.map(arr =>
+      // `Predef.intArrayOps` is not implicit as of Scala 2.12.0-M3
+      // https://github.com/scala/scala/blob/v2.12.0-M3/src/library/scala/Predef.scala#L336
+      // TODO replace with `_.toList` after it gets back
+      intArrayOps(arr).toList
```
guersam: It's the only (temporary) fix needed for 2.12.0-M3, as well as in `c/s/m/s/d/hi/statement.scala`.

core/src/main/scala/doobie/free/blob.scala
```diff
@@ -156,14 +156,7 @@ object blob {
...
    */
-  type BlobIO[A] = F.FreeC[BlobOp, A]
-
-  /**
-   * Monad instance for [[BlobIO]] (can't be inferred).
-   * @group Typeclass Instances 
-   */
-  implicit val MonadBlobIO: Monad[BlobIO] = 
-    F.freeMonad[({type λ[α] = Coyoneda[BlobOp, α]})#λ]
+  type BlobIO[A] = F[BlobOp, A]
```
guersam: I don't really like aliasing `Free[_, _]` to `F[_, _]` here and there, but I'll keep it for now otherwise it conflicts with `blob.Free` op.

tpolecat:
:+1: this looks great, thanks!

I'm going to sit on this for the moment because I'm not sure if the next release will be 0.2.4 or 0.3 ... once I figure that out I will know where to merge this.

guersam:
No problem, I still have to deal with tut and argonaut (maybe circe) anyway
:)

guersam:
Current status:
- [x] Upgrade scalaz to 7.2.0
- [x] Upgrade scalaz-stream to 0.8a
- [x] Cross build with scala 2.12.0-M3
- [x] Apply workaround for tut regarding scala 2.12
- [ ] Upgrade shapeless to 2.3.0 (currently RC2)

tpolecat:
Thanks for your continued work on this. I'll take a closer look this week, I hope.

PR #268, opened 2016-02-05
add result type to imports.freeMonadC

refried:
I had copied `freeMonadC` and `unapplyMMFA` into my own project, but they wouldn't work for me until I added the result type to `freeMonadC`.  Thought you might want it here too, for good measure.


tpolecat:
Oh very good catch. :+1: thanks!

PR #272, opened 2016-02-18
add NonEmptyList `.nel` support to RS, Query, Query0, docs

refried:
What do you think about this, to convert the following sort of construction (and its accompanying warning):

```scala
implicit class toNelUnsafe[F[_] : Functor, A](fa: F[List[A]]) {
  def toNelUnsafe: F[NonEmptyList[A]] = fa map {
    case a :: as => NonEmptyList.nel(a, as) // not total
  }
}

def getItems(orderId: OrderId): ConnectionIO[NonEmptyList[ItemId]] =
  sql"""
    SELECT item_id FROM order_contents WHERE order_id = $orderId
  """.query[ItemId].list.toNelUnsafe
```

into 

```scala
def getItems(orderId: OrderId): ConnectionIO[NonEmptyList[ItemId]] =
  sql"""
    SELECT item_id FROM order_contents WHERE order_id = $orderId
  """.query[ItemId].nel // safe!
```

P.S. Some whitespace changes got caught up in the first commit too, despite having carefully reverted them once; I converted some legacy tabs to spaces in the second commit.

P.P.S. Sorry for all the edits, if they triggered multiple emails.

### DIFF
--- a/core/src/main/scala/doobie/hi/resultset.scala
+++ b/core/src/main/scala/doobie/hi/resultset.scala
@@ -210,7 +210,7 @@ object resultset {
     * @throws `UnexpectedEnd` if there is not at least one row remaining
     * @group Results
     */
-  def getNel[A: Composite]: ResultSetIO[NonEmptyList[A]] =
+  def nel[A: Composite]: ResultSetIO[NonEmptyList[A]] =
     (getNext[A] |@| list) {
       case (Some(a), as) => NonEmptyList.nel(a, as)
       case (None, _)     => throw UnexpectedEnd
--- a/core/src/main/scala/doobie/hi/resultset.scala
+++ b/core/src/main/scala/doobie/hi/resultset.scala
@@ -25,7 +25,7 @@ import scala.collection.JavaConverters._
 import scala.collection.generic.CanBuildFrom
 import scala.Predef.intArrayOps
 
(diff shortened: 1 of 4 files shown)


refried:
3 out of 4 successful builds is pretty good, right Travis?

tpolecat:
Huh. I restarted that build. Confidence high.

core/src/main/scala/doobie/hi/resultset.scala
```diff
@@ -205,6 +205,17 @@ object resultset {
       case (None, _)            => None
     }
 
+  /**
+    * Consumes the remainder of the resultset, but verifies that there is at least one row remaining.
+    * @throws `UnexpectedEnd` if there is not at least one row remaining
+    * @group Results
+    */
+  def getNel[A: Composite]: ResultSetIO[NonEmptyList[A]] =
```
tpolecat: Can we just call this `nel`?

tpolecat:
Looks great! Please do that rename unless there's a reason not to. Otherwise :+1: 

Thanks!!

core/src/main/scala/doobie/hi/resultset.scala
refried: Yep!  I'll send an update.

refried:
(Just fyi: I had named it like `resultset#getOption` & `getUnique`)

PR #285, opened 2016-04-22
wip: new transactor design

tpolecat:
This is work in progress on series/0.3.x that turns `Transactor` into a multi-parameter typeclass (with a weaker `Connector` superclass). The goal is to get rid of subclassing as a means of abstraction and allow direct use of existing types as connection/transaction arbiters without wrapping.

- `Connector[M, A]` witnesses that a value of type `A` can construct an `M[Connection]` and gives rise to `ConnectionIO ~> M` *without* transaction semantics. This lets you run a `ConnectionIO` on a raw `Connection` without a transaction, or hop onto an existing connection managed by some other service.
- `Transactor[M, A]` is a refinement of `Connector[M, A]` that provides "safe" versions of the natural transformations defined by `Connector` that wrap a `ConnectionIO` program in transactional semantics as defined by a *pluggable* `LiftXA`. This provides the same functionality as in the `0.2.x` series but allows the user to manipulate the transactional structure, which currently requires a custom implementation.

User-level changes are:
- Current `Transactor` implementations go away but source-level stubs will allow most code to continue to compile. Anything with a declared type of `Transactor[M]` will need a new ascribed type.
- Calls to `.transact(xa)` become `transact[M](xa)` for some supported `M` like `Task` or `IO`.
- Constructing a `Yolo` from a transactor requires a type argument now, so `import xa.yolo._` must now be written in two steps: `val y = xa.yolo[IO]; import y._` due to stable identifier requirement for imports.

Task list:
- [x] core
- [x] h2
- [x] hikari
- [x] postgres
- [x] specs2
- [x] examples
- [x] bench
- [ ] docs
  - [x] get compiling
  - [ ] edit for transactor updates
  - [ ] publish book draft
  - [ ] publish scaladoc
- [x] bring readme and changelog up to date

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ This will be a **major release** with **breaking changes** to come.
 Special thanks to **@guersam** for updating the build and all the dependencies; this was nontrivial.
 
 - **doobie** artifacts now include OSGi headers.
-- Added `.nel` handlers for reading result sets that are expected to be no-empty.
+- Added `.nel` handlers for reading result sets that are expected to be non-empty.
 
 Upgrades:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This file summarizes **notable** changes for each release, but does not describe
 
 This will be a **major release** with **breaking changes** to come. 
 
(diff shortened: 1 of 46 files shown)


tpolecat:
Closing, re-opening on a new branch.

PR #293, opened 2016-06-08
Allow construction of `HikariTransactor` instances from an existing `HikariTransactorDatasource`

raulraja:
This PR allows reusing `HikariDataSource` instances potentially initialized elsewhere (ex. Play DB facilities) to construct Doobie's `HikariTransactor`.

@tpolecat Thanks for your consideration!

### DIFF
--- a/contrib/hikari/src/main/scala/doobie/contrib/hikari/HikariTransactor.scala
+++ b/contrib/hikari/src/main/scala/doobie/contrib/hikari/HikariTransactor.scala
@@ -4,7 +4,6 @@ import com.zaxxer.hikari.HikariDataSource
 import doobie.imports._
 
 import scalaz.{ Catchable, Monad }
-import scalaz.syntax.id._
 import scalaz.syntax.monad._
 
 object hikaritransactor {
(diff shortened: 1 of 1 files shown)


contrib/hikari/src/main/scala/doobie/contrib/hikari/HikariTransactor.scala
```diff
@@ -28,6 +27,10 @@ object hikaritransactor {
     def initial[M[_]: Monad : Catchable: Capture]: M[HikariTransactor[M]] = 
       Capture[M].apply(new HikariTransactor(new HikariDataSource))
 
+    /** Constructs a program that yields a `HikariTransactor` from an existing `HikariDatasource`. */
+    def apply[M[_]: Monad : Catchable: Capture](hikariDataSource : HikariDataSource): M[HikariTransactor[M]] = 
+      Capture[M].apply(new HikariTransactor(hikariDataSource))
```
tpolecat: This doesn't need to be captured in `M` because it's not a side-effect; you can just return `HikariTransactor[M]`.

raulraja:
@tpolecat I updated the code to just return the Transactor as suggested. Thanks!

tpolecat:
Thanks!

PR #296, opened 2016-07-05
Add support for UUID array

### LINKED ISSUE #295: Implement support for Vector[UUID] by default
By default, it doesn't look like querying `Vector[UUID]` is supported, but it's easy enough to fix by adding:

```scala
implicit val (unliftedUUIDArrayType, liftedUUIDArrayType)  = boxedPair[java.util.UUID] ("uuid", "_uuid")
```

veegee:
This should fix #295


tpolecat:
Looks good. Can you add a test [here](https://github.com/tpolecat/doobie/blob/series/0.2.x/contrib/postgresql/src/test/scala/doobie/contrib/postgresql/pgtypes.scala#L149)?

veegee:
Yes, done.

tpolecat:
Thanks!

PR #304, opened 2016-07-15
Add support for cats via preprocessor.

tpolecat:
This adds support for two FP environments by building doobie natively for both. This means there are no shims or indirection; it's distinct builds.

- **scalaz 7.2** with **scalaz-stream 0.8**
- **cats 0.6** with **fs2 0.9** (M5)

This is accomplished via a preprocessor called `yax` that provides `#ifdef`-style switches. It's very primitive. The high points:

- All source is now under `yax/`.
- There are now two versions of each sbt project: `core`, `core_cats`, and so on. They have no source of their own and don't appear in source control, but they're generated when you run sbt and you see them under `modules/` and `modules-cats/`.
- Line numbers in errors are wrong. There's a fix for that but it messes up tut so it needs more work.
- cats artifact names are the same but end in `-cats`, so `doobie-core-cats`, `doobie-h2-cats`, and so on.

Unrelated changes in this PR (sorry):

- The `contrib` segment in artifacts and package names is gone. So `doobie-h2` is the artifact now and `doobie.h2` is the package name.
- The `posgresql` segment and package name has been shortened to `postgres`.

👎  still need
- [x] document module/package/artifact renaming
- [ ] update readme
- [ ] get 2.10 and 2.12 builds working for scalaz (cats + fs2 is 2.11 only for now)


tpolecat:
Aaaaand only 33 commits later everything builds, all tests pass, and the doc builds for both environments.

tpolecat:
Subsumed by #309

PR #315, opened 2016-08-03
Adding pginterval from java.time.Duration support

brianpendleton:


guersam:
It will be so great to have proper support for the interval type.

Unfortunately, `java.time.Duration` represents "A time-based amount of time, such as '34.5 seconds'" whereas `org.postrgesql.util.PGInterval` consists of each chronological unit so that there's no exact 1:1 mapping between two types. For instance, the current implementation doesn't pass the round trip test at `Duration.ofDays(360)` or above. It also loses nanosecond level precision because `PGInterval` uses floating point arithmetic to store small units.

On the other hand, `java.time.Period` is a date-based interval type which does the same thing as `PGInterval` does, but it doesn't have any time-related info. I have some queries that use week/month level intervals so that `Period` fits my use case better, but when users need an interval in seconds or hours they'll always get zero in this case.

tpolecat:
@guersam Thanks for looking into it. I'm not sure what the right approach is from a library standpoint. Maybe just provide the `PGInterval` mapping and leave it at that.

PR #318, opened 2016-08-11
Boost Travis build using some tricks from https://monix.io

guersam:
- Use build matrix with different commands according to scala versions
- Use SBT/Ivy cache

The cache sometimes causes weird build failure which needs a manual flush, but it's unusual case and generally works good.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,21 +5,23 @@ language: scala
 jdk:
   - oraclejdk8
 
-script: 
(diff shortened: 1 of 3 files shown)


guersam:
Impressive, it reduced the build time from 27 min 55 sec to 9 min 38 sec. I expect further reduction since the cache is applied.

tpolecat:
Brilliant, this is great.

PR #319, opened 2016-08-11
Fix #316 - OutOfMemoryError when fetching data from large tables

### LINKED ISSUE #316: OutOfMemoryError when fetching data from large tables
Doobie: v0.3.0, 0.2.4
Scala: 2.11.7
System: Mac OS El Capitan
DB: PostgreSQL v9.4.4

Considering a table with 10M records, processing of q1 and q2 behaves differently:
```
val q1 = sql"SELECT * FROM tbl LIMIT 1000".query[Blah]
val q2 = sql"SELECT * FROM tbl LIMIT 10000000".query[Blah]

q1.process.take(5).quick.run // r1
q2.process.take(5).quick.run // r2
```

`r1` usually works fine, while `r2` fails with OutOfMemory (it's trying to fetch the whole 10M records beforehand).

guersam:
(NOTE: It's based upon #318)

This PR fixes two issues:
- `doobie.hi.resultset.getNextChunk` ignores given chunk size
- [The entire ResultSet is loaded at once](https://jdbc.postgresql.org/documentation/94/query.html#query-with-cursor)

### DIFF
--- a/yax/core/src/main/scala/doobie/hi/resultset.scala
+++ b/yax/core/src/main/scala/doobie/hi/resultset.scala
@@ -206,7 +206,7 @@ object resultset {
       val b = Vector.newBuilder[A]
       while (n > 0 && rs.next) {
         b += A.unsafeGet(rs, 1)
-        n += 1
+        n -= 1
       }
       b.result()
     }
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,21 +5,23 @@ language: scala
 jdk:
   - oraclejdk8
 
-script: 
-  - sbt ++2.11.8 scalastyle compile test:compile test
-  - sbt ++2.11.8 docs/tut
-  - sbt ++2.11.8 docs_cats/tut
-  - sbt ++2.10.6    core/test h2/test hikari/test postgres/test specs2/test
-  - sbt ++2.12.0-M4 core/test h2/test hikari/test postgres/test specs2/test
+scala:
+  - 2.10.6
+  - 2.11.8
+  - 2.12.0-M4
+
+env:
+  global:
+    - MAIN_SCALA_VERSION=2.11.8
 
 addons:
   postgresql: "9.3"
 
 before_script:
-  - psql -c 'create database world;' -U postgres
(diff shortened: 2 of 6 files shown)


yax/core/src/main/scala/doobie/hi/resultset.scala
```diff
@@ -206,7 +206,7 @@ object resultset {
       val b = Vector.newBuilder[A]
       while (n > 0 && rs.next) {
```
guersam: BTW, it'll yield empty vector when a chunkSize <= 0 is given. Should it be prevented at all or be handled as a special case?

tpolecat:
Thanks for figuring this out, I wasn't seeing it.

yax/core/src/main/scala/doobie/hi/resultset.scala
tpolecat: This is by design. I guess we could throw an exception since it doesn't really make any sense but it doesn't bother me as-is.

tpolecat:
Huh I thought merging #318 would remove those commits from this PR. I guess not.

yax/core/src/main/scala/doobie/hi/resultset.scala
```diff
@@ -206,7 +206,7 @@ object resultset {
...
         b += A.unsafeGet(rs, 1)
-        n += 1
+        n -= 1
```
tpolecat: hurr

tpolecat:
Works. This is great.

yax/core/src/main/scala/doobie/hi/resultset.scala
guersam: `(P)S.setFetchSize(0)` disables cursor so that the driver fetches all results at once. There might be some users who want this behavior, but I'm okay to be opinionated here.

tpolecat: If you're doing that there's not much reason to be using a stream, but yeah we should review the doc. Opened #322 for followup.

Issue #310, opened 2016-07-23
compile for scalaz+fs2?


tpolecat:
Moving to from scalaz-stream to fs2 is nontrivial so I'm not sure how much demand there will be initially, but this seems like a natural combination to support.

PR #323, opened 2016-08-12
WIP: Support for fs2.Task dropping own Catchable

### LINKED ISSUE #311: Support fs2.Task (but as generically as reasonably possible)
Currently, the FS2/cats port of doobie supports only `IOLite`, which is probably not a useful effect type in the larger ecosystem outside of doobie.

The port should support other "IO" types, the obvious questions are which ones, and how.

1. `fs2.Task` isn't very difficult to support explicitly, it includes enough functionality to implement both `Catchable` and `Capture`. Calling `Task` functionality directly is probably also more performant than more generic alternatives.

2. We could support every effect type that FS2 supports, which would involve using `fs2.util.Effect` to implement the appropriate typeclasses (and the functionality is sufficient, I think, though the `attempt` call has to be folded into an `Xor` result). It does mean that people who want to interoperate with doobie have to implement this FS2 typeclass instead of something in cats.

3. We could support every effect type that cats supports. Because of the various levels of indirection to support this, probably the least performant. But it would allow (for example) Monix Task support OOTB. How to implement `Catchable` seems obvious (use `MonadError`); the question is whether there is a reasonable implementation of `Capture` that we can reuse. I thought wrapping the value in `Eval.Always` might be sufficient, but perhaps not (I hadn't given it detailed thought). If cats doesn't have the appropriate abstraction, it _might_ be worth adding.

Soo... ticket is about implementing one or more of these, because we will need more than `IOLite` support.

guersam:
An attempt to fix #311.

Although I didn't mean to touch all these stuff at first, eventually I ended up with dropping `doobie.util.catchable.Catchable` in favor of `fs.util.Catchable` fighting a lot of ambiguous implicit errors. But probably there would be a much simpler way...

Changes made (which can be reverted) or things to discuss:
- [x] Replace most cats compat utils with [fs2-cats](https://github.com/functional-streams-for-scala/fs2-cats)
- [x] Replace `cats.data.Xor` with `scala.util.Either` to support `fs2.util.Catchable` (related discussion: https://github.com/typelevel/cats/issues/1192)
- [x] Use `fs2.util.Suspendable` instead of `Capture[fs2.Task]`
- [x] ~~Or revive `doobie.util.catchable.Catchable` and directly support only for `fs2.Task`?~~
- [x] ~~Drop `IOLite`?~~
- [x] Move `fs2.util.Suspendable` instance for `cats.data.Kleisli` into fs2-cats
- [ ] Decide which typeclasses (especially Monad) to support, of cats, or of fs2?

TODO:
- [ ] Align yax directives in a consistent way
- [ ] Update docs


project/FreeGen.scala
```diff
@@ -290,11 +295,24 @@ class FreeGen(managed: List[Class[_]], log: Logger) {
...
     |   */
+    |#+scalaz
     |  implicit val Catchable${sname}IO: Catchable[${sname}IO] =
     |    new Catchable[${sname}IO] {
     |      def attempt[A](f: ${sname}IO[A]): ${sname}IO[Throwable \\/ A] = ${sname.toLowerCase}.attempt(f)
     |      def fail[A](err: Throwable): ${sname}IO[A] = ${sname.toLowerCase}.delay(throw err)
     |    }
+    |#-scalaz
+    |#+cats
+    |#+fs2
```
guersam: I think `cats` directive is unnecessary here. Right?

guersam:
The second commit is slightly out of scope. I just found unnecessary diffs caused by varying method order with environments annoying.

tpolecat:
Yeah the ever-changing codegen is irritating. I will look at this more closely this weekend. It seems like a big change but it certainly makes sense to get rid of our own `Catchable` if one is provided by fs2.

guersam:
I have been struggling to combine `f.u.Catchable` and `d.u.Capture` into `f.u.Effect`, but got no clue to implement [`def unsafeRunAsync[A](fa: F[A])(cb: Attempt[A] => Unit): Unit`](https://github.com/functional-streams-for-scala/fs2/blob/v0.9.0-RC1/core/shared/src/main/scala/fs2/util/Effect.scala#L20) for `Free` algebras and `Kleisli[M[_]: Effect, E, A]`. Is it possible at all?

mpilquist:
FWIW, I think we could introduce the following to FS2:

  trait Suspendable[F[_]] extends Monad[F] {
    def suspend[A](fa: => F[A]): F[A]
    def delay[A](a: => A): F[A]
  }

  trait Effect[F] extends Catchable[F] with Suspendable[F] {
    def unsafeRunAsync[A](fa: F[A])(cb: Attempt[A] => Unit): Unit
  }

guersam:
@mpilquist That's perfect, I managed to write a modified version having every `unsafeRunAsync` to be `???`, which somehow passes all tests. `Suspendable` must work with it out of the box.

tpolecat:
@guersam thanks for your continued hacking on this. I didn't have any time over the weekend to get into it but I will catch up soon.

guersam:
@tpolecat No problem, and sorry about some inessential diffs which will the review harder. As noted above I didn't expect the changes to be this huge 😨

tpolecat:
Ok, so here are a few thoughts.

> Drop `IOLite`?

This is really irritating because scalaz `Task` has `.unsafePerformSync` and fs2 has `.unsafeRunSync` which means every exec in the book (and there are a lot) needs a yax directive.

> Move `fs2.util.Suspendable` instance for `cats.data.Kleisli` into fs2-cats

👍 

> Decide which typeclasses (especially Monad) to support, of cats, or of fs2?

My instinct is to support only cats typeclasses, with the exception of fs2's `Suspendable` and `Catchable` which have no equivalent in cats (yet ... they're talking about it).

------

So I'm wondering if this is too much to do for 0.3.1 … I was kind of thinking of just providing instances for fs2 `Task` and calling it good for now.

guersam:
Sounds reasonable. A little thought about cats/fs2 typeclass support:

1. From usability perspective, when dealing with `ConnectionIO` direct support for cats typeclasses would usually be the better bet.

2. Regarding maintainability, fs2 typeclasses will give us more consistency if we eventually will replace scalaz-stream with fs2 (#310).

3. I'm not sure about performance, because that whichever we choose to support for, it involves typeclass conversion anyway. I guess (and hope) the effect will be negligible though.

---

Breaking the changes smaller for 0.3.1 will also be fine. Although there's no official release for cats/fs2 yet, there's good reason for keep migration steps small enough. In my case, there are actually no big differences since I'm not using the APIs lower than `ConnectionIO` directly.

guersam:
BTW, is there a reason for the lack of automatic `Free` generation for postgres classes and jdbc drivermanager? I found it a bit tiring modifying all of them by hand, even for a small change.

tpolecat:
> is there a reason for the lack of automatic Free generation for postgres classes and jdbc drivermanager?

I think there was a reason but I don't remember. Opened #329 to follow up.

So for 0.3.1 maybe do the following and push off generalizing support for fs2 typeclasses

- Update to latest cats/fs2 (new cats out later tonight)
- Add doobie `Catchable` and `Capture` instances for fs2 `Task`

I really doubt we will get any actual cats+fs2 users since both libraries are in so much flux at the moment, so I think we have time to generalize later. I'd just like to get a release out that people can play with.

What do you think?

guersam:
Agreed. I'm willing to port to my scalaz-stream codebase to cats/fs2 to get rid of mixed imports from scalaz and cats. I'll handle #329 first and then add missing instances for `fs2.Task`.

guersam:
After second trial, I still find it harder than expected to add a few `fs2.Task` instances, because that `fs2.util.Catchable` is already a monad so that it always ends up with dropping own `doobie.util.Catchable` and touch here and there to avoid ambiguous/diverging implicit caused by the compatibility utils.

In this nature, it's getting more and more natural to me to fully support `fs2.util.Suspendable` from the initial cats/fs2 release, but if it feels like a too big change for 0.3.1 I'm okay to wait for 0.4.0-M1.

tpolecat:
Ok, I will take a look tomorrow and see if I run into the same issue.

guersam:
Latest build failures seem like the cache problem anticipated in #318. @tpolecat Could you flush the cache for this PR once?

tpolecat:
I don't know how to flush the cache but I restarted that build.

guersam:
Cache settings will appear in the travis repo page: https://docs.travis-ci.com/user/caching/#Clearing-Caches

Tried to do this by myself but I don't have the permission. Wish that travis allows PR openers to clear their PR specific cache.

tpolecat:
Ah ok, thanks. Cleared and restarted the build.

tpolecat:
Looks like it worked.

guersam:
Thanks :)

project/FreeGen.scala
```diff
@@ -152,17 +154,17 @@ class FreeGen(managed: List[Class[_]], log: Logger) {
   // Ctor values for all methods in of A plus superclasses, interfaces, etc.
   def ctors[A](implicit ev: ClassTag[A]): List[Ctor] =
     methods(ev.runtimeClass).groupBy(_.getName).toList.flatMap { case (n, ms) =>
-      ms.toList.zipWithIndex.map {
+      ms.toList.sortBy(_.getGenericParameterTypes.map(toScalaType).mkString(",")).zipWithIndex.map {
```
tpolecat: Thanks for this change, I should have done this a long time ago. Stuff was getting shuffled around every time I ran freegen.

tpolecat:
Ok, I think this looks great. Huge amount of work. I'm fine with `foldMapUnsafe` for now; we can add the other constraint later if we want to.

Is there anything else that needs to be done in this PR before merging? I'd prefer to get this in and then follow up as needed rather than keep this open longer. What do you think?

adelbertc:
I would hold off on adding the `RecursiveTailRecM` stuff, I'm becoming convinced that maybe we should just have it on `Monad` proper. I don't really like the marker trait and reading about the experience here made me reopen this https://github.com/typelevel/cats/issues/1278

tpolecat:
@adelbertc thanks, that was my instinct too. Given the volatility we should do the simplest thing that works.

guersam:
Although some docs (especially the book) should be updated to catch up these changes, I'd like merging this for now too.

tpolecat:
Alright let's do it.

macalinao:
How does this work with `monix.Task`? The docs mention this briefly but I get the following error:

```
[error] could not find implicit value for evidence parameter of type fs2.util.Catchable[monix.eval.Task]
[error]     DriverManagerTransactor[Task](
```

adelbertc:
@macalinao You will need to define an instance of `Catchable[monix.eval.Task]`. `Catchable` looks like:

```scala
trait Catchable[F[_]] extends Monad[F] {
  /** Lifts a pure exception in to the error mode of the `F` effect. */
  def fail[A](err: Throwable): F[A]
  /** Provides access to the error in `fa`, if present, by wrapping it in a `Left`. */
  def attempt[A](fa: F[A]): F[Attempt[A]]
}
```

tpolecat:
You'll need instancs of `fs2.util.Catchable` and `fs2.util.Suspendable`.

macalinao:
Cool. For future readers, here's my typeclass:

```scala

  implicit object taskCatchable extends Catchable[Task] with Suspendable[Task] {
    def pure[A](a: A): Task[A] = Task.pure(a)

    def flatMap[A, B](a: Task[A])(f: A => Task[B]): Task[B] =
      a.flatMap(f)

    def fail[A](err: Throwable): Task[A] =
      Task.raiseError(err)

    def attempt[A](fa: Task[A]): Task[Attempt[A]] =
      fa.materialize.map(_.toEither)

    def suspend[A](fa: => Task[A]): Task[A] =
      Task.suspend(fa)
  }
```

PR #327, opened 2016-08-18
Avoiding the deprecated <| operator in the Meta example

heliocentrist:
The compiler now complains about <| being deprecated. This version of the example should work well.


tpolecat:
Thanks! Can you switch the target branch to `series/0.3.x`? It will probably want you to rebase. I need to make a contributor guide so this is more clear.

heliocentrist:
Looks like this PR has conflicts with `series/0.3.x`. I've created a new one from that same branch.

Issue #303, opened 2016-07-11
String is not coercible to VARCHAR when using Enum types


veegee:
Create an Enum as follows:

```
CREATE TYPE support_request_type AS ENUM ('Feedback', 'ContactUs');
```

``` scala
implicit val SupportRequestTypeAtom = pgEnum(SupportRequestType, "support_request_type")

object SupportRequestType extends Enumeration {
  val Feedback, ContactUs = Value
}
```

Then query:

``` scala
def insertContactUsQuery(senderId: UUID, requestType: SupportRequestType.Value, message: String): Update0 =
    sql"INSERT INTO contact_us (sender_id, request_type, message) VALUES ($senderId, $requestType, $message)".update
```

But query checking fails:

```
INSERT INTO contact_us (sender_id, request_type, message) VALUES (?, ?, ?)

  ✓ SQL Compiles and Typechecks
  ✓ P01 UUID    →  OTHER   (uuid)

  ✕ P02 String  →  VARCHAR (support_request_type)
    - String is not coercible to VARCHAR (support_request_type) according to the JDBC
      specification. Fix this by changing the schema type to OTHER, or the Scala type
      to String.
  ✓ P03 String  →  VARCHAR (text)
```

Using Doobie 0.3.0

PR #331, opened 2016-08-31
wip: add scalatest module

### LINKED ISSUE #350: scalatest support
wip on #331

KadekM:
This is just the most naive port of Specs2 stuff to Scalatest 

and option to ignore checking of some columns, because in our codebase we use postgres enums and they check fails for them even when query is valid. There is hacky way I extract column name using regex, so if someone could check it out ... I can also remove this functionality if it's desired to match specs2 stuff 1:1.

Lets see if it builds ...

resolves #350


KadekM:
I guess I should not have updated that readme... since it's only on snapshot and it's going to be without contrib part anyways...

tpolecat:
I don't have time to look at this right now but I'll try to get to it by the end of the week. Thanks!!

KadekM:
@tpolecat is there intention to get his merged? Anything I should do/change ? : )

tpolecat:
Sorry, thanks for the ping. Yes, I want to get this in for 0.3.1 … I will add an issue to be sure it's tracked. You will need to rebase but wait until I have a chance to review it.

KadekM:
the code to ignore column is only due to #303 , I'm happy to remove it once we have it fixed or replace with something more sensible than regex :-)

PR #333, opened 2016-09-01
WIP: cats 0.7

KadekM:
I tried to see how it would look if we updated cats to 0.7...

I didn't spend much time on it but from quick peek I'm not exactly sure what the right way with last two remaining errors regarding `cats.RecursiveTailRecM[M]` (any tips?). I'll come back to it later and give it a try.


tpolecat:
Ugh, yeah this is going to suck. We need to add `RecursiveTailRecM` as a constraint everywhere we have `[M[_]: Monad: Catchable: Capture]` which means a ton of new yax directives.

I'll experiment and see how bad this is going to be. There are 5491 instances but most are in generated code.

KadekM:
I kind of hoped there would be cleaner way than spread YAX all over the place :/

tpolecat:
Yeah I need to drink some bourbon and think about our options.

I really appreciate your work on this. Many many thanks.

KadekM:
I pushed some yax that I had started beforehand...

guersam:
I've chosen plain `Monad` + `Free#foldMapUnsafe` instead of adding `RecursiveTailRecM` constraint everywhere for easier compatibility. Is it one of the hard requirement of our algebra to ensure 100% stack safety anyway?

guersam:
Here's what's done regarding this in #323: https://github.com/tpolecat/doobie/pull/323/commits/19e77c7d6f42b434578a354ccd4ea11929530951

tpolecat:
Ok sorry, I haven't been keeping up closely. @guersam it looks like this work is subsumed by your work in progress on #323, is that correct?

guersam:
Right, it includes cats 0.7.0 + fs2 0.9.0-RC2 support.

>

KadekM:
Great :) I'll close this then.

btw latest now is 0.7.2

guersam:
Didn't know about that, thanks for pointing it out!

PR #337, opened 2016-09-12
Readme fix, minor version bump

KadekM:
I've noticed README shows old version... and I've also bumped versions... that should be fine I guess


tpolecat:
Thanks! I'll update the snapshot momentarily.

tpolecat:
I think the 2.11 snapshot is up but the others aren't. I'm in England now and it's taking forever to publish so I gave up. I'll be back in the US in a few days.

PR #341, opened 2016-09-17
coursier

tpolecat:
This swaps in coursier for dependency management, which seems to be a whole hell of a lot faster than ivy. The question is whether or not the Travis caching is set up correctly; I'm not quite sure how that bit works. @guersam does this look right?


guersam:
Yeah it looks fine and I like coursier a lot. The reason that I didn't include it at first was the possibility of conflict between the project-wise and the global plugin configs, but it won't be a big deal when it actually happens. :+1:  and recommend flushing all travis cache before merging it at well.

.travis.yml
```diff
@@ -38,6 +38,7 @@ cache:
   - $HOME/.sbt/cache
   - $HOME/.sbt/launchers
   - $HOME/.ivy2
+  - $HOME/.coursier
 
 before_cache:
```
tpolecat: Done.

tpolecat:
Ok I made that change and flushed the caches and restarted the builds.

PR #352, opened 2016-09-27
fs2 0.9.0 -> 0.9.1

KadekM:
... just being nice to people who are getting fs2 only via transitive dependency ...


tpolecat:
thanks!

PR #353, opened 2016-09-27
bump h2, postgresql and hikariCP versions

KadekM:
Since doobie 0.3 depends on jdk8+ , can't we use hikariCP for java8 for 0.3.1 ?

### DIFF
--- a/yax/postgres/src/main/scala/doobie/postgres/free/pgconnection.scala
+++ b/yax/postgres/src/main/scala/doobie/postgres/free/pgconnection.scala
@@ -19,6 +19,8 @@ import doobie.util.capture._
 
 import java.lang.Class
 import java.lang.String
+
+import org.postgresql.util.PGobject
 import org.postgresql.PGConnection
 import org.postgresql.PGNotification
 import org.postgresql.copy.CopyManager
@@ -83,7 +85,7 @@ object pgconnection {
 
     // Primitive Operations
     case class  AddDataType(a: String, b: String) extends PGConnectionOp[Unit]
-    case class  AddDataType1(a: String, b: Class[_]) extends PGConnectionOp[Unit]
+    case class  AddDataType1(a: String, b: Class[_ <: PGobject]) extends PGConnectionOp[Unit]
     case object GetBackendPID extends PGConnectionOp[Int]
     case object GetCopyAPI extends PGConnectionOp[CopyManager]
     case object GetFastpathAPI extends PGConnectionOp[Fastpath]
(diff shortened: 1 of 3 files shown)


tpolecat:
Ah good catch. Thanks!

tpolecat:
It looks like the Postgres driver may have added a new JDBC type. Odd that it only turned up in the 2.11 build. I'm going to flush the caches and rebuild.

KadekM:
Yea the fail only on scala 2_11 weirded me out, my first thought was that
we run Analysis tests only on 2_11 but it's hard to check from cellphone :)

tpolecat:
Yeah it looks like there's a driver change. We'll need to investigate.

KadekM:
I've gathered all changes between driver versions (1201, 1211] in chronological order (on top of file changes from 1202, then 1203 etc...):
https://gist.github.com/KadekM/4318774344d90f68f132fb62a816d642

First file is narrowed down to what I guess may interest us, second file is all PRs unfiltered (including style, tests etc...)

I thought it may help us.

KadekM:
It seems to have been refcursor

yax/postgres/src/main/scala/doobie/postgres/free/pgconnection.scala
```diff
@@ -83,7 +85,7 @@ object pgconnection {
 
     // Primitive Operations
     case class  AddDataType(a: String, b: String) extends PGConnectionOp[Unit]
-    case class  AddDataType1(a: String, b: Class[_]) extends PGConnectionOp[Unit]
+    case class  AddDataType1(a: String, b: Class[_ <: PGobject]) extends PGConnectionOp[Unit]
```
tpolecat: Why is this upper bound needed? Did the driver change the `.addDataType` method?

tpolecat:
Do you know which test was failing due to the missing `RefCursor` constant? I don't recall writing a test that uses it.

KadekM:
Everything in AnalysisTestSpec...

```
  check(AnalysisTest.speakerQuery2)
  check(AnalysisTest.arrayTest)
  check(AnalysisTest.pointTest)
  check(AnalysisTest.update2)
```
so even the simplest queries that have nothing to do with it... i.e. this would fail (but only the check of AnalysisSpec)

```
sql"""SELECT C.NAME FROM COUNTRYLANGUAGE""".query[String]
```

tpolecat:
Ah I gotcha, the problem is when it's loading up driver metadata. The database is saying it supports a mapping for `RefCursor` and we didn't recognize it. Makes sense now.

KadekM:
I see. Thanks for info :)

tpolecat:
Anyway thanks for this! 👍

yax/postgres/src/main/scala/doobie/postgres/free/pgconnection.scala
KadekM: Yes they did.
`void addDataType(String type, Class<? extends PGobject> klass) throws SQLException;`

KadekM: Before it was just `Class<?>`

PR #354, opened 2016-09-28
WIP - simple logging

### LINKED ISSUE #292: Expose a mechanism for instrumenting query with timing
In #267 @tpolecat was nice enough to help me figure out how to instrument my query so I could determine how long it takes to execute. In gitter, @tpolecat suggested I open an issue about a more generalized way to get the same effect.

For bonus points, I'd sure love to be able to log the final query that was generated along with the timing information.

tpolecat:
Ok so this adds statement logging for queries, which has long been desired and long been delayed. This isn't the best approach but I think it addresses the most common used case.

So, with this change you can indicate your desire to have a statement logged by saying `.queryL` rather than `.query`, or by providing a `LogHandler` when constructing a `Query/0` directly.

``` scala
import doobie.util.log.LogHandler._

def q(name: String, pop: Long) = 
  sql"""
    select code from country 
    where name like ${"%" + name + "%"} 
    and population > $pop"""
  .queryL[String](jdkLogHandler).list // using the provided log handler
```

With output like this:

```
scala> q("u", 1000000).transact(xa).unsafePerformIO

Sep 27, 2016 8:53:06 PM doobie.util.log$LogHandler$$anonfun$jdkLogHandler$1 apply
INFO: Successful Statement Execution:

  select code from country where name like ? and population > ?

 arguments = [%u%, 1000000]
   elapsed = 2 ms exec + 0 ms processing (2 ms total)

res12: List[String] = List(AUS, BEL, BTN, BGR, BFA, BDI, DOM, ECU, ZAF, GTM, GIN, GNB, HND, AUT, YUG, CAF, COD, KOR, CUB, KWT, LTU, MRT, MUS, MOZ, NIC, PNG, PRY, PER, PRT, PRI, SAU, SDN, CZE, TUN, TUR, TKM, HUN, URY, BLR, VEN, RUS)
```

The handler is given the following information:
- the raw SQL string
- the arguments (usually as an `HList`)
- time taken for first row to arrive (if successful)
- time taken to process all rows (if successful)
- the exception, if execution or processing failed

`LogHandler` is simply a wrapper of a function from `LogEvent[A] => Unit` where `A` is the composite type for statement parameters. You can provide your own easily:

``` scala
import doobie.util.log.LogHandler

def q(name: String, pop: Long) = 
  sql"""
    select code from country 
    where name like ${"%" + name + "%"} 
    and population > $pop"""
  .queryL[String](LogHandler(println)).list // just print out the event
```

Thus

```
scala> q("u", 1000000).transact(xa).unsafePerformIO

Success(
    select code from country 
    where name like ? 
    and population > ?,%u% :: 1000000 :: HNil,2,0)

res14: List[String] = List(AUS, BEL, BTN, BGR, BFA, BDI, DOM, ECU, ZAF, GTM, GIN, GNB, HND, AUT, YUG, CAF, COD, KOR, CUB, KWT, LTU, MRT, MUS, MOZ, NIC, PNG, PRY, PER, PRT, PRI, SAU, SDN, CZE, TUN, TUR, TKM, HUN, URY, BLR, VEN, RUS)
```

Open questions:
- Should this be provided implicitly to `.query` so there is no call-site cost to turning logging on and off? I don't like this idea but it might be more friendly.
- Should we try to provide a better built-in `LogHandler` or just assume that all users are likely to want to provide their own?

To be done:
- [x] update support
- [ ] ~~stream support~~ // later, it's too complicated
- [ ] chapter in book
- [x] scaladoc
- [ ] tests

Fixes #292 

/cc @coltfred @leifwickland @vzpaul


fanf:
That seems to be a solid proposition, thanks ! 
Regarding the questions (easy first):

> Should we try to provide a better built-in LogHandler or just assume that all users are likely to want to provide their own?

Most likelly people will provide their own. Even so, defaults are generally assumed to be "the correct way of doing things", even if it's documented otherelse (see: puppet server coming by default with a rails mod documented "you shall not used that in prod, it does not scale", and still having it deployed almost exclusivly everywhere). 
Just to say: it is better if the default log handler can be use as a general source of inspiration about the correct way of building a log handler, even if at the end, it it not the one used. 

> Should  this be provided implicitly to .query so there is no call-site cost to turning logging on and off? I don't like this idea but it might be more friendly.

I have mixed feeling here. I'm almost sure that I want to be able to configure a logger for all requests, even if by application configuration, it is configured to be a no-op logger. Because I want to be able to tell the app user "change that log parameter and send me the results", and so it must already be coded for. So most likelly, I will have the queryL everywhere, so more cluter (from my lib-user point of view). 
I'm almost pretty sure I want to be able to have different logger for different queries set.This is explicit with an explicit parameter, and less so with  "what is the implicit in scope for that query". 

So for now, I see pros and cons on both way. Why do you dislike the idea? Maybe there is clear args against it?

fiadliel:
I'm not a fan of a default logger implementation just because... SLF4J, Logback, Apache logging, JUL, the list goes on. Perhaps an example using one of these libraries could be in an optional artifact (I personally use SLF4J)?

Don't really care that much about an implicit logger parameter, though there are reasonable examples where people may want to have different logger input for every query, e.g. `.queryL(logWithTag("addUser"))`

coltfred:
@tpolecat 

I like the current proposition! Thanks for taking the time to look at this!

I think making the call to queryL explicit is a good idea. I personally would pass a `log.debug` type call in there and just adjust the logging level if I wanted the info.

I think using implicits would be a mess as you don't really want the same "logging" for each type of query, some might be more mission critical than others and you might want them logged all the time. I find managing the implicit scope more mental tax.

>Should we try to provide a better built-in LogHandler or just assume that all users are likely to want to provide their own?

What would the default be? I can't think of a sensible answer if the logging nightmare than is the JVM.

tpolecat:
Thanks for the comments. I will provide an example `LogHandler` just so you have something you can use initially to see that it works, but maybe add `@deprecated("Example logger. Not intended for production use")` or something so it will be clear that you shouldn't use it in real life.

brianpendleton:
> Should this be provided implicitly to .query so there is no call-site cost to turning logging on and off? I don't like this idea but it might be more friendly.

I'm still a newbie to doobie 😉, but this pattern actually comes up a lot at my work...

At work our current method (in java) is to have a standard "db" logger that we put into a base DBI class that everyone extends.  It hooks into the connection context and does logging of slow queries.  A bit hacky, but, the nice thing is I can flip a switch in a config for db to set to DEBUG level to get metrics around queries/results.  Everyone extends from that class, so your application config just needs to specify a DB logging level to get it.

If the parameter was implicit here, I could get that same effect for free.  Just have people mixin standard db logging trait that would put an implicit logger in scope that can be configured.

That said, I don't mind being explicit, especially when logging statements will be looked at by non-developers and then they may try to track down where those logging statements came from in your code.  Having explicit logging parameters would ease in the "how did this query actually get logged" scenario.  Especially if you name a logger "query1-db", then someone can quickly look for where you have a `val queryLogger = getLogHandler("query1-db")` and find out the exact query it was sent too.

I'd lean towards explicit.  Swapping out my `.query` to `.queryL` just sounds like a Friday morning chore over some coffee.

yax/core/src/main/scala/doobie/util/log.scala
```diff
@@ -0,0 +1,109 @@
...
+        case Success (s, a, m1, m2)         => Success(s, f(a), m1, m2)
+        case ProcessingFailure(s, a, m1, t) => ProcessingFailure(s, f(a), m1, t)
+        case ExecFailure(s, a, t)           => ExecFailure(s, f(a), t)
+      }
+
+  }
+  object LogEvent {
+
+    case class Success          [A](sql: String, args: A, msExec: Long, msProcessing: Long) extends LogEvent[A]
+    case class ProcessingFailure[A](sql: String, args: A, msExec: Long, failure: Throwable) extends LogEvent[A]
```
leifwickland: I wouldn't mind a comment on this class explaining that it indicates failure in the user's code when processing the result of a query. Initially I assumed that this was an error coming out of JDBC when preparing the query.

leifwickland:
@tpolecat Thanks so much for doing this.

I agree with what other folks have said about a default handler; I don't think in the JVM's nightmare logging nightmare that you can provide a reasonable default. An example would be swell. Using the `@deprecated` mechanism to provide that example is a little weird, but probably not the worst.

I wouldn't mind having an example/default function provided by Doobie that converted `LogEvent[A] => String`. In my mind, that's the barrier to understanding how to use Doobie's logging mechanism. If you provide that method, and then an example that creates the `LogHandler` with it, somebody ought to be able to get from there to done reasonably easily. Helping bridge the gap between an `HList` and normal humans would be huge.

yax/core/src/main/scala/doobie/util/log.scala
tpolecat: Yep I need to go back and write doc. Will do.

tpolecat:
@leifwickland yeah the `HList` thing is complex and just one possibility, but that's what you get with the `sql` interpolator. It's important to parameterize on the input type so loggers can inspect the arguments.

So yeah I'll see if I can make the "roll your own" option as simple as possible.

tpolecat:
@brianpendleton thanks for your comments, that gives me a few ideas.

leifwickland:
@tpolecat As I was thinking about what `A => String` would look like, I realized that it'd probably need to take `implicit ev: ToTraversable.Aux[A, List, L]`, which would kind of defeat the intention of shielding the user from `HList`, so it could be that an example is the best you can do.

Maybe you could hide `HList` by converting it to a `List[String]` in a simplified form of the API.

```
object LogHandler {
  def simple[A](f: LogEvent[List[String]] => Unit)(implicit ev: ToTraversable.Aux[A, List, L]): LogHandler[A] = {
    // ...
  }
}
```

I think that'd let a Shapeless naif implement:
```
case Success(s, a, e1, e2) =>
  jdkLogger.info(s"""Successful Statement Execution:
    |
    |  ${s.lines.dropWhile(_.trim.isEmpty).mkString("\n  ")}
    |
    | arguments = [${a.mkString(", ")}]
    |   elapsed = $e1 ms exec + $e2 ms processing (${e1 + e2} ms total)
  """.stripMargin)

// etc...
```

:shrug:

yax/core/src/main/scala/doobie/util/log.scala
```diff
@@ -0,0 +1,109 @@
...
+      this match {
+        case Success (s, a, m1, m2)         => Success(s, f(a), m1, m2)
+        case ProcessingFailure(s, a, m1, t) => ProcessingFailure(s, f(a), m1, t)
+        case ExecFailure(s, a, t)           => ExecFailure(s, f(a), t)
+      }
+
+  }
+  object LogEvent {
+
+    case class Success          [A](sql: String, args: A, msExec: Long, msProcessing: Long) extends LogEvent[A]
```
tpolecat: Good idea. I guess I could also use `System.nanoTime` since these values end up rounding to 0ms pretty often.

yax/core/src/main/scala/doobie/util/log.scala
```diff
@@ -0,0 +1,109 @@
...
+    case class ProcessingFailure[A](sql: String, args: A, msExec: Long, failure: Throwable) extends LogEvent[A]
+    case class ExecFailure      [A](sql: String, args: A,               failure: Throwable) extends LogEvent[A]
```
tpolecat: Yeah I guess that's not a bad idea. 👍

tpolecat:
@leifwickland I made your suggested changes regarding timing measurements. Thanks for the suggestions, I think it's better.

tpolecat:
@leifwickland (fixed the `ms` thing in c5b63fc)

tpolecat:
My benchmarks are showing that a no-op log handler has no measurable effect on execution time so I think I may simplify and say ok there's always a handler, rather than making it optional. The fast path isn't any faster.

tpolecat:
Ok, so I changed the logger to be an argument to `.query` and `.update` and got rid of `.queryL` and `.updateL`. The argument defaults to `LogHandler.nop` which does nothing. In addition it's `implicit` so *if you feel like it* you can provide an implicit logger that will get picked up. Or you can pass it manually or mix and match. Up to you.

brianpendleton:
Love it.  :+1:

tpolecat:
In the interest of getting this merged so people can play with it I'm not going to worry about support it for `.process` right now.

tpolecat:
Continuing on #388

PR #356, opened 2016-10-03
add adopters

KadekM:


brianpendleton:
👍 , we're using in the `fintech` group at Circle.  Managing FX pricing feeds and market data aggregation.  I plan to roll it out as I replace more of our stack with Scala

tpolecat:
I have mixed feelings about this kind of thing … I like the code to speak for itself. But as you point out it's a common practice and can help people worry less. I'm going to think about it.

mdmoss:
I think this could be a good thing. Knowing other businesses are users might make it more attractive to adopt, especially when people who are less technical involved in making the decision. It would have smoothed the way for my team a little in the beginning.

KadekM:
I'm with @mdmoss , very often, if not always, one of first questions raised is "who uses it", at least from my experience.

tpolecat:
Yeah, I know they're using it at some big impressive-sounding companies but I'm not confident that they would want their names used. So if it's a list of companies nobody has heard of it could have a negative net effect. Dunno. Thanks for the input though, I'm on the fence.

KadekM:
Sounds reasonable. I didnt think of possible negative net effect, but yea,
then it may be better not include it at least for now.

Issue #357, opened 2016-10-05
Please make Postgis optional


gwils:
Currently anyone wanting to use Postgres with doobie must also have Postgis dependencies pulled in, even if they aren't going to use Postgis. I would prefer not to have a dependency on Postgis since it is LGPL.
Having even a transitive LGPL dependency makes it harder or even impossible for me to convince the powers that be to use doobie in a particular project I'm working on. There are probably other people in similar positions, or who have other reasons (for example JAR size) for not wanting this additional dependency unnecessarily.

PR #359, opened 2016-10-05
separate postgis via provided

KadekM:
Addresses #357 by separating postgis stuff into new object.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -247,6 +247,8 @@ lazy val example_cats = project.in(file("modules-cats/example"))
 /// POSTGRES
 ///
 
+val postgisDep = "org.postgis" % "postgis-jdbc" % "1.3.3"
+
 def postgresSettings(mod: String): Seq[Setting[_]] =
   doobieSettings  ++
   publishSettings ++ Seq(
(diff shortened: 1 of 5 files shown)


tpolecat:
This seems like a much lighter approach than #358 so I'm inclined to take this route. What do you think?

KadekM:
I'm all for simpler (this) solution. We can always split it later if needed...
I'll update doobie book in next days... Is there anything in-code I can change to make it more intuitive/better for end users?

tpolecat:
This looks great. Let's get the doc updated and we're good. I'm going to close #358 in favor of this one.

KadekM:
@tpolecat I had added little bit to book of doobie... do you think it's sufficient?

gwils:
Looks good to me.

tpolecat:
This is blocked on #362 but otherwise ready to go.

tpolecat:
Ok I think we're good to go here. Can you rebase?

KadekM:
Rebased.

tpolecat:
Brilliant! Thanks for all your work on this.

PR #360, opened 2016-10-05
Add Instant and LocalDate meta instances

fiadliel:
Since Doobie 0.3.x is targeting JDK 1.8+, we can probably start depending on its API?

Meta instances for `java.time.Instant` and `java.time.LocalDate` are trivial to add, and have a single unambiguous representation.


tpolecat:
Perfect. Thank you!

PR #363, opened 2016-10-09
Run tut in Travis builds (fix #362)

### LINKED ISSUE #362: CI doesn't seem to be building tut doc
Hey @guersam could you take a look at this? Here is a [recent build](https://travis-ci.org/tpolecat/doobie/jobs/165525653) that claims to be building docs, but there's nothing in the transcript to indicate that this is happening. The log message appears but everything after the `&&` [here](https://github.com/tpolecat/doobie/blob/series/0.3.x/project/travis-build.sh#L16-L19) looks like it's ignored.

guersam:
It seems that the first `exec` replaces the shell and swallows the following ones. [1]  This PR just merges all sbt commands into a single run.

[1] http://stackoverflow.com/a/7194536

### DIFF
--- a/yax/core/src/test/scala/doobie/syntax/catchable.scala
+++ b/yax/core/src/test/scala/doobie/syntax/catchable.scala
@@ -24,7 +24,13 @@ object catchablespec extends Specification {
     }
 
     "work on unaliased IConnection" in {
+#+scalaz
+      42.pure[ConnectionIO].map(_ + 1).attempt
+#-scalaz
+#+cats
+      // TODO Remove yax (https://github.com/tpolecat/doobie/issues/369)
       (42.pure[ConnectionIO].map(_ + 1): ConnectionIO[Int]).attempt
+#-cats
       true
     }
 
--- a/yax/core/src/test/scala/doobie/syntax/catchable.scala
+++ b/yax/core/src/test/scala/doobie/syntax/catchable.scala
@@ -5,7 +5,9 @@ import scalaz.syntax.monad._
 import scalaz.syntax.catchable._
 #-scalaz
 #+cats
-import doobie.util.catchable._
+#+fs2
+import fs2.interop.cats._
+#-fs2
 import cats.implicits._
 #-cats
 
@@ -22,7 +24,7 @@ object catchablespec extends Specification {
     }
 
     "work on unaliased IConnection" in {
-      42.pure[ConnectionIO].map(_ + 1).attempt
+      (42.pure[ConnectionIO].map(_ + 1): ConnectionIO[Int]).attempt
       true
     }
 
--- a/yax/docs/src/main/tut/03-Connecting.md
+++ b/yax/docs/src/main/tut/03-Connecting.md
@@ -24,8 +24,8 @@ import doobie.imports._
 import scalaz._, Scalaz._
 #-scalaz
 #+cats
-import doobie.util.compat.cats.monad._ // todo: make this automatic
 import cats._, cats.data._, cats.implicits._
+import fs2.interop.cats._
 #-cats
 ```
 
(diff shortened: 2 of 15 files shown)


guersam:
Confirmed that `tut` is running again, sorry for missing it from the first place. Will fix the doc itself within a few days.

tpolecat:
Cool thanks for looking into this.

yax/core/src/test/scala/doobie/syntax/catchable.scala
```diff
@@ -22,7 +24,7 @@ object catchablespec extends Specification {
     }
 
     "work on unaliased IConnection" in {
-      42.pure[ConnectionIO].map(_ + 1).attempt
+      (42.pure[ConnectionIO].map(_ + 1): ConnectionIO[Int]).attempt
```
guersam: Additional indirection confuses type inference here

yax/docs/src/main/tut/03-Connecting.md
```diff
@@ -120,7 +120,7 @@ program3a.replicateM(5).transact(xa).unsafePerformIO.foreach(println)
 #-scalaz
 #+cats
 ```tut
-program3a.replicateA(5).transact(xa).unsafePerformIO.foreach(println)
+ApplicativeError[ConnectionIO, Throwable].replicateA(5, program3a).transact(xa).unsafePerformIO.foreach(println)
```
guersam: And here.

guersam:
Tried removing cats-only `attempt` syntax in favor of `cats.ApplicativeError` due to an ambiguous implicit error when `import cats.implicits._` and `fs2.interop.cats._` is used.

The problem is that the type cannot be inferred in some situation due to the additional implicit search step(s) as `ApplicativeError` is derived via the conversion from `fs2.util.Catchable` to `cats.MonadError` provided by fs2-cats. For example, even the simple test case `42.pure[ConnectionIO].map(_ + 1).attempt` doesn't work whereas `42.pure[ConnectionIO].attempt` works.

UPDATE: Not only `ApplicativeError` but other type classes such as `Applicative` suffer from it. Do you have any idea?

tpolecat:
Yeah I ran into this and didn't try to track it down. I won't have time to look at this until the weekend but I'll have a couple weeks with a lot of free time because I'll be in Chile by myself with nothing better to do.

In any case thanks for your help with this.

guersam:
Seems related to https://github.com/typelevel/cats/issues/1210

guersam:
And this https://github.com/scala/scala/pull/5106

yax/core/src/test/scala/doobie/syntax/catchable.scala
tpolecat: Let's leave this test as-is and yax it out for cats, then open a separate issue to sort this out. I know Adelbert is pretty focused on it so maybe he'll have some ideas.

tpolecat:
If you don't mind making that one change in the tests I think this is ok to merge. Opened #369 to track the underlying issue.

guersam:
Old postgis driver (1.3.3) is compiled against postgres driver 8.3 and [pulls it as a transitive dependency](https://travis-ci.org/tpolecat/doobie/jobs/168011798#L7404-L7407). Bumped it to the latest version.

yax/core/src/test/scala/doobie/syntax/catchable.scala
```diff
@@ -22,7 +24,13 @@ object catchablespec extends Specification {
     }
 
     "work on unaliased IConnection" in {
+#+scalaz
       42.pure[ConnectionIO].map(_ + 1).attempt
+#-scalaz
+#+cats
+      // TODO Remove yax (https://github.com/tpolecat/doobie/issues/369)
+      (42.pure[ConnectionIO].map(_ + 1): ConnectionIO[Int]).attempt
+#-cats
```
guersam: Forgot to ping @tpolecat, is this what you mean?

tpolecat: Yes, perfect. Thanks!

PR #366, opened 2016-10-15
Revert "Add query checker for scalatest"

tpolecat:
Reverts tpolecat/doobie#361

@Milyardo oops, we already had an open PR for this as #331 … I need to look more closely and see what the differences were.


KadekM:
@tpolecat I think this is a bit cleaner than my stuff... I have taken the code straight from our internal codebase, because some scalatest support is better than none, but I think this seems more fitting (as it does not have some stuff we need)

tpolecat:
@KadekM ok, sounds good.

@Milyardo I will undo the undo. Sorry for the trouble/confusion.

Issue #200, opened 2015-07-21
Derive Composite[Option[A]] given Composite[A]?


tpolecat:
Can we generalize deriving of `Option[A]` here, for helping with outer joins where you'll have a column vector that is all null or all non-null? 

``` scala
import scalaz._, Scalaz._
import doobie.imports._

case class Company(name: String, address: String)

// Map a Company to a pair of Option[String] columns
implicit val CompositeOptionCompany = 
  Composite[(Option[String], Option[String])].xmap[Option[Company]](
  { case (a, b) => (a |@| b)(Company.apply) }, _.foldMap(c => (some(c.name), some(c.address))))
```

A problem with this is, if you had `Company(name: Option[String], address: Option[String])` then you can't distinguish `Some(`Company(None, None))`from`None` because they have the same column representation in the join.

/cc @ramnivas

ramnivas:
Looks like the problem you mentioned can come up easily. I tried to use this workaround in a real project which have ids defined as `Option[Long]`. So mapping this to `Company`:
```
case class Company(name: String, address: String, id: Option[Long])
```

When I update the mapping as follows:
```
  implicit val CompositeOptionCompany =
    Composite[(Option[String], Option[String], Option[Option[Long]])].xmap[Option[Company]](
    { case (a, b, id) => (a |@| b |@| id)(Company.apply) },
    _.foldMap(c => (some(c.name), some(c.address), some(c.id))))
```

I start to get the error back:
```
Error:(14, 14) Could not find or construct Composite[(Option[String], Option[String], Option[Option[Long]])].
    Composite[(Option[String], Option[String], Option[Option[Long]])].xmap[Option[Company]](
             ^
```

tpolecat:
Yep, ok. This is an ancient problem going back to the origin of SQL ... there is an infinite tower of `NULL` and there is no way to tell them apart. So my earlier suggestion of selecting the primitive column types and then using `.map` might end up being what you need to do, using a non-nullable field like `name` to decide whether the join failed or not (or by adding an explicit discriminant column).

ramnivas:
Yes. I am going with the `.map` approach for now. 

Separately, I am considering making `id`s simple `Long` (or other primitive types such as `String` as appropriate). Once I do that in many cases, the implicit `Composite` mapper approach will become useful (manually added or when you implement it, automatically derived).

ramnivas:
[...]

PR #372, opened 2016-10-18
Add automatic derivation of Composite[Option[A]] for A: Composite

jto:
Following our discussion on https://github.com/tpolecat/doobie/issues/200, here's a first draft of semi-automatic derivation (sorry for the delay, I had less time than I expected last week...).

See the example in test. The user has to explicitly pass the "empty" value. It's a bit annoying because it's only necessary if every fields in A is an Option. The alternative would be to just use None in that case I guess ? In which case we could automatically derive everything.

### DIFF
--- a/yax/core/src/main/scala/doobie/util/composite.scala
+++ b/yax/core/src/main/scala/doobie/util/composite.scala
@@ -34,10 +34,10 @@ import java.sql.ParameterMetaData
  */
 object composite {
 
-  @implicitNotFound("""Could not find or construct Composite[${A}]. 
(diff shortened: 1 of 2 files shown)


tpolecat:
I'm finally having a look at this (one thousand apologies) and I was thinking maybe always derive it if there's at least one non-`Option` field, otherwise require the user to explicitly pull in an instance that gives you either `None` or a record full of `None`s, both of which we could also provide automatically. I'll mess around with this a bit.

jto:
No problem :) Let me know if you need anything.

tpolecat:
Closed via #540, thanks for your ideas here!

Issue #282, opened 2016-04-11
Add Composability with other Free Algebras


beefyhalo:
I'd like to compose doobie programs with other free algebras "data a la carte" style (using `scalaz.Inject`).

peterneyens:
@tpolecat I looked into your [suggestion](https://gitter.im/tpolecat/doobie?at=57f173f8d38f186520bbbe1c) to @raulraja of using `transK` to get transactions when combining Free algebras where some of them can be interpreted to `ConnectionIO`.

I created a [proof of concept](https://github.com/peterneyens/freedoobie) (using the _doobie-core-cats_ snapshot). In short there are 2 interpreters for 2 ADTS, one to `IOLite`, one to `ConnectionIO` which are both lifted to return a `Kleisli[IOLite, Connection, ?]`.

To get transactions for `Kleisli[M, Connection, ?]`, I essentially need to recreate something like a `Transactor` but for `Kleisli[M, Connection, ?]` (which I dubbed `ConIOK[M, ?]` in my example). It would be nice if I could reuse the `connect` method from `Transactor` to get a `M[Connection]`, but is protected (in order to not leak connections I suppose). In my example I created a `Transactor` subclass which exposes the `connect` method as `connectHack` as a hacky workaround.

I am interested in your opinion about this technique.

Like you mentioned, Interleaving of effects can be difficult / dangerous ("how to call back the missiles in case of a rollback"): I could require a type class instance of something like `ReExecutable[F[_]]` / `CanBeInterleaved[F[_]]` before lifting an `F ~> M` interpreter to `F ~> ConIOK[M, ?]` so you don't end up shooting yourself in the foot.

wedens:
it is solved by new transactor/interpreters design, isn't it?

tpolecat:
Somehow I missed this. There is a [coproduct example](https://github.com/tpolecat/doobie/blob/series/0.4.x/yax/example/src/main/scala/example/Coproduct.scala) that addresses this question I think.

beefyhalo:
Nice! Marking as closed. FWIW, I forgot about this "issue"... and after gaining some experience, I don't think I needed this all along.

PR #377, opened 2016-10-31
Make Transactor methods package protected

peterneyens:
Would you be open to making the methods of `Transactor` package protected (`protected[doobie]`) ?

That would make it possible to create a small doobie extension which provides transactions for `Kleisli[M, Connection, ?]` by reusing `Transactor`.
```scala
def execute[M[_]: Monad: Catchable: Capture, A](
  ciok: Kleisli[M, Connection, A],
  xa: Transactor[M]
): M[A] = {
  def safe(ma: Kleisli[M, Connection, A]): Kleisli[M, Connection, A] = {
    val before = xa.before.transK[M]
    val after  = xa.after.transK[M]
    val oops   = xa.oops.transK[M]
    val always = xa.always.transK[M]

    (before *> ma <* after) onException oops ensuring always
  }

  xa.connect flatMap safe(ciok).run
}
```

This extension could then also provide the functionality to combine multiple interpreters, so we could have only one transaction for a program expressed using a combination of algebras (see also my comment on #282) :

```scala
type App[x] = Coproduct[FooOp, Coproduct[BarOp, BazOp, ?], x]

val interp1:`FooOp ~> ConnectionIO = ???
val interp2: BarOp ~> Task = ???
val interp3: BazOp ~> ConnectionIO = ???

// combine and lift interp1, interp2 and interp3
val appInterp: App ~> Kleisli[Task, Connection, ?] = ???

val program: Free[App, Int] = ???
val kl: Kleisli[Task, Connection, Int] = program.foldMap(appInterp)
val task: Task[Int] = execute(kl, xa)
```


tpolecat:
👍  I don't have a problem doing this for 0.3 but just FYU `Transactor` will probably become a multi-parameter typeclass in 0.4 so any custom work will probably need to be redone. Should be much easier in any case.

There was another PR in line ahead of you, sorry. Can you rebase?

PR #384, opened 2016-11-18
composable query fragments

tpolecat:
This is kind of embarrassing. The `Builder` type produced by the `sql` interpolator is trivially composable but somehow I never noticed, so this enables concatenating them together. `Builder` has been renamed `Fragment` in the process.

The end result is that we can construct little query hunks with interpolated args and then add them up. They have fancy types so you can't put them in a list and sum them generically, but for the common use case where you want to swap out `WHERE` clauses or whatever it seems useful and safer than dealing with raw strings.

```scala
case class Info(name: String, code: String, pop: Int)

def select(name: Option[String], pop: Option[Int]): ConnectionIO[List[Info]] = {
  val sel = sql"select name, code, population from country"
  val f1  = name.map(s => sql"name like $s")
  val f2  = pop.map(n => sql"population > $n")
  val q = (f1, f2) match {
    case (None, None)         => sel
    case (None, Some(w))      => sel + sql"where" + w
    case (Some(w), None)      => sel + sql"where" + w
    case (Some(w1), Some(w2)) => sel + sql"where" + w1 + sql"and" + w2
  }
  q.query[Info].list
}
```

Still need to do:
- [x] scaladoc
- [x] book chapter
- [x] fix the stackframe hack
- [x] tests

... assuming this seems useful. WDYT?

/cc @guersam @leifwickland


yax/core/src/main/scala/doobie/syntax/string.scala
```diff
@@ -91,20 +97,72 @@ instance for each element in the REPL. See the FAQ in the Book of Doobie for mor
...
           val placeholders = List(t.foldMap(_ => 1))
-#-cats          
+#-cats
     }
 
   }
 
-  /** 
+  // If we have two composites of HLists we can concatenate them.
+  implicit class MoreCompositeOps[A <: HList](ca: Composite[A]) {
```
tpolecat: I guess this should probably go on `Composite`.

tpolecat:
```scala
scala> case class Info(name: String, code: String, pop: Int)
defined class Info

scala> val q = sql"select name, code, population from country"
q: doobie.syntax.string.Fragment.Aux[shapeless.HNil] = doobie.syntax.string$SqlInterpolator$sql$$anon$8@44e0583a

scala> def filter(s: String) = sql"where name like $s"
filter: (s: String)doobie.syntax.string.Fragment.Aux[shapeless.::[String,shapeless.HNil]]

scala> q.query[Info].check.unsafePerformIO

  select name, code, population from country

  ✓ SQL Compiles and Typechecks
  ✓ C01 name       VARCHAR (varchar) NOT NULL  →  String
  ✓ C02 code       CHAR    (bpchar)  NOT NULL  →  String
  ✓ C03 population INTEGER (int4)    NOT NULL  →  Int

scala> (q + filter("U%")).query[Info].check.unsafePerformIO

  select name, code, population from country where name like ?

  ✓ SQL Compiles and Typechecks
  ✓ P01 String  →  VARCHAR (text)
  ✓ C01 name       VARCHAR (varchar) NOT NULL  →  String
  ✓ C02 code       CHAR    (bpchar)  NOT NULL  →  String
  ✓ C03 population INTEGER (int4)    NOT NULL  →  Int

scala> (q + filter("U%")).query[Info].quick.unsafePerformIO
  Info(United Arab Emirates,ARE,2441000)
  Info(United Kingdom,GBR,59623400)
  Info(Uganda,UGA,21778000)
  Info(Ukraine,UKR,50456000)
  Info(Uruguay,URY,3337000)
  Info(Uzbekistan,UZB,24318000)
  Info(United States,USA,278357000)
  Info(United States Minor Outlying Islands,UMI,0)
```

guersam:
Regarding list, a rough idea pops up that adding something like `Composite[Option[Fragment.Aux[?]]]` with more combinators such as `&&` to handle this common use cases: https://gist.github.com/cvogt/9193220

fanf:
Oh god yes!
With that, I believe my only need to go back to HPS/HC disapear (in https://github.com/Normation/rudder/blob/branches/rudder/4.0/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/EventLogJdbcRepository.scala#L146)

:)

yax/core/src/main/scala/doobie/syntax/string.scala
```diff
@@ -91,20 +97,72 @@ instance for each element in the REPL. See the FAQ in the Book of Doobie for mor
...
+          ca.update(n, a) *> cb.update(n + ca.length, b)
+        }
+      }
+  }
+
+
+  sealed trait Fragment { outer =>
+    type A <: HList
+    val A: Composite[A]
+    val a: A
```
tpolecat: Yeah I confused myself when I was writing this. I'll make it more clear.

tpolecat:
@guersam I'll play around with adding some more combinators. I'm preserving the input type `A` because I need it to for the composite's *read* case which is never used, so I could probably just cheat and leave that case unimplemented and erase `A`, which would allow for generic treatment. I'll play around with it.

tpolecat:
Yeah I can totally make `A` existential and simplify. Commits incoming, but this lets you do stuff like this:

```scala
  // `WHERE c1 AND c1 AND ... cN` if any args are defined, otherwise Fragment.empty. This is an
  // easy way to crush a list of optional filters into a single fragment.
  def where(clauses: Option[Fragment]*): Fragment =
    clauses.flatten.toList.foldSeq1(fr"WHERE", fr"AND")

  // Construct a Query0 with two optional filter conditions.
  def select(name: Option[String], pop: Option[Int]): Query0[(String, String, String)] = {
    val sel = fr"SELECT name, code, population FROM country"
    val f1  = name.map(s => fr"name LIKE $s")
    val f2  = pop.map(n => fr"population > $n")
    val q = sel + where(f1, f2)
    q.query[(String, String, Int)]
  }
```

I don't think I will provide things like `where` in the lib because it will be never-ending and will evolve into a terrible DSL, but I can provide generic `Foldable/Monoid` ops like `foldSeq1` above.

tpolecat:
Ok I made the argument type existential in `Fragment` which works well. Moved a few other things around but that's mainly it. `sql` is just an alias for `fr` now.

It's unclear how much end-user support to add. Any combinators you want can be defined in terms of `Foldable` and `Monoid`.

tpolecat:
Fragments can also subsume the `IN` hack which is nice … `FragmentExample.scala` demonstrates this.

tpolecat:
I added the generalized `mkString` for `Foldable` because it's very useful for composing `WHERE` and `IN` clauses, but I need a better name. Right now it's `foldSmash`. Suggestions welcome.

tpolecat:
Ok I'm pretty happy with this if anyone wants to have another look. `FragmentExample.scala` has an example usage. I'll back-fill some tests but I'm pretty sure it's right.

guersam:
:+1: Removing `Concat` in favor of `zip` is brilliant and the combinators seem to cover the most of frequently requested features (`in` will be my new favorite, goodbye, `.narrow`!)

One last request is adding `Param[Fragment]` if possible to replace most `Param.many` cases with something like this:

```scala
// Before
val nel = list.toNel.get
implicit val pm = Param.many(nel)
sql"""
SELECT name, population FROM country
WHERE code IN (${nel.type})
ORDER BY population DESC
"""

// Current
sql"SELECT name, population FROM country WHERE" +
fragments.in(list) +
sql"ORDER BY population DESC"

// After
sql"""
SELECT name, population FROM country
WHERE ${fragments.in(fr"code", list)}
ORDER BY population DESC
"""
```

guersam:
quick brainstorming for `foldSmash` naming:  `foldWrap(ped)`, `foldSep`, `foldBorder`, `foldUnsplit`, `unsplit`, `wrapUnsplit`

tpolecat:
`Param[Fragment]` is an interesting idea, I'll look into that.

yax/core/src/main/scala/doobie/util/fragments.scala
```diff
@@ -0,0 +1,51 @@
...
+#+cats
+import cats.{ Reducible => Foldable1, _}, cats.implicits._
+#-cats
+
+/** Module of `Fragment` constructors. */
+object fragments {
+
+  /** Returns `f IN (fs0, fs1, ...)`. */
+  def in[F[_]: Foldable1, A: Param](f: Fragment, fs: F[A]): Fragment =
+    fs.toList.map(a => fr0"$a").foldSmash1(f + fr0"IN (", fr",", fr")")
```
leifwickland: What protects against an empty `fs`?

tpolecat: `Foldable1` means it can't be empty.

yax/core/src/main/scala/doobie/util/fragments.scala
```diff
@@ -0,0 +1,51 @@
...
+#-scalaz
+#+cats
+import cats.{ Reducible => Foldable1, _}, cats.implicits._
+#-cats
+
+/** Module of `Fragment` constructors. */
+object fragments {
+
+  /** Returns `f IN (fs0, fs1, ...)`. */
+  def in[F[_]: Foldable1, A: Param](f: Fragment, fs: F[A]): Fragment =
```
leifwickland: The `NOT IN` use case would look like:
```
in(fs"user_name NOT", NonEmptyList("Jim", "Bob"))
```
right?

tpolecat: Yeah but that's kind of janky. Should we add a `notIn` constructor? I fear this will never end but at least we can do the obvious ones.

yax/core/src/main/scala/doobie/util/fragments.scala
leifwickland: Ah. Sorry. I should've figured that out.

yax/core/src/main/scala/doobie/util/fragments.scala
leifwickland: _shrug_
I'd be willing to live with this to avoid infinite combinators.

yax/core/src/main/scala/doobie/util/fragments.scala
```diff
@@ -0,0 +1,51 @@
...
+    whereAnd(fs.flatten: _*)
+
+  /** Returns `WHERE f1 OR f2 OR ... fn` or the empty fragment if `fs` is empty. */
+  def whereOr(fs: Fragment*): Fragment =
+    if (fs.isEmpty) Fragment.empty else fr"WHERE" + or(fs: _*)
+
+  /** Returns `WHERE f1 OR f2 OR ... fn` for defined `f`, if any, otherwise the empty fragment. */
+  def whereOrOpt(fs: Option[Fragment]*): Fragment =
+    whereAnd(fs.flatten: _*)
+
```
leifwickland: Now that I've mentioned `NOT` do you think a `def not()` would be useful? You could head off the `andNot` by suggesting:
```
and(fs.map(not))
```

leifwickland:
@tpolecat 
In general I think this is awesome and would've allowed me to keep @coltfred a lot happier when I built up complex queries dynamically. Thanks so much!

tpolecat:
Ok I changed the concatenation operator to `++` but that's also an issue because `"foo" ++ fr"..."` results in an error about `GenIterableOnce` or something like that due to the `String` to `Seq[Char]` conversion. But I don't want to use something weird so I might just call it good. I considered `||` which is what SQL uses for string concatenation but it says "or" to me. Blah.

I also got rid of `Param.many` entirely, which some people may not like. But it was a hack to begin with and it's solved much better with `Fragments.in`. I can't figure out how to implement `Param[Fragment]` without exposing the type and bringing back the `Concat` machinery again, so I think I will not do this. I prefer a clean solution even if it requires a tiny bit more typing.

Also added a bare-bones book chapter on fragments. Still need to review the remainder of the doc to see if there are more things that are better done with `fr` instead of `sql`. But I think it's close.

tpolecat:
Aaaaaaaand GC limit exceeded building the doc, whee.

guersam:
I'd prefer `++` to `+` either. 

I'll be missing `Param[Fragment]` for several `IN`s in complex queries which should be rewritten in terms of `Fragments.in`, but it's fine as I get alternative solution which is far better than `Param.many`.

tpolecat:
I made a new `0.3.1-SNAPSHOT` if anyone wants to play with this. Thanks everyone for your feedback. I think this is a good change.

PR #398, opened 2016-12-06
Fix Meta registration

### LINKED ISSUE #394: more memleak woes
i'm trying to stream inserts into a postgres db, and quickly running out of memory.

what i am doing:
i have one task writing into a ```scalaz.stream.async.boundedQueue```, ```queue.dequeue``` is mapped to ```Update0``` values for an INSERT, chunked by 10000 and fed into postgres like this:
```
dequeue
 .map(makeInsert) // it => sql"""something""".update
.chunk(10000)
.evalMap(
   _.traverse(_.run).transact(transactor)
)
.run
.unsafePerformSync
```

in a heap dump i see millions of ```meta$Meta$$anon$1``` instances and the same amount of
```meta$Meta$$anonfun$1```, ```meta$BasicMeta$$anonfun$2```, 
```meta$Meta$$anon$1$$anonfun$xmap$1```, ```meta$Meta$$anon$1$$anonfun$xmap$2```, ```meta$Meta$$anon$1$$anonfun$xmap$3```, ```meta$LowerPriorityImplicits$$anonfun$unaryProductMetaNonNullable$1``` and ```meta$LowerPriorityImplicits$$anonfun$unaryProductMetaNonNullable$2```

i checked some of these instances, and the nearest GC root is a ```ClassLoader``` referencing the ```meta$Meta$``` ```MODULE``` referencing a ```HashSet``` "instances" (```HashTrieSet```) referencing said ```meta$Meta$$anon$1```. 

interestingly almost (!) the same code seems to have been working with doobie 0.3, with doobie 0.3.1-SNAPSHOT it does not. i went back a few commits and tried which that, no change.

tpolecat:
`Meta` instances are registered on construction as a side-effect. This is used by query checking to provide hints about known mappings if the asserted ones don't work.

In 0.3.0 there is a `scalaz.Order[Meta[_]]` and the instances live in an `ISet` but this doesn't work in Cats, so I temporarily switched it to `Set` to get things compiling but I never switched it back to a set that uses reasonable equality, which resulted in #394 … inductively derived instances were getting re-registered.

This fixes that by defining both `scalaz.Order` and `cats.Order`, and from there deriving `scala.Ordering` and using a `TreeSet` as storage.

fixes #394

### DIFF
--- a/yax/core/src/main/scala/doobie/util/meta.scala
+++ b/yax/core/src/main/scala/doobie/util/meta.scala
@@ -165,7 +165,9 @@ object meta {
       Order.orderBy(_.fold(
 #-scalaz
 #+cats
-      Order.by(_.fold(
+      // Type argument necessary to avoid spurious "illegal cyclic reference involving object Meta"
+      // only in Scala 2.11, and only with cats for whatever reason. Confidence high!
+      Order.by[Meta[_], (String, NonEmptyList[JdbcType], NonEmptyList[JdbcType], List[JdbcType]) \/ (String, NonEmptyList[JdbcType], NonEmptyList[JdbcType], NonEmptyList[String])](_.fold(
 #-cats
         b => -\/((b.scalaType, b.jdbcTarget, b.jdbcSource, b.jdbcSourceSecondary)),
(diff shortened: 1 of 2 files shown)


tpolecat:
Aaaand compiler bug. Sigh.

PR #400, opened 2016-12-06
HikariDatasource.shutdown is deprecated, use HikariDatasource.close

channingwalton:


tpolecat:
Oh cool, I keep forgetting to fix that. Thanks!

channingwalton:
Great, thanks for accepting it.

PR #424, opened 2017-01-03
consistent imports

### LINKED ISSUE #279: Consistent all-in-one import clauses among contrib subprojects
Currently we need `import doobie.contrib.postgresql.pgtypes._` for some postgres-specific typeclasses such as array support, which is inconsistent with `import doobie.imports._` and easy to forget.

Incomplete IDE support makes it worse by dropping the import clause during optimizing import or move refactoring due to a false implicit resolution (it's not responsibility of doobie but it sometimes makes my life harder.)

I'd suggest making all-in-one imports consistent as `import doobie.contrib.*.imports._` for newcomers and victims of today's IDEs.

tpolecat:
This sets things up so `import doobie.<module>.imports._` will tend to do what you want.

resolves #279

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -233,7 +233,7 @@ def postgresSettings(mod: String): Seq[Setting[_]] =
     ),
     initialCommands := """
       import doobie.imports._
-      import doobie.postgres.pgtypes._
+      import doobie.postgres.imports._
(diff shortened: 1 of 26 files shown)


tpolecat:
@guersam is this what you had in mind?

guersam:
Wow there are much more changes than I thought; exactly as what I needed :+1:

PR #425, opened 2017-01-06
Add disclaimer about 1-based JDBC indexing to MappingViolation instances

leifwickland:
I'm happy to do this differently if you like. I considered adding the disclaimer to the `MappingViolation` constructor.

It seemed likely that there'd be a need to add the disclaimer to a message in `yax/core/src/main/scala/doobie/util/analysis.scala`. However I couldn't find any messages which included the index, so I didn't make a change there. It might be worth double-checking my (non-)work in that file.


tpolecat:
👍 Thanks!

PR #432, opened 2017-01-11
Minor README nitpick

gatorcse:
Not sure if necessary, but I find the `.x` after the version numbers reassuring that the library isn’t hard coded to only support the `0.9.0` release, but in fact can support (compile) with any `0.9.x` release. Especially with `0.x` releases where some library authors play a little fast and lose with SemVer (not saying these libraries do). Feel free to close if you disagree.


tpolecat:
I'm not convinced this adds any more information since all the libs in question are well-behaved in this regard. But I'll leave this open and think about it. Thank you for reviewing!

gatorcse:
Stupid autocorrect SemVer => Server

tpolecat:
I think I'm going to say no for this one, but thanks! We can revisit if people become confused.

PR #434, opened 2017-01-12
Ignore .idea

gatorcse:
Just adding IntelliJ to the ignored IDEs list


KadekM:
You may instead wish to add it to your global ~/.gitignore file.

gatorcse:
Yeah I will that, but all of the other IDEs are in there, and .figured it should just be added for consistency

PR #442, opened 2017-01-21
new interpreter design

### LINKED ISSUE #399: New Interpreter Design
Consider a new interpreter design that uses a visitor-like pattern for efficiency and encodes the embedding constraints differently to make it possible for others to write alternative interpreters. Sketch below.

```scala
import scalaz._, Scalaz._, scalaz.effect._
import java.sql._

// embeddable free languages
object alg {

  /** A pair (J, Free[F, A]) with constructors that tie down J and F. */
  sealed trait Embedded[A]
  case class EmbeddedConnection[A](x: Connection, k: Free[ConnectionOp, A]) extends Embedded[A]
  case class EmbeddedPreparedStatement[A](x: PreparedStatement, k: Free[PreparedStatementOp, A]) extends Embedded[A]

  trait Embeddale[F[_], J] {
    def embed[A](j: J, fa: Free[F, A]): Embedded[A]
  }

  /////////////////////// ConnectionOp

  sealed trait ConnectionOp[A] {
    def visit[F[_]](v: ConnectionOp.Visitor[F]): F[A]
  }

  object ConnectionOp {

    implicit val ConnectionOpEmebeddable: Embeddale[ConnectionOp, Connection] =
      new Embeddale[ConnectionOp, Connection] {
        def embed[A](j: Connection, fa: Free[ConnectionOp, A]) = EmbeddedConnection(j, fa)
      }

    case class Embed[A](e: Embedded[A]) extends ConnectionOp[A] {
      def visit[F[_]](v: Visitor[F]) = v.embed(e)
    }

    case class PrepareStatement(sql: String) extends ConnectionOp[PreparedStatement] {
      def visit[F[_]](v: Visitor[F]) = v.prepareStatement(sql)
    }

    // Visitor

    trait Visitor[F[_]] extends (ConnectionOp ~> F) {
      final def apply[A](fa: ConnectionOp[A]) = fa.visit(this)
      def embed[A](e: Embedded[A]): F[A]
      def prepareStatement(sql: String): F[PreparedStatement]
    }

    trait VisitorStub[F[_]] extends Visitor[F] {
      override def embed[A](e: Embedded[A]) = sys.error("not implemented: embed")
      override def prepareStatement(sql: String) = sys.error("not implemented: embed")
    }

    // smart ctors

    def embed[F[_], J, A](j: J, fa: Free[F, A])(implicit ev: Embeddale[F, J]): Free[ConnectionOp, A] =
      Free.liftF(Embed(ev.embed(j, fa)))

    def prepareStatement(sql: String): Free[ConnectionOp, PreparedStatement] =
      Free.liftF(PrepareStatement(sql))

  }

  implicitly[Embeddale[ConnectionOp, Connection]]

  /////////////////////// PreparesStatementOp

  sealed trait PreparedStatementOp[A] {
    def visit[F[_]](v: PreparedStatementOp.Visitor[F]): F[A]
  }
  object PreparedStatementOp {

    case class Embed[A](e: Embedded[A]) extends PreparedStatementOp[A] {
      def visit[F[_]](v: Visitor[F]) = v.embed(e)
    }

    // Visitor

    trait Visitor[F[_]] extends (PreparedStatementOp ~> F) {
      final def apply[A](fa: PreparedStatementOp[A]) = fa.visit(this)
      def embed[A](e: Embedded[A]): F[A]
    }

    // smart ctors

    def embed[F[_], J, A](j: J, fa: Free[F, A])(implicit ev: Embeddale[F, J]): Free[PreparedStatementOp, A] =
      Free.liftF(Embed(ev.embed(j, fa)))

  }




  /////////// INTERPRETERS

  trait KleisliInterpreter[M[_]] {
    implicit val M: Monad[M]

    def ConnectionInterp:        ConnectionOp        ~> Kleisli[M, Connection,        ?]
    def PreparedStatementInterp: PreparedStatementOp ~> Kleisli[M, PreparedStatement, ?]

    trait Embedding[J] {
      def embed[A](e: Embedded[A]): Kleisli[M, J, A] =
        e match {
          case EmbeddedConnection(a, k)        => Kleisli(_ => k.foldMap(ConnectionInterp).run(a))
          case EmbeddedPreparedStatement(a, k) => Kleisli(_ => k.foldMap(PreparedStatementInterp).run(a))
        }
    }

    trait ConnectionInterp extends ConnectionOp.Visitor[Kleisli[M, Connection, ?]] with Embedding[Connection]
    trait PreparedStatementInterp extends PreparedStatementOp.Visitor[Kleisli[M, PreparedStatement, ?]] with Embedding[PreparedStatement]

  }



  trait MainKleisliInterpreter extends KleisliInterpreter[IO] {
    val M: Monad[IO] = implicitly

    trait KleisliConnectionInterp extends ConnectionInterp {
      def prepareStatement(sql: String) = Kleisli(c => IO(c.prepareStatement(sql)))
    }

    trait KleisliPreparedStatementInterp extends PreparedStatementInterp {
    }

    override val ConnectionInterp = new KleisliConnectionInterp { }
    override val PreparedStatementInterp = new KleisliPreparedStatementInterp { }

  }



  /** Same as the normal interpreter but it induced an exception for statements called "foo". */
  val TestKleisliInterpreter = new MainKleisliInterpreter {
    override val ConnectionInterp = new KleisliConnectionInterp {
      override def prepareStatement(sql: String) = {
        if (sql == "foo") Kleisli(_ => IO(sys.error("induced failure!")))
        else super.prepareStatement(sql)
      }
    }
  }

}

```

tpolecat:
Resolves #399 

This new design is much cleaner/simpler and allows for custom interpreters. The big change is that you now define a *family* of interpreters that know about each other and must support embedding of programs. This was baked into the algebras before and was a mess, which made it more or less impossible to write an alternative interpreter.

Each `Op` algebra defines a `Visitor[F[_]]` that you can use to define your `Op ~> F` without pattern-matching, which is kind of OO but much more efficient.

Look at `kleisliinterpreter.scala` to see the new default interpreter.

This looks like a big PR but there are relatively few code changes outside of the `free2` package, which is entirely new (and will be renamed back to `free` in a later PR to avoid massive nonsense diffs in this one). The end-user API looks the same and the rest of `core` has very few changes.

This also removes all uses of `Unapply` since a 2712 fix is available for all targeted Scala versions.

I think this is ready for review if anyone is interested.


tpolecat:
ok not quite

tpolecat:
Ok that last commit switches us to the Typelevel compiler and removes 2.10 support … mostly in anger, I will probably revert it.

tpolecat:
Actually making it work on 2.10 is nontrivial.

tpolecat:
Benchmarks on `series/0.4.x`

```
    * jdbc |  6.52    -- |  6.46    -- |  8.59    -- | 18.23    -- |102.53    -- |1468.14    --
   process |  9.23   142 | 10.89   169 | 19.86   231 | 48.63   267 |219.75   214 |2764.42   188
     ilist |  7.06   108 |  7.01   109 |  7.88    92 | 19.02   104 |105.49   103 |1432.11    98
      list |  9.15   140 |  7.10   110 |  8.23    96 | 19.05   105 |107.05   104 |1660.82   113
    vector |  7.37   113 |  8.73   135 | 13.21   154 | 17.16    94 |109.03   106 |1340.23    91
      jdbc |  6.47    99 |  7.71   119 |  7.83    91 | 15.13    83 | 94.67    92 |1374.94    94
```

Benchmarks on `freegen2` are basically identical, so no perf regression.

```
    * jdbc |  6.88    -- |  6.31    -- |  7.95    -- | 17.49    -- | 98.39    -- |1371.15    --
   process | 10.42   151 | 11.42   181 | 17.73   223 | 40.46   231 |222.49   226 |2596.97   189
     ilist |  6.05    88 |  6.26    99 |  7.92   100 | 18.39   105 |101.62   103 |1393.09   102
      list |  6.83    99 |  6.33   100 |  8.03   101 | 20.05   115 |104.85   107 |1659.45   121
    vector |  7.44   108 |  6.41   102 |  8.99   113 | 17.01    97 |102.35   104 |1325.50    97
      jdbc |  6.81    99 |  7.30   116 |  6.83    86 | 15.66    90 | 95.42    97 |1371.38   100
```

tpolecat:
Experimenting with custom interpreters.

```scala
scala> def max(pat: String) = sql"select max(population) from country where name like $pat".query[Int].unique.transact(xa)
max: (pat: String)doobie.imports.IOLite[Int]

scala> max("A%").unsafePerformIO
conn.setAutoCommit(false)
conn.prepareStatement(select max(population) from country where name like ?)
  ps.setString(1, A%)
  ps.«captured effect»
  ps.executeQuery()
  ps.«captured effect»
    rs.next()
    rs.«opaque operation»
    rs.next()
    rs.close()
  ps.«captured effect»
  ps.«captured effect»
  ps.close()
conn.commit()
conn.close()
res2: Int = 37032000
```

purefn:
This is very nice! One thing I've been wondering about is mixing the doobie algebra with other algebras. If I write an algebra for MongoDB, or an HTTP client, it would be nice to be able to combine the doobie algebra with those via a `Coproduct`. This seems to get us closer to being able to do that, but it's still unclear to me how to get there. Any thoughts?

tpolecat:
@purefn good question. I'll see if I can get an example working. Should work fine, heh-heh.

However the problem *in general* is that doobie programs are transactional and might be rolled back partially or completely (or you might want to retry in case of deadlock) but anything else you do in an interleaved program stays done. So I don't really know how to reconcile this.

purefn:
Sounds good. I was considering the transactional nature and I'm not sure how to reconcile it either. But I also want to be able to stream results from doobie into MongoDB (or vice-versa) without having to drop to something low-level like `Task` or `IO`.

tpolecat:
@purefn so, this works if I define the injecty stuff for each algebra, but in order to make it usable I also have to parameterize the higher-level structures (the `.hi` modules ,`Query/Update`, `Composite`, etc.) on the coproduct, which kind of infects everything. I'll play around a bit more but I think it will be too much for this PR unless there's some kind of generalization I'm missing.

tpolecat:
@purefn ok, yeah it works but you have to interpret into a Kleisli arrow to get the transaction stuff where you want it. See [here](https://gist.github.com/tpolecat/aa8dc37734cc3dd173bfc857b732ac53). wdyt?

purefn:
That looks pretty good to me! It's unfortunate that the manual injection is necessary in the `combined` though. Ideally we could abstract the need for that with some type-class

```
def combined[F[_]](implicit connection: ConnectionF[F], trivial: TrivialF[F]) =
  for {
    n <- connection.query[Int](sql"select count(*) from country").unique
    b <- trivial.trivial("blah")
  } yield (n, b)

val k  = combined[Foo].foldMap(KleisliInterpreter[IOLite].ConnectionInterpreter or trivialInterp)
```

Then the fact that we're using a free monad is abstracted away from our program entirely. I wonder if it would be possible to write the `ConnectionF` type-class as doobie is now. 

At any rate, this is definitely moving in the right direction! Thanks!

tpolecat:
@purefn right, in order to do that all the machinery has to be parameterized over the coproduct which is a big API change.

tpolecat:
You can still abstract over the coproduct btw. Seems usable.

```scala
def combined[F[_]](implicit ev: Inject[ConnectionOp, F], t: TrivialF[F]): Free[F, (Int, String)] =
  for {
    n <- sql"select count(*) from country".query[Int].unique.inj[F]
    b <- t.trivial("blah")
  } yield (n, b)
```

PR #446, opened 2017-01-23
Fix doc

n4to4:


tpolecat:
Brilliant, thank you!

n4to4:
Oops, rebased (again)

tpolecat:
I see you pronounce "SQL" as "s-q-l" and not "sequel" like I do ;-)

n4to4:
Oh, I've never thought of that. I will revert that changes if you prefer. (there is one occurrence which says "an SQL" in 08-Fragments.md, though)

And one thing I forgot to metion. There's one more occurrence here:

```
yax/docs/src/main/tut/07-Updating.md
162:[...], which is the type of a `sql"..."` expression, represents[...]
```

I wasn't sure if it is pronounced as in normal English or it is just a symbol or something.

tpolecat:
I think "S-Q-L" is probably more common. I'm ok with your changes.

n4to4:
Should I change `a sql"..."` to `an sql"..."` or just leave it?

tpolecat:
Let's go with "an sql" everywhere.

n4to4:
Done

tpolecat:
Thanks!

PR #451, opened 2017-01-28
wip on new transactor design

### LINKED ISSUE #443: Query.executeQuery doesn't explicitly close resultset
Query timing logic introduced this problem, which isn't a huge deal but it will cause the cursor to be held open until the connection is closed. Would like to merge #442 first, which will allow us to write a test for this.

```scala
  et <- c.attempt(FPS.lift(rs, k))
```
should be

```scala
  et <- c.attempt(FPS.lift(rs, k.ensuring(FRS.close)))
```

tpolecat:
This turns `Transactor` into a non-obvious but simple data that supports all current implementations and allows the user to configure the transaction-handling strategy.

- [x] Coproduct example (!)
- [x] Need to add some resource-safety tests, which will be easy now that we can write custom interpreters!

fixes #443


tpolecat:
BTW this started out as an example of how to use doobie in a coproduct but once I started pulling the string the whole sweater unraveled.

tpolecat:
I'm irritated that the `A` in `Transactor[M, A]` can normally be treated existentially. It should be a type member. Think about it a bit more.

PR #453, opened 2017-01-30
Replaces a link with potentially chain-opening behaviour

AlexaDeWit:
In the old link it was possible for some browser configurations to spawn a new open request each time one opened, infinitely looping until browser crash.


tpolecat:
Thanks!

Issue #280, opened 2016-03-29
Refined support for doobie


tpolecat:
see initial work in #265

PR #456, opened 2017-02-03
Refined support for doobie #280

btre:
Meta and Composite instances for refinement-types (including tests/ I think tests will also work as examples)


tpolecat:
Thanks! CI issues in pre-2.12 Scala … looks minor.

yax/refined/src/main/scala/doobie/refined/package.scala
```diff
@@ -0,0 +1,42 @@
...
+        t => rightOrException[T Refined P](refineV[P](t)(v)),
+        _.value
+      )
+#-scalaz
+#+cats
+      compositeT.imap[T Refined P](t => rightOrException[T Refined P](refineV[P](t)(v))(_.value)
+#-cats
+
+  private[this] def rightOrException[T](either: Either[String, T]): T = either match {
+    case Left(err) => throw new IllegalArgumentException(err)
```
tpolecat: Can you create a new subtype of `InvariantViolaton` [here](https://github.com/tpolecat/doobie/blob/series/0.4.x/yax/core/src/main/scala/doobie/util/invariant.scala#L27-L29) and throw that instead? It should have a generic name since it's in core … maybe `SecondaryValidationFailed` or something.

yax/refined/src/main/scala/doobie/refined/package.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  import eu.timepit.refined.api.{Refined, Validate}
+  import eu.timepit.refined.auto._
+  import eu.timepit.refined._
+  import scala.reflect.runtime.universe.TypeTag
+
+  import doobie.imports._
+
+  implicit def refinedMeta[T, P](implicit metaT: Meta[T], v: Validate[T, P], tag: TypeTag[T Refined P]): Meta[T Refined P] =
+    metaT.xmap[T Refined P](
```
tpolecat: I think you need `nxmap` here because the values here can be null. There should be a test for read/write of `Option[A]` for some refined type `A` to make sure nullity is respected.

yax/refined/src/test/scala/doobie/refined/refinedtypes.scala
```diff
@@ -0,0 +1,86 @@
...
+      Composite[PointInQuadrant1]
+
+      true
+    }
+  }
+
+  "Query" should {
+    "return a refined type when conversion is possible" in {
+      val pInt: PositiveInt =
+        Query[Unit,PositiveInt]("select 123", None).unique(()).transact(xa).unsafePerformIO
```
tpolecat: Any reason not to say `sql"select 123".query[PositiveInt].unique` here and elsewhere?

yax/refined/src/main/scala/doobie/refined/package.scala
tpolecat: It also looks like `private[this]` may not work pre-2.12 … there's a CI failure there.

tpolecat:
Also need to rebase, sorry. I slipped in a change underneath you.

btre:
thanks for your feedback. I'll work on the issues you've mentioned as soon as I can find time to do that.

yax/refined/src/main/scala/doobie/refined/package.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  import eu.timepit.refined.api.{Refined, Validate}
+  import eu.timepit.refined.auto._
+  import eu.timepit.refined._
+  import scala.reflect.runtime.universe.TypeTag
+
+  import doobie.imports._
+  import doobie.util.invariant._
+
+  implicit def refinedMeta[T, P](implicit metaT: Meta[T], v: Validate[T, P], tag: TypeTag[T Refined P]): Meta[T Refined P] =
```
fthomas: Besides `Refined`, refined (the library) also supports `shapeless.tag.@@` and `scalaz.@@` as types for attaching a predicate to a base type. You could abstract over the tag type by adding another type parameter `F[_, _]` and an implicit parameter `rt: RefType[F]` which allows you to change the return type to `Meta[F[T, P]]`. In the implementation you can then use `rt.refine` and `rt.unwrap` for refining / unwrapping values. See for example [circe-refined](https://github.com/circe/circe/blob/master/modules/refined/shared/src/main/scala/io/circe/refined/package.scala) how you could use `RefType` here.

Btw, the same also applies to the `Composite` instance.

tpolecat: Aha, that's a good suggestion. Thanks @fthomas

btre: Thanks @fthomas! 
My current problem is making it null-safe. My (maybe naive) attempt was 
`  implicit def refinedMeta[T >: Null, P, F[T,P] >: Null](implicit metaT: Meta[T], v: Validate[T, P], refType: RefType[F], ev: Null <:< T, tag: Manifest[F[T,P]]): Meta[F[T,P]] =
    metaT.xmap[F[T,P]](
      t => if (t == null) null else rightOrException[F[T,P]](refType.refine[P](t)(v)),
      ftp => if (ftp == null) ev(null) else refType.unwrap(ftp)
    )
`
but with this the compiler is not able to find a typeclass for types like `type PositiveInt = Int Refined Positive`

tpolecat: If you use `.nxmap` you don't have to worry about nulls and it might clean up your implementation. Let me know if you can't get it to work.

beefyhalo:
I've encountered the Refined issue with the >: Null constraint in doobie. 
FWIW, I use an Atom[A] and an Atom[Option[A]] instead of a Meta.

btre:
thanks @beefyhalo, providing Atom-instances solved my problem

tpolecat:
Is this ready to go? Looks good to me. Needs another rebase, sorry.

fthomas:
Looks good to me too. :+1:

tpolecat:
@btre can you rebase and also add your name/handle and a bullet point to the 0.4.2 in-progress section of the changelog?

btre:
@tpolecat I merged and updated the changelog

tpolecat:
Thanks!

PR #458, opened 2017-02-04
Fix execution flow for AnalysisSpec checks

### LINKED ISSUE #454: AnalysisSpec doesn't follow specs2 execution flow
`check` in `AnalysisSpec` seems to run its queries before it returns, only using the resulting `Fragments` for outputting results. This circumvents specs2 execution flow, meaning that some of its features don't work as expected. For example:

```scala
import doobie.imports._
import doobie.specs2.imports._
import doobie.h2.imports._
import org.specs2.mutable.Specification
import org.specs2.specification.BeforeAll

class Spec extends Specification with AnalysisSpec with BeforeAll {

  // Setup
  val initQ = sql"create table some_table (value varchar not null)".update
  val targetQ = sql"select * from some_table".query[String]

  val transactor: Transactor[IOLite] = H2Transactor[IOLite](
    "jdbc:h2:mem:",
    "sa",
    ""
  ).unsafePerformIO

  // The test itself
  check(targetQ)

  // A hook for database initialization
  def beforeAll() = initQ
    .run
    .transact(transactor)
    .unsafePerformIO
}
```

This fails, since `beforeAll` is executed _after_ the test query:
```
Table "SOME_TABLE" not found
```
An obvious workaround in this particular case is running `beforeAll` explicitly without relying on hooks.

A proper solution would probably have to deal with dynamically generating fragments. I'm not proficient enough in specs2 to find a way to add new fragments from an existing one, though.

Tested with doobie v0.4.0, specs2 v3.8.4

nigredo-tori:
fixes #454

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -343,6 +343,7 @@ lazy val specs2 = project.in(file("modules/specs2"))
     scalazCrossSettings
   )
   .dependsOn(core)
+  .dependsOn(h2 % "test")
 
 lazy val specs2_cats = project.in(file("modules-cats/specs2"))
   .settings(
--- a/build.sbt
+++ b/build.sbt
@@ -351,6 +351,7 @@ lazy val specs2_cats = project.in(file("modules-cats/specs2"))
     specs2Settings("specs2-cats")
   )
   .dependsOn(core_cats)
+  .dependsOn(h2_cats % "test")
 
 ///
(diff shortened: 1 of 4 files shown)


tpolecat:
Thanks, I will take a look at this later today.

tpolecat:
rebased as #501

PR #476, opened 2017-03-03
Problem: update statements not being logged

vladimir-lu:
This PR corrects a trivial mistake where the custom `executeUpdate` function was never being called, which resulted in no logging output even if a handler was defined.

I dislike not having a test for the logging functionality for executeUpdate - but I wasn't sure how to add one so would appreciate some pointers.

### DIFF
--- a/yax/core/src/main/scala/doobie/util/update.scala
+++ b/yax/core/src/main/scala/doobie/util/update.scala
@@ -115,7 +115,7 @@ object update {
      * @group Execution
      */
     def run(a: A): ConnectionIO[Int] =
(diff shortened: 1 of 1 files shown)


tpolecat:
Ugh, thanks. Yep definitely need a test. I'll follow up over the weekend.

tpolecat:
Branched to #496.

Arcovion:
On 0.4.2 inserts using `withUniqueGeneratedKeys` aren't being logged:
```scala
sql"INSERT INTO...".update.run
```
works, but this doesn't:
```scala
sql"INSERT INTO...".update.withUniqueGeneratedKeys[Long]("id")
```
I'm using an implicit `LogHandler`

PR #488, opened 2017-03-20
Fixes image links in readme

dozed:
Github did not render the image links


tpolecat:
Huh, I guess GH changed something. Thanks.

(CI failure is sporadic because where's the fun in a having a deterministic compiler.)

PR #492, opened 2017-03-21
Do not use refined's macros in tests

fthomas:
Using refined's macro in doobies tests leads to nondeterministic
compiler crashes (https://github.com/fthomas/refined/issues/260).
Fortunately these macros are not essential for testing doobie-refined,
so I removed its usage from the tests.


tpolecat:
Aha, cool. Thank you!

PR #498, opened 2017-04-11
Composite instance for Unit

wedens:
`Composite[Unit]` has the same meaning as `Composite[HNil]`.
It allows fields that are ignored by database mapping. Such fields are useful for parametric types.
For example:
```scala
case class Topic[A](name: String, posts: A)
type TopicWithoutPosts = Topic[Unit]
// note there is only one column in select statement
sql"select name from topic".query[TopicWithoutPosts]
```


tpolecat:
Cool, I think this makes sense.

PR #509, opened 2017-04-22
Remove 'contrib-' in artifact names

wedens:
There were some leftovers.


tpolecat:
Thanks, can't believe I missed this.

PR #514, opened 2017-05-09
Add scalaz and cats flags for the unit testing documentation

suhasgaddam:
If there is a better way to do this, feel free to close the PR.


tpolecat:
Thank you! Yes, the doc needs some review for this kind of thing.

PR #515, opened 2017-05-10
Postgres enum can actually have a valid Meta instance

wedens:
well... it just works as is


tpolecat:
Well, cool. This may work now because of #484. Thanks!

PR #518, opened 2017-05-11
tpolecat has not been seen on IRC in quite a while iirc

SethTisue:


tpolecat:
True, yes. Thank you.

PR #519, opened 2017-05-13
generic effect type for scalacheck

### LINKED ISSUE #448: Don't force IOLite usage in scalatest bindings
Because `QueryChecker` wants a `override def transactor: Transactor[IOLite]`, people not using IOLite in their main code base are forced to change their code to use things like `def dbTransactor[M[_]: Monad: Catchable: Capture]: Transactor[M] = ???` in their configuration object, just so that it's usable in both real code and test, whereas I'd rather `def dbTransactor: Transactor[Task]`. This is especially impactful for users not as familiar with scala.

It should suffice to change `trait QueryChecker { def transactor: Transactor[IOLite] }` to be `trait QueryChecker[M[_]: Monad: Catchable: Capture] { def transactor[M]: Transactor[M] }`

tpolecat:
This lets the scalacheck trait work with an arbitrary effect type.

resolves #448

### DIFF
--- a/yax/scalatest/src/main/scala/doobie/scalatest/imports.scala
+++ b/yax/scalatest/src/main/scala/doobie/scalatest/imports.scala
@@ -5,7 +5,7 @@ object imports {
   type Checker[M[_]] = doobie.scalatest.Checker[M]
   type IOLiteChecker = doobie.scalatest.IOLiteChecker
 
-  @deprecated("Use CheckerIOLite.", "0.4.2")
+  @deprecated("Use IOLiteChecker.", "0.4.2")
   type QueryChecker = IOLiteChecker
 
 }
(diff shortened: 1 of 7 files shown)


tpolecat:
cc @Daenyth

yax/scalatest/src/main/scala/doobie/scalatest/imports.scala
```diff
@@ -1,5 +1,11 @@
 package doobie.scalatest
 
 object imports {
-  type QueryChecker = doobie.scalatest.QueryChecker
+
+  type Checker[M[_]] = doobie.scalatest.Checker[M]
+  type IOLiteChecker = doobie.scalatest.IOLiteChecker
+
+  @deprecated("Use CheckerIOLite.", "0.4.2")
```
Daenyth: CheckerIOLite here but below is IOLiteChecker?

Is that an error or am I misunderstanding how this should be used?

tpolecat: Yeah, typo. Thanks.

tpolecat:
@Daenyth there is a usage in the examples project but yeah that's a good idea, just as a sanity check. I will add a test.

PR #522, opened 2017-05-19
bump scala and library versions

dvic:
Not sure what the policy is regarding dependency versions, but let's see if tests pass with the latest versions.


tpolecat:
Thanks, fingers crossed.

dvic:
trying with JDBC reverted to 42.0.0, seems like a JDBC type with id '2014' is causing problems. JDBC/database version mismatch?

tpolecat:
Yeah it looks like they added a new one. I need to change the way that's handled. JDBC specifies it as an enumeration but people keep adding new values outside the spec.

dvic:
Looks like 2014 is a missing JDBC type in the library (`TimestampWithTimezone`, introduced in JDK 8)

tpolecat:
Ah cool. Thanks.

dvic:
Fixes #523

tpolecat:
Looks good. Can you add your name and a summary of these changes to `CHANGELOG.md`?

dvic:
Done.

tpolecat:
Thanks!

PR #532, opened 2017-06-15
WIP: normalize read/write treatment in meta/composite, remote atom

tpolecat:
This generalizes #484 in an attempt to increase performance for bulk inserts, but it turns out it doesn't really have much effect. However it does simplify meta and composite by extracting their common core, and gets rid of atom altogether.

Still WIP, may change. Mostly opening so I can look at the diff.

- [x] `Cartesian` instances for `Kernel` and `Composite`.
- [x] Update doc
- [x] update implicitNotFound on Composite


tpolecat:
```
scala> Composite[(Object,String)]
<console>:15: error: Could not find or construct Composite[(Object, String)].

  If (Object, String) is a simple type (or option thereof) that maps to a single column, you're
  probably missing a Meta instance. If (Object, String) is a product type (typically a case class,
  tuple, or HList) then probably one of its component types is missing a Composite instance. You can
  usually diagnose this by evaluating Composite[Foo] for each element Foo of the product type in
  question. See the FAQ in the Book of Doobie for more hints.

       Composite[(Object,String)]
                ^
```

tpolecat:
I'm inclined to merge this. I think it's an improvement and I'd like to get refocused on getting a release out. Any last words?

PR #539, opened 2017-06-27
restore osgi headers

tpolecat:
I removed these after 0.3 but we need them again for Gemini stuff.


tpolecat:
Well, I now have a runtime dependency on https://github.com/lihaoyi/sourcecode and need to add headers there.

tpolecat:
On hold pending https://github.com/lihaoyi/sourcecode/pull/32

PR #540, opened 2017-06-29
WIP on option composites

### LINKED ISSUE #200: Derive Composite[Option[A]] given Composite[A]?

tpolecat:
resolves #200, subsumes #372 

needs doc

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -52,7 +52,7 @@ lazy val commonSettings = Seq(
       "-language:implicitConversions",
       "-language:experimental.macros",
       "-unchecked",
-      "-Xlint",
+      // "-Xlint",
       "-Yno-adapted-args",
       "-Ywarn-dead-code",
       "-Ywarn-value-discard"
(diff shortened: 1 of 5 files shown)


tpolecat:
Closed via #576

PR #544, opened 2017-07-12
add streaming copy example

### LINKED ISSUE #543: Example for streaming rows from one database to another.
As per [gitter](https://gitter.im/tpolecat/doobie?at=5960146c4bcd78af568a0255) conversation, I'm looking to stream rows from one database to another in an efficient fashion.

Ideally the book of doobie would have an example showing how to do this.

tpolecat:
This adds an example with a combinator that demonstrates streaming from one database to another. Resolves #543


tpolecat:
I think this is good enough. PRs welcome if anyone has improvements.

PR #548, opened 2017-07-16
Replace `Manifest` with `TypeTag`

wedens:


tpolecat:
This is great, thank you!

PR #549, opened 2017-07-16
Hide extra type parameter in `Transactor`

wedens:
Type parameter `A` is removed from `Transactor` type signature. `Transactor` constructor now return `Aux` type so that you can retain `A` or throw it away if you don't need it.


tpolecat:
Thanks, I think this is an improvement.

Issue #552, opened 2017-07-20
Failing to map H2 UUID to scala UUID


Jacoby6000:
Continuing the discussion from gitter:

```scala
import $ivy.`org.tpolecat::doobie-h2-cats:0.4.1`
// import $ivy.$
import $ivy.`org.tpolecat::doobie-core-cats:0.4.1`
// import $ivy.$
import doobie.imports._
// import doobie.imports._
import doobie.h2.imports._
// import doobie.h2.imports._
val transactor = DriverManagerTransactor[IOLite]( 
  "org.h2.Driver", "jdbc:h2:./test:test", "", "" 
  // using a file db + DriverManagerTransactor rather than an H2Transactor because yolo. 
  // results seem to be the same either way.
)
// transactor: Transactor[IOLite] = doobie.util.transactor$DriverManagerTransactor$$anon$2@2cf29dce
import transactor.yolo._
// import transactor.yolo._
fr"""
      CREATE TABLE IF NOT EXISTS fail(
        id UUID PRIMARY KEY,
        time TIMESTAMP NOT NULL,
        bool BOOLEAN NOT NULL DEFAULT FALSE
      )
""".update.quick.unsafePerformIO
//   0 row(s) updated

sql"select * from fail".query[(java.util.UUID, java.time.Instant, Boolean)].check.unsafePerformIO
/*
  select * from fail

  ✓ SQL Compiles and Typechecks
  ✕ C01 ID   BINARY    (UUID)      NOT NULL  →  UUID
    - BINARY (UUID) is not coercible to UUID according to the JDBC specification or
      any defined mapping. Fix this by changing the schema type to OTHER or
      JAVAOBJECT, or the Scala type to Array[Byte].
  ✓ C02 TIME TIMESTAMP (TIMESTAMP) NOT NULL  →  Instant
  ✓ C03 BOOL BOOLEAN   (BOOLEAN)   NOT NULL  →  Boolean
*/
```

Jacoby6000:
To add on to this, a `MERGE INTO` query exhibits strange behavior

```scala
sql"""
    MERGE INTO fail(id, time, bool)
      KEY (id)
      VALUES ($uuid, $t, $b)
  """.update.check.unsafePerformIO
/*
    MERGE INTO fail(id, time, bool)
      KEY (id)
      VALUES (?, ?, ?)

  ✓ SQL Compiles and Typechecks
  ✕ P01 UUID     →  VARCHAR (VARCHAR)
[...]

PR #555, opened 2017-07-24
Fix analysis when using h2 + UUIDs

Jacoby6000:
Fixes a portion of #552. I haven't been able to figure out what's up with merge statements reporting VARCHAR for every type.. I was able to kind of fix it by using the composite to set the params before running the analysis, but things still acted funky in situations involving NULL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This file summarizes **notable** changes for each release, but does not describe
 
 ### <a name="0.4.2"></a>Notable Work in Progress for Version 0.4.2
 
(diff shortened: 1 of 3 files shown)


tpolecat:
Looks good. 👍 Can you add a bullet point to `CHANGELOG.md`?

Jacoby6000:
👍 done.

tpolecat:
Thanks!

Jacoby6000:
@tpolecat the travis errors that are happening are probably because the default linux dist in Travis changed from `precise` to `trusty` recently.  Your most recent successful build was on `precise` and all failing builds are on `trusty`.

tpolecat:
Hell.

PR #556, opened 2017-07-24
Fix Travis builds for the new ubuntu trusty default environment

Jacoby6000:
Looks like the `precise` environment had the `postgresql-9.3-postgis-2.3` apt package installed by default, and `trusty` does not.  This was causing a failure whenever the before-script tried to create the postgis extension.  I added the package to `addons.apt.packages` in `.travis.yml`.  This [seems to have fixed the problem](https://travis-ci.org/Jacoby6000/doobie/jobs/257047857). Another solution would be to set the `dist` to `precise` in the build file, but if the default env is going to be `trusty` I figure it's best just to roll with it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,9 @@ env:
 
 addons:
   postgresql: "9.3"
+  apt:
+    packages:
+    - postgresql-9.3-postgis-2.3
 
 before_script:
   - project/travis-before-build.sh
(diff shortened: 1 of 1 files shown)


tpolecat:
Brilliant, thanks! We'll see what Travis says.

Jacoby6000:
Well, that is certainly an interesting failure mode.

tpolecat:
fascinating

Jacoby6000:
Ah, it's failing while building the docs.  The failures also seem to be valid failures.  I wonder why these didn't show up before.

tpolecat:
Closed via #557

PR #569, opened 2017-08-08
fix HikariTransactor.apply

aoiroaoino:
This bug exists only in `series/0.4.x`.
https://github.com/tpolecat/doobie/blob/series/0.5.x/modules/hikari/src/main/scala/doobie/hikari/HikariTransactor.scala#L24-L25


tpolecat:
Yikes, thanks for catching this. I fixed in in 0.5.x because it causes a compilation error with the new warnings turned on, then forgot to back-port it.

PR #578, opened 2017-08-24
fix aliases and imports

tpolecat:
Ok, so this is kind of a monster PR but it doesn't do that much.
- The importing of `FC`, `HC`, etc., are now managed by mixins so they're consistent everywhere.
- The `foo.imports` objects are deprecated in favor of distinct imports `foo._` for normal symbols and `foo.implicits._` for implicit symbols.


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Notable changes thus far:
 - The doc has been ported to [sbt-microsites](https://github.com/47deg/sbt-microsites) and is undergoing review (help wanted!).
 - The release process is somewhat better.
 - Syntax classes are now organized as in cats. Much cleaner but end users probably won't notice.
+- Rather than `foo.imports._` for both names and implicits, there are now distinct imports `foo._, foo.imports._`. The old `foo.imports._` still works but is deprecated.
```
xavier-fernandez: TYPO:

...distinct imports `foo._, foo.*imports*._`. The old...

tpolecat: Doh! Thank you.

PR #587, opened 2017-09-01
fix invalid to check TEXT column (on such as MySQL)

x1-:
If there is a TEXT column (on such as MySQL), Analysis is going to fail.
Because thease have `JdbcType = LongVarChar`  and `vendorTypeName = Varchar`,
but `jdbcSourceSecondary` of `StringMeta` has `LongVarChar`.


tpolecat:
Hm why won't ci run?

tpolecat:
There we go.

tpolecat:
Ugh finally. Thanks for these, sorry about the delay!

PR #588, opened 2017-09-01
fix invalid to check TEXT column (on such as MySQL)

x1-:
If there is a TEXT column (on such as MySQL), Analysis is going to fail.

please see: https://github.com/tpolecat/doobie/pull/587


tpolecat:
Closed via #622

tpolecat:
Thanks, sorry for the delay.

PR #589, opened 2017-09-03
wartremover

tpolecat:
This adds [WartRemover](http://www.wartremover.org/) and fixes a bunch of warts, as well as adding a lot of `@SuppressWarnings` annotations for things that aren't immediately fixable.


modules/core/src/main/scala/doobie/free/Aliases.scala
```diff
@@ -39,18 +41,46 @@ trait Modules {
...
+    callablestatement.AsyncCallableStatementIO
+
+  implicit lazy val AsyncClobIO: Async[ClobIO] =
+    clob.AsyncClobIO
+
+  implicit lazy val AsyncConnectionIO: Async[ConnectionIO] =
+    connection.AsyncConnectionIO
+
+  implicit lazy val AsyncDatabaseMetaDataIO: Async[DatabaseMetaDataIO] =
+    databasemetadata.AsyncDatabaseMetaDataIO
```
xavier-fernandez: Missing an empty line bellow this...

Issue #593, opened 2017-09-05
restore the yaxed-away examples in 0.5.x


tpolecat:
Some of the examples were scalaz-only and resulted in empty source files on the 0.5.x branch. Need to pull them over and port to cats.

PR #599, opened 2017-09-15
yaxed examples ported to 0.5.x

TharinduDG:
[restore the yaxed-away examples in 0.5.x](https://github.com/tpolecat/doobie/issues/593)


modules/example/src/main/scala/example/FirstExample.scala
```diff
@@ -3,3 +3,140 @@
 // For more information see LICENSE or https://opensource.org/licenses/MIT
 
 // relies on streaming, so no cats for now
+package example
+
+import cats.implicits._
+import cats.effect.IO
+import fs2.Stream
+import doobie._, doobie.implicits._
```
TharinduDG: using `fs2.Stream`

tpolecat:
Hey this is great! Thank you. I'll have a close look this weekend.

PR #606, opened 2017-09-29
Specs2, ScalaTest: refactor, add matcher syntax

### LINKED ISSUE #605: Specs2: make typechecking a matcher
We use query typechecking feature in Doobie extensively in our project. Ideally, every query should be `check`ed. However, we have a few queries that are built depending on input data, and we want to be able to use ScalaCheck to check them - but this results in a huge amount of output.

The problem is that the checker is implemented in terms of `Fragments`, using mutable `Specification` and its DSL, so there is no way to stop it from printing results.

Ideally I would like to see some kind of matcher syntax for this in addition to `check`, for more flexibility. E.g. something like this:
```scala
"it works" >> {
  myQuery must typecheck
}
```
where the message is only printed if the check fails.

nigredo-tori:
Closes #605

1. Factor out parts of the `Checker` - leave only things that directly use `Specification`, such as control flow and formatting. Rewrite formatting logic using `Block`s where possible.
2. Make the unsafe `ConnectionIO` execution an implicit, untangled from `Checker` (should allow using it in immutable specifications, too).
3. Make a typeclass for unpacking common parts of `Query/Update` variants (to avoid having duplication like `check` overloads).
4. Implement `AnalysisMatchers` trait with specs2 matchers, providing syntax like this:
```scala
  myQuery must typecheck
```

The results:
1. Basic analysis logic (and most of formatting) is untangled from `specs2` control flow, allowing reuse in other testing packages in the future (though this would require moving it to some other module).
2. Matcher syntax allows using the checks in scalacheck props (as described in #605), as well as in immutable specifications.

Compatibilty:
1. `Checker` interface did not change except for a new ancestor and an added implicit.
2. `Checker` output format is mostly unchanged (except for minor wrapping/whitespace changes).
3. `AnalysisMatchers` output mirrors `Checker` output.


tpolecat:
Thanks for this! Looks great. I'll try to get caught up with PRs this week.

tpolecat:
So, I guess my questions here are (a) can we generalize any of this to ScalaTest, and (b) should we strengthen `Aync` to `Effect` here, which would give us the ability to run without requiring the `UnsafeXXX` machinery? I think it might be easiest to merge first and then sort it out.

nigredo-tori:
I finally got around to updating this MR:
1. Removed `unsafeRunSync` function and switched to `Effect` (transformation to IO is simple enough).
2. Deleted `UnsafeTransactionSupport` implicit, instead injecting transactor and `Effect` instance through trait implementation. This prevents us from enabling matchers via an import, but implementing the trait should be easy enough.
3. `UnsafeTransactions` base trait was renamed to `CheckerBase` (it doesn't include anything unsafe anymore).
4. Moved framework-independent parts to the core (package `doobie.util.testing`). Creating a separate module for one file without extra dependencies is overkill.
5. Rewrote scalatest checker. Difference in behavior with specs2 is preserved (the analysis is not printed if the check succeeds). Changes to shape, color and content of the output are relatively minor.
6. Added scalatest matcher.

Breaking changes:
1. Provided instance now has to be an `Effect`;
2. `unsafeRunSync` method removed from both checkers;
3. Function signatures in scalatest checker were changed a little (`WeakTypeTag` was replaced by `TypeTag`).

aeons:
This looks great. Are there any blockers on this?

tpolecat:
Thanks a lot for this, sorry it took so long!

PR #615, opened 2017-10-21
Point to branch 0.5.x in the curl call

mads-hartmann:
Once 0.5.x is merged into master this will have to be changed again ☺️


tpolecat:
Thanks! 

(We're not using `master` anymore.)

PR #621, opened 2017-11-09
Experimental Playground: Off Implicits Branch Changes

ChristopherDavenport:
For When/If Cats Implicits for FS2 is merged this is an available starting point.

This is more a playground so feel free to close rather than hang onto it to next release if you don't want it.


tpolecat:
Awesome, thank you!

tpolecat:
I think we're up to date here. Thanks!

PR #625, opened 2017-11-10
Fast inserts via COPY ... FROM STDIN

tpolecat:
This ports the WIP from the 0.4.x branch. Benchmark shows it's 10x faster than raw JDBC batch inserts.

```
[info] Benchmark            Mode  Cnt       Score        Error  Units
[info] d.b.c.fast_copyin   thrpt   10  579890.430 ± 102112.866  ops/s
[info] d.b.c.jdbc_copyin   thrpt   10   57975.573 ±   9040.026  ops/s
[info] d.b.c.naive_copyin  thrpt   10   44645.075 ±   6318.579  ops/s
```

Need to run longer to even out IO jitter but it's clearly much better.


TimothyKlim:
Can you include this awesome feature in the new milestone, please?

tpolecat:
Mm this kinds needs some tests and more instances but I'll try to get it in soon. Thanks for the encouragement!

tpolecat:
I'll target this for 0.5.1

tpolecat:
Switched from csv to text format because the encoding is simpler. Dealing with arrays is a mess.

tpolecat:
Anyway. 
- [x] tests
- [ ] doc

PR #627, opened 2017-11-14
Add Hikari Transactor Resource Bracketing

ChristopherDavenport:
I implemented manually rather than importing ops, but I could understand if you would improve that as well.

Is there an easy way to test that this behavior is properly handling resources in tests?


tpolecat:
Integrated into #632 thanks!

PR #630, opened 2017-11-15
Update 14-Managing-Connections.md

jfraudeau:
The configure method does not take a `A => IO[B]` but a `A => B`, hence the code inside IO will be discarded. This fixes the doc but having the method take a `A => IO[B]` would be nicer !


tpolecat:
Thanks, and yes, good point about making `configure` take an `F`. I'll leave this open for now and add that change to your PR.

tpolecat:
Hm at some point `configure` was changed as you suggest but this was not closed.

PR #638, opened 2017-11-28
Allow mapping between N(VAR)CHAR, NTEXT and string

asvanberg:
As per https://edelivery.oracle.com/otn-pub/jcp/jdbc-4_1-mrel-spec/jdbc4.1-fr-spec.pdf


tpolecat:
Thanks!

PR #644, opened 2017-12-15
Remove link to fixed IJ issue

keirlawson:


tpolecat:
Are we confident it's actually fixed?

keirlawson:
The example given in the description of the ticket certainly works for me on the latest IntelliJ and seems like it is working for others on the ticket. Happy to try any other snippets if there are other tests that would be helpful?

tpolecat:
Nope, that's fine. If it's working for you that's good enough for me. Thanks!

PR #649, opened 2017-12-27
Use cats 1.0.0

BlackVegetable:
(Hoping travis will run these tests for me)

Cats 1.0.0 is now live! circe on master now uses it. I'm not sure if we'll need to update circe at the same time here or if that can be done independently.


tpolecat:
Thanks! Incorporated into #650

PR #651, opened 2018-01-02
[cats-1.0.0] Update Travis cache setup, bump to circe 0.9

guersam:
_This PR is for cats-1.0.0 branch #650_

We've been still caching `~/.sbt/0.13` after migrating to sbt 1.0. I'm not sure it's the most significant cause of increaset build time, but it must be helpful.

Bumped to circe 0.9 as well, including example docs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,8 @@ cache:
   - $HOME/.sbt/launchers
   - $HOME/.ivy2
   - $HOME/.coursier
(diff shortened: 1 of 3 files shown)


guersam:
Looking at recent build times, the spike seems somewhat random and caching is not significant...

tpolecat:
Brilliant, thanks!

PR #653, opened 2018-01-04
Update fs2 to 0.10.0-M11

guersam:
_This PR is for cats-1.0.0 branch #650_

Updated to FS2 0.10.0-M11 and Scala 2.11.12.

In FS2, `Stream#run*` are deprecated in favor of `Stream#compile{.drain, .toVector, .toList, ...}`. Should we do something similar about `Update#run`? At this point, `.run` looks a bit misleading to me.
https://github.com/functional-streams-for-scala/fs2/blob/v0.10.0-M11/docs/migration-guide-0.10.md#compiling--interpreting-streams

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ jdk:
   - oraclejdk8
 
 scala:
-  - 2.11.11
+  - 2.11.12
   - 2.12.4
 
 env:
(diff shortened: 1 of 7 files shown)


guersam:
BTW there are frequent CI failure in FS2 after M11. Maybe we need to wait for another milestone.

tpolecat:
Thanks, this seems good. I'm open to deprecating `Update.run` and giving another name if you can think of a good one.

I'm not worried about CI failures until we're off milestone dependencies. At that point everything needs to work ;-)

guersam:
Thanks, opened #654 to discuss the API alignment.

PR #659, opened 2018-01-09
Align API with cats and fs2 (fixes #654) 

### LINKED ISSUE #654: Align API with cats/fs2
There have been several API changes in cats, cats-effect and fs2. As we are still in milestone phase, it'd be helpful for newcomers to keep the names as consistent as possible with our dependencies.

Things to consider (maybe incomplete):
- `Update#fail` -> `Update#raiseError` ([MonadError#raiseError](https://typelevel.org/cats/api/cats/MonadError.html#raiseError[A](e:E):F[A]))
- `Update(0)#run` -> `Update(0)#compile` ([Stream#compile](https://github.com/functional-streams-for-scala/fs2/blob/v0.10.0-M11/docs/migration-guide-0.10.md#compiling--interpreting-streams)): might be confusing because `Stream#compile` needs a consequent combinator unlike deprecated `Stream#run`
- deprecate `Query(0)#process` in favor of `Query(0)#stream`
- `{Kernel, Composite}#zip` -> `{Kernel, Composite}#product` ([Semigroupal#product](https://typelevel.org/cats/api/cats/Semigroupal.html#product[A,B](fa:F[A],fb:F[B]):F[(A,B)])) or `Kernel.product -> Kernel.zip`

guersam:
This PR introduces following API changes to keep it
consistent with cats and fs2:

 Before               |    After
-----------------------  | ----------------------
 `Update(0)#run`         | `Update(0)#compile`
 `Update(0)#updateMany*` | `Update(0)#compileMany*`
 `Query(0)#process*`     | `Query(0)#stream*`
 `HC.process`            | `HC.stream`
 `HRS.process`           | `HRS.stream`
 `Composite#zip`         | `Composite#product`
 `Kernel#zip`            | `Kernel#product`

The old ones are deprecated.

### DIFF
--- a/modules/bench/src/main/scala/doobie/bench/select.scala
+++ b/modules/bench/src/main/scala/doobie/bench/select.scala
@@ -49,7 +49,7 @@ class bench {
   def doobieBenchP(n: Int): Int =
     sql"select a.name, b.name, c.name from country a, country b, country c limit $n"
       .query[(String,String,String)]
-      .process
+      .stream
       .list
       .transact(xa)
       .map(_.length)
--- a/modules/core/src/main/scala/doobie/hi/connection.scala
+++ b/modules/core/src/main/scala/doobie/hi/connection.scala
@@ -67,7 +67,16 @@ object connection {
    * action, and return results via a `Stream`.
    * @group Prepared Statements
    */
+  @deprecated(message = "Use `stream`", since = "0.5.0")
   def process[A: Composite](sql: String, prep: PreparedStatementIO[Unit], chunkSize: Int): Stream[ConnectionIO, A] =
+    stream(sql, prep, chunkSize)
+
+  /**
+   * Construct a prepared statement from the given `sql`, configure it with the given `PreparedStatementIO`
+   * action, and return results via a `Stream`.
+   * @group Prepared Statements
+   */
+  def stream[A: Composite](sql: String, prep: PreparedStatementIO[Unit], chunkSize: Int): Stream[ConnectionIO, A] =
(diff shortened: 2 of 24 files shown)


wedens:
I don't like this compile* naming for some reason.

guersam:
I respect your feeling. Do you have any suggestion? At first we considered `.single` and `.many` in #654 but found out that it's quite misleading in case of `deleteWhere(cond).single`, when many rows match the condition.

guersam:
Maybe we don't need to follow fs2's API change and just stick to `run*`, because there are still `Kleisli#run` and `State#run` in cats. In that case, I'd suggest renaming `updateMany` to `runMany`.

wedens:
> I'd suggest renaming updateMany to runMany.

👍

guersam:
The initial intention was to keep the API consistent with fs2. And unlike Kleisli and State, database access involves side effect and `run` might sound misleading, even though a doobie program is just a pure Free structure, especially because `run` is a method in `Update`.

@wedens could you elaborate your thought a bit more? I have no strong opinion regarding `run` vs `compile` and would like to get more feedback or different idea.

tpolecat:
Is there any consensus on where to go with this?

guersam:
Not yet, I like either way as long as it's consistent enough. Do you have any preferences?

tpolecat:
This is going to be hard to untangle because of all the changes I made in the meantime, and I think really `process` → `stream` is the only important rename so I'll do that myself on another branch. Thanks for this though, we can keep thinking about `run`. I can do `zip` → `product` on the "split" branch.

PR #661, opened 2018-01-11
Use `IO.ioEffect` to find `Effect[IO]` in test modules

aeons:
This uses `IO.ioEffect` to find the `implicit val M: Effect[IO]` in `IOChecker` and `IOAnalysisMatchers`. Without this, I had some runtime issues with uninitialized fields when attempting to find the implicit `Effect`.

(also update sbt to 1.1.0 final)


tpolecat:
Thanks!

PR #665, opened 2018-01-15
postgres large object streaming with fs2

wedens:
Provides functions for working with `LargeObject` using fs2.

I'll add an example later (can do it in another PR).


tpolecat:
This is great, thanks!

PR #666, opened 2018-01-16
simplify `repeatEvalChunks`

wedens:
I've also removed `VoidStreamOps` syntax in `Strategy`. It's redundant and makes code less straightforward IMHO. I'll undelete it, if you think it's useful.


tpolecat:
Thank you! There's probably more simplification to be done elsewhere. The fs2 code started as a pretty direct port from scalaz-stream and it's kind of crufty.

tpolecat:
(The Stream syntax should go away altogether if you're looking for another thing to delete.)

PR #667, opened 2018-01-17
WIP - split up meta and composite

### LINKED ISSUE #4: Split Atom
There is no reason in principle why every readable type needs to be writable; split `Atom` into read/write pairs (like Argonaut does).

tpolecat:
This is a change I have been putting off for a long time but I think it's a good simplification that aligns with circe and other serialization libs.

- It splits `Meta` into `Get` + `Put` and splits `Composite` into `Read` + `Write`.
- `Composite` is gone. `Meta` remains, but only for introducing symmetric `Get`/`Put` pairs, and for deriving simple mappings since it's used pretty commonly.

This is probably too much to introduce in 0.5 since it's such a big release already. Maybe 0.6

Remaining to do:

- [x] delete `Kernel`
- [x] update all scaladoc and comments
- [x] update book
- [x] more tests
- [x] ~~look into making `Fragment` derivation stack safe since it only needs `Put` now.~~ (do in a follow-up PR)
- [x] rename methods as needed to align with typeclass names
- [x] upgade to cats 1.1 so we can remove `ContravariantCoyoneda`.
- [x] 0.6.x release notes
- [x] Remaining stuff in the tut doc.

Resolves #4 (!)


tpolecat:
Ok this is now (correctly) failing because it fails bincompat checks.

tpolecat:
Any last words?

unoexperto:
Yeah, please publish transition guide in the documentation :)

tpolecat:
Also need to update the infographic.

modules/docs/src/main/tut/docs/12-Custom-Mappings.md
```diff
@@ -62,8 +62,8 @@ HPS.set(1, ("foo", 42))
 // Or leave the 1 out if you like, since we usually start there
 HPS.set(("foo", 42))
 
-// Which simply delegates to the Composite instance
-Composite[(String,Int)].set(1, ("foo", 42))
+// Which simply delegates to the Write instance
+Write[(String,Int)].set(1, ("foo", 42))
 ```
 
 **doobie** can derive `Composite` instances for primitive column types and options thereof, plus tuples, `HList`s, shapeless records, and case classes whose elements have `Composite` instances. These primitive column types are identified by `Meta` instances, which describe single-column mappings.
```
asoltysik: It looks like there are some leftovers of `Composite` in the docs

modules/docs/src/main/tut/docs/05-Parameterized.md
```diff
@@ -158,8 +158,8 @@ When reading a row or setting parameters in the high-level API, we require an in
 `Composite` instances are derived automatically for column types (and options thereof) that have `Meta` instances, and for products of other composites (via `shapeless.ProductTypeclass`). We can summon their instances thus:
```
asoltysik: also in this file

modules/docs/src/main/tut/docs/12-Custom-Mappings.md
tpolecat: Ah yeah I need to update the doc. Thanks.

tpolecat:
Ok the doc is updated, including a complete rewrite of the custom mappings chapter.

PR #669, opened 2018-01-18
monix and fs2 versions upgrade & documentation update

radusw:


tpolecat:
Unfortunately fs2 RC1 has a critical bug so we're stuck on M11 for now. 👍 for the Monix update though.

tpolecat:
For reference the fs2 issue is https://github.com/functional-streams-for-scala/fs2/issues/1070

wedens:
M11 is also affected by the bug

pchlupacek:
@tpolecat can you do me a favor and please try dobie build against the fs2 branch bugfix/translate2 ? That shall have translate fixed, Just want to verify there are no corner cases that are not covered by current tests. Thanks

tpolecat:
@pchlupacek hey @ChristopherDavenport just tried it and all the tests pass.

modules/docs/src/main/tut/migration.md
```diff
@@ -28,12 +28,13 @@ Doobie imports have been split into normal and implicit slices, to mirror the pa
...
 ## Data Types
 
-The IO data types provided by fs2 and doobie have been subsumed by `cats.effect.IO` or any data type with a `cats.effect.Aync` instance, such as `monix.eval.Task`.
+The IO data types provided by fs2 and doobie have been subsumed by `cats.effect.IO` or any data type with a `cats.effect.Async` instance, such as `monix.eval.Task`.
 
 | 0.4.x Data Type      | 0.5.x Data Type  |
 |----------------------|------------------|
 | `fs2.Task`           | `cats.effect.IO` |
 | `doobie.util.IOLite` | `cats.effect.IO` |
+| `monix.eval.Task`    | `cats.effect.IO` |
```
wedens: This table row contradicts with what's written above the table :) If someone was using `monix.eval.Task`, it doesn't need to be changed to `cats.effect.IO`.

modules/docs/src/main/tut/docs/03-Connecting.md
```diff
@@ -145,31 +145,22 @@ Currently cats has no typeclass for monads with **effect-capturing unit**, so th
...
+Or, with Hikari:
+```
+import doobie.hikari.HikariTransactor
+import scala.concurrent.Future
 
-  def suspend[A](fa: => Task[A]): Task[A] =
-    Task.suspend(fa)
-}
+// constructing it is a side-effect!
+val hxa: Future[Transactor[Task]] = HikariTransactor
```
wedens: Why do you need to convert it to `Future` if you use `Task`?

radusw: Yes, I wanted to point what that constructing the Hikari transactor is a side effect

tpolecat: I'd rather not introduce `Future` here. I don't quite understand.

PR #693, opened 2018-03-16
Add Partial Constructor for pgEnums

ChristopherDavenport:
Partial Constructor for pgEnums, as discussed in gitter.


modules/postgres/src/main/scala/doobie/postgres/Instances.scala
```diff
@@ -160,6 +160,13 @@ trait Instances {
   def pgEnumString[A: TypeTag](name: String, f: String => A, g: A => String): Meta[A] =
     enumPartialMeta(name).xmap[A](f, g)
 
+  /**
+   * Construct a `Meta` for values of the given type, mapped via `String` to the named PostgreSQL
+   * enum type with tranparent partiality.
+   */
+  def pgEnumStringOpt[A: TypeTag](name: String, f: String => Option[A], g: A => String): Meta[A] =
```
wedens: I think `PartialFunction[String, A]` is more convenient than `String => Option[A]`:

```scala
pgEnumStringOpt("my_enum", {
    case "A" => A
    case "B" => B
  }, {
    case A => "A"
    case B => "B"
  })
```

tpolecat:
I think this looks fine, sorry for the delay. Can you add a line to `CHANGELOG.md`? Don't worry about the wording, I will edit the notes before the next release. Just want to be sure you're on the board.

tpolecat:
Thanks!

PR #696, opened 2018-03-18
Allow clients to explicitly derive generic Composites

srp:
I have a project where I have case classes which get used in doobie queries in
multiple different files. The case classes need assorted Meta instances to build
a Composite. Currently that means that each file that uses the case classes with
doobie has to manage making sure that the right Metas are in scope. Essentially
it's leaking the implementation into the other layers of the code.

It would be nice to define the Composite once for each case class in a companion
object, much like is done with circe. However, the following seems to compile
recursively and results in a null pointer exception:

    case class MyThing(...)
    object MyThing {
      implicit val composite: Composite[MyThing] = implicitly[Composite[MyThing]]
    }

The implicit "generic" function is not caller friendly, as you have to specify
the G hlist.

This commit adds the ability to do the following instead:

    case class MyThing(...)
    object MyThing {
      implicit val composite: Composite[MyThing] = Composite.ForGeneric[MyThing].derive()
    }

### DIFF
--- a/modules/core/src/main/scala/doobie/util/composite.scala
+++ b/modules/core/src/main/scala/doobie/util/composite.scala
@@ -79,6 +79,15 @@ object composite {
 
     def apply[A](implicit A: Composite[A]): Composite[A] = A
 
+    final case class ForGeneric[F]() {
+      def derive[G]()(implicit gen: Generic.Aux[F, G], G: Lazy[Composite[G]]): Composite[F] =
+        new Composite[F] {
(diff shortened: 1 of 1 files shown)


tpolecat:
(I have an open PR against srp's branch [here](https://github.com/srp/doobie/pull/1) with proposed tweaks.)

tpolecat:
Oh just one more thing, can you add a line to `CHANGELOG.md`?

PR #707, opened 2018-04-06
add imap to meta

goedelsoup:


tpolecat:
Thanks for your work here, but unfortunately this change breaks binary compatibility.  I don't seen an obvious way to fix it but if you can figure out a way I will certainly accept the change.

This is subsumed by #667 which will be the first PR on 0.6 so I'm closing this for now, but you are welcome to re-open if you find a workaround.

PR #709, opened 2018-04-11
Implement stack-safe A=>List[Any] conversion

### LINKED ISSUE #426: Stack overflow with `Fragments`
Hi, I'm working with the new `Fragments`, and I found a problem with the following code:

```scala
import $ivy.`org.tpolecat:doobie-core-cats_2.11:0.3.1-M3`

import cats.data.NonEmptyList
import cats.implicits.catsSyntaxList
import doobie.imports._
import doobie.util.fragments._

def query(ids: Option[NonEmptyList[Long]]): Query0[Long] = {
  val q: Fragment = fr"""
    SELECT
      id
    FROM table
  """ ++ Fragments.whereAndOpt(ids.map(Fragments.in(fr"id", _)))
  q.query[Long]
}

query(List.range(0L,10000L).toNel).vector
```
This throws the following exception:
```
java.lang.StackOverflowError
  scala.runtime.BoxesRunTime.boxToInteger(BoxesRunTime.java:65)
  doobie.util.atom$Atom$$anon$3$$anonfun$6.apply(atom.scala:47)
  doobie.util.atom$Atom$$anon$3$$anonfun$6.apply(atom.scala:47)
  doobie.util.composite$LowerPriorityComposite$$anon$6$$anonfun$11.apply(composite.scala:109)
  doobie.util.composite$LowerPriorityComposite$$anon$6$$anonfun$11.apply(composite.scala:109)
  doobie.util.composite$Composite$$anon$3$$anonfun$5.apply(composite.scala:56)
  doobie.util.composite$Composite$$anon$3$$anonfun$5.apply(composite.scala:56)
  doobie.util.composite$Composite$$anon$3$$anonfun$5.apply(composite.scala:56)
  doobie.util.composite$Composite$$anon$3$$anonfun$5.apply(composite.scala:56)
  ...
```
I wonder if this can be resolved without limiting the length of `ids`.

nigredo-tori:
Fixes #426 

This is the implementation of a solution proposed [here](https://github.com/tpolecat/doobie/issues/426#issuecomment-379656799), with the following differences:
1. Made `Serializer` operate on ordinary functions (by matching them to its constructors) as a hack to preserve binary compatibility. This should be removed in 0.6.0 to restore exhaustivity.
2. Added `Suspend` constructor to preserve laziness.
3. Renamed `Divide` to `Contramap2` to match `ContravariantMonoidal`.

Compatibility:
1. Binary compatibility seems to hold.
2. Changed operand processing order in `Composite`. This should not matter for pure functions and data.
3. Some of the combinators in `Composite` (those that didn't have `Lazy` arguments) now immediately evaluate `.toList` in operands. I don't see this breaking someone's code unless they are somehow manually tie `Composite` in a knot, in which case  adding a `Serializer.Suspend` should solve the problem.

Possible improvements:
1. Add documentation for `Serializer`. I didn't bother since for now it is a minor utility class, but it might be useful for someone.
2. Add some constructor functions. Again, I didn't do that yet since the type is intended for internal use.
3. Add instances (at least for `ContravariantMonoidal`) and tests for them.

@tpolecat, how fleshed out do you want this to be?


tpolecat:
Can you add a test for `Fragment` showing that the failing case in #426 no longer fails?

nigredo-tori:
1. Renamed `Serializer` to `SafeFold`, parameterized it on its output type.
2. Made subclasses private, added some combinators.
3. Separated dirty conversion to/from functions, made them package private - this way they will be easy to remove. This also restores exhaustivity.
4. Switched to `SafeFold` in `Kernel`.

I had to use effectful `SafeFold`s in `Kernel`, which feels dirty. I've added some aliases to at least make this explicit. In the future we can write something like
```scala
def foreach[A, R](s: SafeFold[A, R])(a: A)(f: R => Unit): Unit
```
to make this more explicit, but with the way `Kernel` is set up there is no way I can think of to use something like that.

I also switched to a left-to-right traversal of the structure to minimize the surprising behavior. This still keeps `toList` performance quadratic - adding a separate right-to-left fold in 0.6.0 will help with that, but with the function hacks this requires too much boilerplate.

I've experimented with `ContravariantMonoidal` instance, but ultimately opted out of it since the syntax it provides doesn't work if the type parameter is not at the last position.

Some other possible instances:
```scala
Profunctor[Serializer]
Arrow[Serializer]
Monoid[A, R]
Applicative[Serializer[A, ?]]
```
However I didn't find a use case for those yet.

A couple of interesting properties of the `SafeFold` structure in comparison to `A => List[Any]`:
- The number of produced items does not depend on the data. This means that, if we rely on `SafeFold[A, Any]` as a replacement for `A => List[Any]`, we can calculate `width` based solely on this conversion.
- It allows expressing hylomorphisms, so we don't need the temporary structure, unlike the current `Composite` implementation.

nigredo-tori:
Actually, it might be beneficial to do something like this:
```scala
def apply[A, R](sa: SafeFold[A, R])(a: A): AppliedSafeFold[R]
```
, where `AppliedSafeFold` is a simple case class, which can then be `Foldable`. I'll try this later.

nigredo-tori:
I'm closing this PR in favor of #713. This implementation is dirty and slow, and cleaning it up would require too much time for something that will go in the bin in the close future.

PR #713, opened 2018-04-21
Make Write stack-safe

nigredo-tori:
### Overview ###
This changes `Write` into an ADT, which we then traverse in a stack-safe fashion.

### Motivation ###
#426

### Implementation ###
Along the lines of #709, but without the dirty pattern-matching trick. `Write` itself was rewritten instead of using a separate type. The lazy `Suspend` constructor was removed - all the lazy values in the old `Write` implementation were immediately forced anyway, so there is no need to account for laziness.

Since the tail-recursive implementation replaces stack with a `List` accumulator, it is slower than the old implementation (2 times slower, according to my benchmark). That's why I kept the original stack-based traversal algorithm, using it for subtrees with height below a certain threshold. This avoids the performance penalty for most structures, while keeping stack safety. Since having this duplication adds more code, I've only used this approach for "hot" methods (`unsafeSet`, `unsafeUpdate`).

### Benchmarks ###
A benchmark for `unsafeSet` is provided with this merge request for comparison with the old implementation (which is also included for convenience). The results on my machine (for a medium-sized structure):
```
Benchmark          Mode  Cnt      Score     Error  Units
unsafeSet         thrpt  100  15493.144 ? 257.413  ops/s
unsafeSetNoStack  thrpt  100   8643.263 ?  53.938  ops/s
unsafeSetOld      thrpt  100  13155.877 ? 199.162  ops/s
```
There's a slight increase in performance in this case. I expect that for structures producing large, skewed trees (e.g. for a case class with 40+ members with the default stack depth limit) performance will start creeping down to that of the tail-recursive algorithm, but for other cases it should remain close to the original implementation.

### Possible drawbacks ###
`Write` can't use custom `unsafeSet`/`unsafeUpdate` implementations anymore, which may limit some exotic instances.

### DIFF
--- a/modules/bench/src/main/scala/doobie/bench/write.scala
+++ b/modules/bench/src/main/scala/doobie/bench/write.scala
@@ -1,0 +1,111 @@
+// Copyright (c) 2013-2018 Rob Norris and Contributors
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package doobie.bench
+
+import java.sql.{
+  DriverManager,
+  PreparedStatement
+}
+import cats.implicits._
+import doobie.util._
+import org.openjdk.jmh.annotations._
+
+// Some huge structure
+
+final case class L1(
+  a: Int,
+  b: Option[Int],
+  c: Int,
+  d: Option[Int],
+  e: Int,
+  f: Option[Int],
+  g: Int,
+  h: Option[Int],
+  i: Int,
+  j: Option[Int]
+)
+
+final case class L2(
+  a: L1,
+  b: L1,
+  c: L1,
+  d: L1,
+  e: L1,
+  f: L1,
+  g: L1,
+  h: L1,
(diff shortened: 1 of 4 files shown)


tpolecat:
I agree this looks better. I hope to get the 0.6 branch bootstrapped this week so I'll have a closer look. Thanks for your work on this!

tpolecat:
Can you rebase and re-open against 0.6.x?

tpolecat:
I don't have the option to re-open, sorry. So something doesn't line up.

nigredo-tori:
@tpolecat, sorry, apparently GitHub requires the branch to be at the state it was before the PR was closed. I've pushed the old commit back, so you should be able to reopen this now. After that I'm going to push the rebased changes again.

nigredo-tori:
Ok, the rebase is done.

tpolecat:
I merged an alternative way to do this in 0.7.x so I'm closing this. Thanks!

PR #716, opened 2018-04-30
Add 'pure' to generated algebras

wedens:


tpolecat:
👍

tpolecat:
Looks like you removed it from `free` instead of `postgres`.

wedens:
Yeah. I've fixed it.

Issue #706, opened 2018-04-05
Update0.updateMany swallows error flags in the returned int[] from Statement.executeBatch


mfelsche:
The docs for [Statement.executeBatch](https://docs.oracle.com/javase/8/docs/api/java/sql/Statement.html#executeBatch--) from the JDBC Api explicitly state that possible values are: >=0 which is the update count, `SUCCESS_NO_INFO (= -2)` or `EXECUTE_FAILED (=-3)`. The current `updateMany` implementation sums the int array and thus treating those values as update counts, which leads to incorrect results. The returned int could even become negative. 

The sum is applied here: https://github.com/tpolecat/doobie/blob/series/0.5.x/modules/core/src/main/scala/doobie/hi/preparedstatement.scala#L72

To me it would make sense to return the int array as list and not sum it up in the first place. Another improvement (which i wouldn't favor) would be to filter out negative values and treat them as rows not being updated, which would not convey the whole information, but at least a little more:

```scala
  def addBatchesAndExecute[F[_]: Foldable, A: Composite](fa: F[A]): PreparedStatementIO[Int] =
    fa.toList.foldRight(executeBatch)((a, b) => set(a) *> addBatch *> b).filter(_ > 0).map(_.sum)
```

tpolecat:
Oops, good catch. Thanks.

tpolecat:
I'm unable to come up with a test case that returns a negative number rather than throwing an exception. Do you have an example?

tpolecat:
Cool I'll try that one.

tpolecat:
Hm yeah I'm getting a `BatchUpdateException` from that as well.

Daenyth:
Is this still a problem?

ayoub-benali:
yes I guess because there isn't yet a way to get the error messages if some rows aren't inserted.

nigredo-tori:
I've came here from the `Update.updateMany` scaladoc. It has a sentence that reads like this:

> Note that failed updates are not reported (see https://github.com/tpolecat/doobie/issues/706).

This sounds terrifying! It also sounds suspicious - it should use Java APIs, and what Java API would miss an opportunity to throw an exception? :smile: So I did some digging - and, apparently, `executeBatch` shouldn't **return** any `EXECUTE_FAILED` values. Consider this fragment of the [`executeBranch` Javadoc](https://docs.oracle.com/en/java/javase/11/docs/api/java.sql/java/sql/Statement.html#executeBatch()):

[...]

PR #719, opened 2018-05-12
interim 'fix' for #706

tpolecat:
This changes the totaling for `updateMany` to ignore failures, which are reported as negative numbers as reported in #706. This is a temporary fix so the issue remains open for a future milestone.


modules/core/src/main/scala/doobie/hi/preparedstatement.scala
```diff
@@ -65,11 +65,15 @@ object preparedstatement {
...
+   * Add many sets of parameters and execute as a batch update, returning total rows updated. Note
+   * that failed updates are not reported (see https://github.com/tpolecat/doobie/issues/706). This
+   * API is likely to change.
    * @group Batching
    */
   def addBatchesAndExecute[F[_]: Foldable, A: Composite](fa: F[A]): PreparedStatementIO[Int] =
-    fa.toList.foldRight(executeBatch)((a, b) => set(a) *> addBatch *> b).map(_.sum)
+    fa.toList
+      .foldRight(executeBatch)((a, b) => set(a) *> addBatch *> b)
+      .map(_.foldLeft(0)((acc, n) => acc + (n max 0))) // treat negatives (failures) as no rows updated
```
ChristopherDavenport: Personally debating whether `foldMap` is more clear here if I had written it. Why did you favor this explicit fold implementation?

tpolecat: Yeah that was my thought as well but JDBC returns an `Array[Int]` so there's no `Foldable` instance. Oh well.

PR #722, opened 2018-05-14
add Fragment.execWith for custom handling

### LINKED ISSUE #710: Allow transforming result from query to arbitrary type
Currently there is no way of dynamically accessing the results from a query without first specifying its result type. After reading the "dynamic" example [here](https://github.com/tpolecat/doobie/blob/series/0.5.x/modules/example/src/main/scala/example/Dynamic.scala), it seems posible to create a `Query` from a `Fragment`, keeping all the interpolated values without specifying a type for the result.
The API could provide a way of transforming the result set via `HC`.

Rob's comment on Gitter was:
> Fragment needs to expose `sql` and a `PreparedStatementIO` that sets the statement arguments. You could then say `HC.prepareStatement(f.sql)(f.prepare *> yourStatementHandler)` to do whatever you want.

If I am not mistaken, we should be able to provide our own `ResultSetIO[A]` so we can get our `ConnectionIO[C]`.

Let me know if need to be more specific on this. I would love to provide a PR but I don't think I am that confident to do so yet.

tpolecat:
This provides a way to intercept the statement handler for a fragment and do whatever you like.

Resolves #710

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ scala:
 
 env:
   global:
-    - MAIN_SCALA_VERSION=2.12.5
+    - MAIN_SCALA_VERSION=2.12.6
 
 before_install:
  - export PATH=${PATH}:./vendor/bundle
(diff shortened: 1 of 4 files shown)


tpolecat:
Also some build hacking.

PR #723, opened 2018-05-15
update migration guide

### LINKED ISSUE #614: List of changes to switch client from 0.4 to 0.5
I just switched my project to 0.5 and I think list of thing I've done may help others. Please add things you had to change that I didn't mention.

- If you use `streamz-converter` to produce akka streams you need to upgrade dependency to `0.9-M1`.
- doobie dependencies `doobie-XXX-cats` no longer have postfix `-cats`.
- I upgraded to Scala 2.11.11 because `sequenceU` is gone in new cats. All occurrences in the code should be replaced with `sequence`.
- `fs2.Task` is replaced with `cats.effect.IO`
- `unsafeRun` is replaced with `unsafeRunSync`
- `unsafeRunAsyncFuture` is replaced with `unsafeToFuture()`
- `exceptSome` is gone. I have to replace with with `exceptSql`. Not sure what to do with non-sql exceptions.
- `doobie.hikari.hikaritransactor.HikariTransactor` is moved to `doobie.hikari.HikariTransactor`
- `doobie.postgres.PGTypes` is moved to `doobie.postgres.Instances`
- `Meta.nxmap` is deprecated in favor of `Meta.xmap`
- `DataSourceTransactor` is replaced with `doobie.util.transactor.Transactor.fromDataSource`
- Explicit import of string interpolator should be replaced from `doobie.syntax.string.SqlInterpolator` to `doobie.syntax.string._`

tpolecat:
Adds a few more notes to the migration guide. I think it's mostly fine. 

resolves #614

### DIFF
--- a/modules/docs/src/main/tut/migration.md
+++ b/modules/docs/src/main/tut/migration.md
@@ -50,6 +50,7 @@ The IO data types provided by fs2 and doobie have been subsumed by `cats.effect.
 | `fs2.Task`           | `cats.effect.IO` |
 | `doobie.util.IOLite` | `cats.effect.IO` |
 
+Note that the "end of the world" operation for `IO` is called `unsafeRunSync`.
 
 ## Typeclasses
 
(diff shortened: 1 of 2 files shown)


tpolecat:
Ugh damnit the travis build exits successfully after the first call to `sbt` so it's not building the tut doc. What on earth.

tpolecat:
Victory!

PR #725, opened 2018-05-16
simple orm

tpolecat:
This adds a super super simple ORM example because people keep asking if it's possible. It is, in simple cases like this.


Igosuki:
Great stuff

PR #732, opened 2018-05-22
Add postgres-circe project with Circe Instances for Postgres

ChristopherDavenport:
Based on the custom type work. I can backport this to Meta for 0.5.x if there is interest but since that was last scheduled I figured I would target 0.6.x for the initial PR.

This implements type constrained `Get` and `Put` for `Json`. Then functions that have `TypeTag` constrained(`T` variants) and unconstrained functions from given types to get their derived `Put` or `Get`


tpolecat:
This looks good. Can you add an item to `CHANGELOG.md`? Maybe start another `####` section for "other changes" and start a bullet list. We'll go back and clean it up before releasing.

ChristopherDavenport:
Current Solution Description: Package json and jsonb instances into individual packages so that they are not brought into scope together.

Advantages: Easy Imports For Most Use Cases. Solves initial problem.

Problems: Instances are not Universally Coherent, Different imports lead to different behavior. Tables with both Json and jsonb columns will need to be created manually and it will be unpleasant.

Other solutions considerred, Pick a winner as the default encoding and make special functions that encode via the other type as they won't be picked up automatically. But which one to pick and why?

Igosuki:
There is a conflict problem here that I have faced implementing my own Circe Json Meta, where Map[String, String] (totally a valid encodable) already has a Meta that uses PG's Hstore extension. 
The workaround is having Json circe type fields which is not really acceptable. 
So my question is what do you plan to do for circe encodable fields that have postgre metas in doobie.postgres.implicits ?

ChristopherDavenport:
You have to explicitly call that. It does not generate the `Meta` implicitly for an an `Encoder[A]`, only for `Json`

Igosuki:
I see, makes sense, I have code for that too though if you are interested.

modules/postgres-circe/src/main/scala/doobie/postgres/circe/Instances.scala
```diff
@@ -0,0 +1,81 @@
...
+import io.circe._
+import io.circe.jawn._
+import io.circe.syntax._
+import doobie.util._
+import org.postgresql.util.PGobject
+import scala.reflect.runtime.universe.TypeTag
+
+object Instances {
+
+  trait jsonbInstances {
```
tpolecat: Can we capitalize these traits? `JsonbInstances` and `JsonInstances`.

tpolecat:
Thanks!

PR #733, opened 2018-05-22
MiMa version Check improvement

ChristopherDavenport:
Trying to get this available as mentioned in the PR. Have a similar issue running in typelevel/cats#2265. 

This seems to apply correctly. But will continue to check to see if it needs to be further improved.


tpolecat:
This is awesome! Thanks!

PR #736, opened 2018-05-23
WIP/experiment: stacksafe fragment

tpolecat:
@nigredo-tori this is another approach that doesn't change `Write`, it just changes the encoding of `Fragment` to use a `List[(A, Put[A])]` which stays flat on concatenation. Yours may end up looking better but I thought I would see what this approach looked like.

### DIFF
--- a/modules/core/src/main/scala/doobie/aliases.scala
+++ b/modules/core/src/main/scala/doobie/aliases.scala
@@ -4,6 +4,8 @@
 
 package doobie
 
+import shapeless.HList
+
 /** Mixin containing aliases for the most commonly used types and modules from doobie-core. */
 trait Aliases extends Types with Modules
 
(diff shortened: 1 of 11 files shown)


nigredo-tori:
Well, it's certainly less of a change then what I did ^_^. It is also more predictable performance-wise. Now that I think about it, in my implementation, a fragment with a lot of columns would result in a worst-case tree (perfectly skewed). I'll have to add some protection against that.

I also like the move away from a trait to something less opaque/verbose - even if we don't use this implementation, making something like this might clean up the `Fragment` code a bit:
```scala
sealed trait Fragment {
  protected val sql: String
  protected type A
  protected def a: A
  protected def ca: Write[A]
  protected def pos: Option[Pos]

  //...
}

object Fragment {
  private case class Impl[A](
    sql: String,
    a: A,
    ca: Write[A],
    pos: Option[Pos]
  ) extends Fragment { type T = A }

  //...
}
```

That said, I'm still irrationally drawn to the idea of building a `Write` instance in `StringContext` fashion. There's little use of the resulting instance with `Fragment`, but there might be a way to do something similar without providing the actual data. That's something for me to fiddle with when I get some free time.

PR #740, opened 2018-05-25
Analysis/testing cleanup

### LINKED ISSUE #729: Few things to clean up around analysis/testing
1. Provide a default `AnalysisReport` formatting implementation in `doobie.util.testing` to simplify using test frameworks other than scalatest and specs2. Moving the formatting implementation from `doobie-scalatest` to here seems to be the cleanest solution.
2. `doobie.util.yolo` has its own formatting/checking implementation. I would like to replace it with `AnalysisReport` and the default formatting implementation mentioned above.
3. The overloaded Checker.check methods in both testing frameworks should be unified like this:
    ```scala
    def check[A: Analyzable](a: A) = checkImpl(Analyzable.unpack(a))
    ```

Since we're breaking binary compatibility with 0.6.0, now would be the best time to fix these.

nigredo-tori:
Closes #729

1. Unify `check` methods.
2. Add two missing `checkOutput` methods. These are specialized, which goes against the spirit of the previous point. However I can't think of a good way to unify them without adding more boilerplate.
3. Move the formatting code to `doobie.util.testing`.
4. Remove custom formatting from `yolo`. This considerably changes `yolo`'s `check` output, making it more verbose. Please tell me if this is undesirable - I'll whip up something close to the original.
    Another concern here is that `yolo` syntax now requires `TypeTag`s for its type parameters. I'm not sure if this would lead to trouble for anyone.
5. Make the argument implicit in `doobie.util.testing.typeName`, cleaning up the call sites.
6. Add missing type ascriptions that were causing wartremover errors with scala 2.11. As an aside, it seems like [your Travis script is swallowing errors](https://travis-ci.org/tpolecat/doobie/jobs/383657723#L6234).


tpolecat:
Ugh why is it so goddamn hard to get a working Travis script?

tpolecat:
This looks great 👍 

I fixed the CI script and the problem in `Query` so if you rebase you should be ok.

nigredo-tori:
Rebase is done.

tpolecat:
This looks good to me. Ready on your end?

PR #742, opened 2018-05-27
simplify logging code

tpolecat:
This factors out `liftOnError` which is kind of like `pure.rethrow` but lets you observe the error. Kind of contrived but it removes some boilerplate.

### DIFF
--- a/modules/core/src/main/scala/doobie/util/update.scala
+++ b/modules/core/src/main/scala/doobie/util/update.scala
@@ -45,8 +45,12 @@ object update {
     private val now: PreparedStatementIO[Long] =
       FPS.delay(System.nanoTime)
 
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/doobie/util/update.scala
```diff
@@ -45,8 +45,12 @@ object update {
     private val now: PreparedStatementIO[Long] =
       FPS.delay(System.nanoTime)
 
-    private def fail[T](t: Throwable): PreparedStatementIO[T] =
-      t.raiseError[PreparedStatementIO, T]
+    implicit class MoreEitherOps[L, R](e: Either[L, R]) {
```
nigredo-tori: Seems to be a leftover duplicate.

PR #747, opened 2018-06-07
fix Text for nested case classes

lucidd:
This fixes the "diverging implicit" error when using `Text` with nested case classes.


tpolecat:
Thanks!

PR #753, opened 2018-06-15
Update to latest version of Magnolia

propensive:
This version (when it hits Maven Central) should support defining derivations which aren't defined on coproducts.


propensive:
Note that the tests don't currently compile with Magnolia 0.7.1 or 0.8.0, so I'm not sure what state the branch was in when it was last worked on...

tpolecat:
Oh cool, thanks! I will give it a go.

PR #759, opened 2018-07-23
Linebacker project

ChristopherDavenport:
Adds blocking transactors for `Linebacker[F]` or `DualContext[F]`.

Applies to a `Transactor` to return a new `Transactor`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ This file summarizes **notable** changes for each release, but does not describe
 
 ----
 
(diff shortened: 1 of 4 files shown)


nigredo-tori:
I don't think that the interpreter is the right place for this - switching contexts twice per `ConnectionOp` seems wasteful. What about moving this to strategy? Of course, you would have to rely on doobie bracket logic and manual shifts instead of nice linebacker combinators.

ChristopherDavenport:
Not really necessary with the new changes. 👍

PR #771, opened 2018-09-10
Updates, ContextShift, work on blocking pool

tpolecat:
This is an alternative approach to #769 which goes ahead and defines `ContextShift` for the doobie monads if there's a `ContextShift[IO]` in scope. @ChristopherDavenport had a concern about fairness so let's discuss here.


tpolecat:
Ok I went ahead and plumbed this down into the algebras and interpreter, and shifted all the blocking work onto a blocking pool. Kind of a big change but everything works except doc, which I need to update.

tpolecat:
Hm it's deadlocking when a `ConnectionIO` program includes a `.map` operation. Investigating.

tpolecat:
It was SI-9076 … adding `-Ydelambdafy:inline` fixed it.

"fixed"

tpolecat:
Transactors need to open connections on the blocking pool. Right now it's synchronous.

Also might want to start every transaction with `HC.shift` just to be sure nothing ever happens on unknown threads. Maybe it doesn't matter.

tpolecat:
Ok this is getting big now. Every `Transactor` constructor now takes a `ContextShift` and two `ExecutionContext`s, which are lifetime-managed so you want to get at them via `Resource`.

tpolecat:
I also need to fix tut to snarf output globally because stuff is getting printed on different threads now and `Console.withOut` doesn't handle that.

tpolecat:
What would we think about `driverManagerTransactor` just using the global EC everywhere since it's only useful for testing and you would never use it in production anyway? It saves having to pass those args and will make this gigantic diff a lot smaller.

ChristopherDavenport:
I don't mind that personally.

tpolecat:
Yeah I think that's a reasonable simplification.

aeons:
Wow, this spiralled out of control :)

Thanks for the hard work!

modules/docs/src/main/tut/docs/04-Selecting.md
```diff
@@ -202,7 +212,7 @@ And just for fun, since the `Code` values are constructed from the primary key,
     .query[(Code, Country)]
     .stream.take(5)
     .compile.toList
-    .map(_.toMap)
+    .map(_.toMap) // this causes a deadlock … why?
```
ChristopherDavenport: Does it still or was this before lambdafy?

tpolecat: Good catch, it was that REPL bug. I'll remove that comment.

aeons:
Is this ready to be released a milestone or something?

tpolecat:
Yes I will release a milestone when this is merged. Later today probably.

aeons:
Great, thanks!

modules/docs/src/main/tut/docs/14-Managing-Connections.md
```diff
@@ -36,99 +39,131 @@ These are the natural transformations that a `Transactor` provides:
...
 
-**doobie** provides several implementations, described below.
+### About Threading
+
+Starting with version 0.6.0 **doobie** provides an asynchronous API that delegates blocking operations to dedicated execution contexts *if you use the provided `Transactor` implementations*. To construct any of the provided `Transactor[M]`s (other than `DriverManagerTransactor`) you need
+
+- `ContextShift[M]`, which provides a CPU-bound pool for **non-blocking operations**. This is typically backed by `ExecutionContext.global`. If you use `IOApp` and interpret into `IO` this will be available for free.
+- An `ExecutionContext` for **awaiting connection** to the database. Because there can be an unbounded number of connections awaiting database access this should be a **bounded** pool.
+- A second `ExecutionContext` for **execution JDBC operations**. Because your connection pool limits the number of active connections it is usually fine for this to be an **unbounded** pool.
```
aeons: s/execution/executing/

modules/docs/src/main/tut/migration.md
```diff
@@ -7,7 +7,23 @@ position: 2
 
 # Upgrading to 0.6.x from 0.5.x
 
-The major change in 0.6.x is the restructing of `Meta` and `Composite` as described below. In preparation for upgrading please **fix all deprecation warnings** as all methods deprecated in 0.5.x have been removed.
+The major changes in 0.6.x are the addition of explicit threading controls for asynchronous operationn and the restructing of `Meta` and `Composite`, as described below. In preparation for upgrading please **fix all deprecation warnings** as all methods deprecated in 0.5.x have been removed.
```
aeons: restructing -> restructuring?

tpolecat:
Thanks @aeons and @ChristopherDavenport, I'll fix those typos and get a milestone out.

Issue #780, opened 2018-10-05
Diverging implicit expansion shapeless error for type parameterized queries


mcanlas:
I'll preface this by saying this is probably more related to `shapeless` than `doobie`, but I wanted to see if the ergonomics were intentional. And to just find out more, I'm not sure what is happening. Compare the following:

```scala
// does not compile
class MinimalBad[A: Meta, B: Meta] {
  def bad(a: A, b: B): Update0 =
    sql"""$a $b""".update
}
```

results in

```scala
diverging implicit expansion for type doobie.util.param.Param[A :: B :: shapeless.HNil]
[error] starting with method fromMeta in object Param
[error]     sql"""$a $b""".update
```

Contrast with

```scala
object MinimalGood {
  def good[A: Meta, B: Meta](a: A, b: B): Update0 =
    sql"""$a $b""".update
}
```

What's the difference?

mcanlas:
cc @Jacoby6000

Jacoby6000:
That doesn't fix anything, it just defers it to the thing that calls `MinimalBad.bad`. Something weird in scala might wind up making that okay, though. I don't know how scala's implicit resolution rules.

Jacoby6000:
Not sure if this is feasible with doobie, but I did a minimal recreation of this, and fixed it.

Minimal reproduction:
```scala
import shapeless._

trait Base[A]

trait Derived[A]

object Derived {
  implicit def fromBase[A: Base] = new Derived[A] {}
  
  implicit val hnil: Derived[HNil] = new Derived[HNil] {}
  implicit def hcons[A: Derived, B <: HList : Derived]: Derived[A :: B] = new Derived[A :: B] {}
}


class Bad[K: Base, V: Base] {
  def foo = implicitly[Derived[K :: V :: HNil]]
}
```

Fixed:
```scala
import shapeless._

trait Base[A]

trait Derived[A]

object Derived {
  implicit def fromBase[A: Base] = new Derived[A] {}
  
  implicit val hnil: Derived[HNil] = new Derived[HNil] {}
  implicit def hcons[A: Derived, B <: HList : Derived]: Derived[A :: B] = new Derived[A :: B] {}

  // this fixes the problem
  implicit def hconsFromBase[A: Base, B <: HList : Derived]: Derived[A :: B] = new Derived[A :: B] {}
}


class Bad[K: Base, V: Base] {
  def foo = implicitly[Derived[K :: V :: HNil]]
}
```

[...]

PR #782, opened 2018-10-09
Resolve diverging implicit expansion problem (#780)

Jacoby6000:
As far as I can tell, this solves the issue without introducing any new ambiguous implicit errors. I don't know how scala's implicit resolution works though. This could've broken things.


tpolecat:
Thanks! I will have a look after work.

Jacoby6000:
If this winds up looking good enough, I'll port it over to series 6 as well. Not that it's much work anyway though. lol

tpolecat:
Yep, thanks for the reminder. Looks good. Thanks!

tpolecat:
I'll port this to 0.6.x, thanks.

PR #796, opened 2018-11-12
Fix import fragments util in tutorial

iamquang95:


tpolecat:
This import works as-is ... the `doobie` package object defines

```scala
val Fragments = doobie.util.fragments
```

which is similar to handful of other module renames.

Are you suggesting we shouldn't provide the module alias?

tpolecat:
Closing for now.

PR #797, opened 2018-11-12
Port fix from #782 in to 6.x

### LINKED ISSUE #780: Diverging implicit expansion shapeless error for type parameterized queries

Jacoby6000:
This should resolve #780 in the 6.x series.

Moving priorities around didn't work, so I had to resort to using `shapeless.Lazy`.  It doesn't seem to be causing any new problems.

Also fixed a couple of typos in `put.scala` while I was poking around.

cc: @mcanlas

### DIFF
--- a/modules/core/src/main/scala/doobie/util/param.scala
+++ b/modules/core/src/main/scala/doobie/util/param.scala
@@ -6,7 +6,7 @@ package doobie.util
 
 import scala.annotation.implicitNotFound
 
-import shapeless.{ HNil, HList, :: }
+import shapeless.{ HNil, HList, ::, Lazy}
 
 /** Module defining the `Param` typeclass. */
 object param {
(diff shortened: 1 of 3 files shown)


Jacoby6000:
Reason this seems to cause issues in the first place, is that scala seems to put pretty low priority on implicit provided by the class.  Because of this, it tries to derive a `Put` instance from _literally any other source_ before trying to use `ParamHList`. Using `-Xlog-implicits` shows some crazy stuff going on.  Scalac goes as far as trying to derive put instances from meta instances before using the already existing param instances.  Seems to be because scalac puts the leap from `TC[B]` to `TC[B :: HNil]` at a very low priority.

tpolecat:
Thanks!

PR #801, opened 2018-11-29
Fix #800

### LINKED ISSUE #800: Freeze in .onSqlException, .exceptSql, etc. all functions in catchSql module
Given this code:

```scala
def example: ConnectionIO[Unit] = {
  connection.raiseError(new Throwable)
    .onSqlException(connection.rollback)
}
```

It freezes forever when running doobie on `scalaz-zio-0.3.1`'s `Task`

Replacing `onSqlException` with MonadError's `onError` behaves as expected, i.e. throws an exception:

```scala
.onError { case e: SQLException =>
  connection.rollback *> connection.raiseError(e)
}
```

It's possible that this line https://github.com/tpolecat/doobie/blob/8bcb53d536dafa8a25b0082f2441f6da2074a758/modules/core/src/main/scala/doobie/util/catchsql.scala#L21 may be the cause.

kaishh:


tpolecat:
So this does fix your issue?

kaishh:
It should, without a naked throw the error should stay in the `error` channel not become a panic.

PR #804, opened 2018-12-06
Update postgres, scala to 2.12.7 and cats 

coltfred:
Update the minor of scala 2.12, patch of postgres driver and minor cats.


tpolecat:
Thanks!

PR #810, opened 2019-01-01
Support compiling on Java 11

guymers:


.travis.yml
```diff
@@ -4,20 +4,21 @@ language: scala
...
+  - oraclejdk11
 
 scala:
   - 2.11.12
   - 2.12.7
 
 env:
   global:
-    - MAIN_SCALA_VERSION=2.12.6
+    - MAIN_SCALA_VERSION=2.12.7
```
guymers: Making this match the value in the `scala` list above means that the mima check is now running again. It seems that the `series/0.6.x` branch is not binary compatible with 0.6.0.

guymers:
Commit is included in https://github.com/tpolecat/doobie/pull/813

PR #813, opened 2019-01-03
Scala 2.13 support

guymers:
Scalac flags that don't exist in 2.13 have been removed.

Monix does not have a version published for 2.13 and sbt would not skip the docs subproject when 2.13 was excluded from `crossScalaVersions` so I commented out the dependency. There doesn't seem to be any compile errors?

 The `scala-collection-compat` library provides a `Factory` abstraction that can be used to support all Scala versions. But it seems to require callers to import `scala.collection.compat._` in order for to summon `Factory` instances built from `CanBuildFrom`. So instead I added a lightweight `FactoryCompat` abstraction. I am not sure it is safe to ignore the binary compatibility errors related to the implicit change however.


tpolecat:
Wow, thanks! This looks like it was a lot of work. Reviewing it now …

tpolecat:
Book Chapter 3 has a Monix example but it looks like someone took out the tut marker so it's just plaintext. We can leave it that way for now.

tpolecat:
I retargeted this for `series/0.7.x` because of the API change. This is fine, 0.7.0 will probably be the next version.

tpolecat:
Having a hell of a time making the JDK8/2.12 build finish. It slows down and stops after a while. I'll need to look into this more tomorrow.

guymers:
Okay cool, the `mimaBinaryIssueFilters` section can be removed then.

In order to run `docs/tutQuick` locally I had to bump `MaxMetaspaceSize` to 2gb.

tpolecat:
Yeah if you could remove the mima stuff that would be good. Thanks.

guymers:
Removed the mima exclusions.

Using G1 as the garbage collector seems to get `docs/tutQuick` to not hang under Java 8.

tpolecat:
> Using G1 as the garbage collector seems to get docs/tutQuick to not hang under Java 8.

Oh cool, I never would have figured that out. Thanks!

PR #822, opened 2019-01-08
HikariTransactor construction from HikariConfig

LMnet:
I added a convenient way to create a `HikariTransactor` from the given `HikariConfig`. This new constructor is useful for cases when a configuration is read separately from the creation of the application components. For example, configuration could be read with the Pureconfig library.

I'm not sure about branches. I added this pull request to the latest 0.7.x series, but this is a pretty small improvement and it could be done in the 0.6.x as well.


LMnet:
Maybe I know him in person, but this username is not familiar to me :)

PR #830, opened 2019-01-19
be community build friendly

SethTisue:
also use latest sbt. might as well

dbuild, the meta-build tool we use in the Scala community build, needs
`onLoad` for its own purposes, and gets confused when a build
overrides it like this.

(also, nbd, but I would also argue that it's maybe not that great an idea
for a open source build to alter sbt's default behavior like this,
contributors may be confused that things aren't working as they would
normally expect)


tpolecat:
I wasn't aware that changing `onLoad` was a weird thing to do.

SethTisue:
for the community build, it isn't strictly necessary that you merge this. it's just convenient because it would spare us from us from having to use our own doobie fork. it's really nbd either way

ChristopherDavenport:
I've been slowly removing it from mine as well. Its not as necessary now that scala-steward is around to help out. At least from my experience.

PR #837, opened 2019-02-04
Remove a majority of calls to throw

### LINKED ISSUE #836: find and destroy any remaining `throw`s
There are probably some remaining `throws` left over from prehistory. Get rid of them because they don't work well with zio.

guymers:
`RaiseError` has been added to the `Free` definitions and is used to define `raiseError` instead of throwing in a `delay`.

Replace `unsafeFromInt` methods with `fromIntF` which requires an `F` that can raise an error.

Update extension attempt methods to raise an error instead of throwing.

Throws in `Get` remain, removing them would probably have a performance impact. These exceptions will be handled in `primitive` as it now catches any exceptions and raises them in the effect.

A `temap` method has been added to `Get` to allow conversion functions to fail with an error message instead of having to throw an exception. Under the hood this just throws a `InvalidValue` invariant exception.

Unfortunately this will fail on 2.13 due to the test dependency on ZIO which does not yet have a published 2.13 version.

Closes #836


tpolecat:
Wow, this looks good, thanks! Is there an ETA on ZIO for 2.13?

guymers:
There is a PR https://github.com/scalaz/scalaz-zio/pull/546 but it is waiting on Monix.

I have a very hacky work around, I built the 2.13.0-M5 jars and put them in the `lib/` directory as they are only needed for tests.

PR #839, opened 2019-02-19
Support for Array[BigDecimal] with Postgres

nremond:
Adding support for Array[BigDecimal] with Postgres. 

Thank you.


tpolecat:
Aha, thanks!

nremond:
Thank you for the quick merge. 

In my app code, I have added the following implicits to make it usable with `scala.BigDecimal`. 
Would it be a good idea to have them here also?

```scala
implicit val bigDecimalMeta: Meta[Array[BigDecimal]] = Meta[Array[java.math.BigDecimal]]
      .timap(_.map(BigDecimal.apply))(_.map(_.bigDecimal))

implicit val optionBigDecimalMeta: Meta[Array[Option[BigDecimal]]] = Meta[Array[Option[java.math.BigDecimal]]]
      .timap(_.map(_.map(BigDecimal.apply)))(_.map(_.map(_.bigDecimal)))
```

tpolecat:
Yes, that would be fine if you'd like to add those in another PR.

PR #847, opened 2019-03-06
Use weak type tags for Analyzable

### LINKED ISSUE #846: Relax TypeTag requirement to WeakTypeTag for Analyzable
`TypeTag` is not flexible enough to support some use cases - like abstract type parameters. For example, in the following code `TypeTag[Foo[Id]]` doesn't exist:
```scala
final case class Foo[F[_]](foo: F[Int])
val fooQ: Query0[Foo[Id]] = sql"select 1".query

"foo" should "typecheck" >> {
  // error!
  fooQ must typecheck[Foo[Id]]
}
```
I feel like we can relax the requirement to `WeakTypeTag` for the testing machinery, if not for the whole codebase. Sure, it can make error messages less useful if the user doesn't provide the `WeakTypeTag`, but I think that the added flexibility is worth it.

nigredo-tori:
Closes #846

The change shouldn't break the existing code, since `TypeTag[A] <: WeakTypeTag[A]`.

### DIFF
--- a/modules/core/src/main/scala/doobie/util/testing.scala
+++ b/modules/core/src/main/scala/doobie/util/testing.scala
@@ -16,7 +16,7 @@ import doobie.util.analysis._
 import doobie.util.pretty._
 import doobie.util.pos.Pos
 import scala.Predef.augmentString
-import scala.reflect.runtime.universe.TypeTag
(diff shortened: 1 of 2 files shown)


nigredo-tori:
There are variance issues in Specs2's `Matcher` in Scala 2.11. Fixed the test.

PR #852, opened 2019-04-04
WIP: Quill Integration

tpolecat:
@deusaquilus here is the work so far


tpolecat:
Ok @deusaquilus I think the implementation is done but I don't know how to execute the `execute[Batch]ActionReturning` … how would I express this kind of query in Quill? I couldn't find an example in the doc.

tpolecat:
Guess I need to add the other database dialects too. Might make sense to move all this stuff to Quill, unclear. We can discuss when you get back from vacation.

tpolecat:
Thanks @ChristopherDavenport for getting CI to work. Next job is getting sbt-release to work. I'll give it a try this evening.

ChristopherDavenport:
These were the local release steps I used to make sure it was operational.

```scala
    releaseProcess := Seq[ReleaseStep](
      checkSnapshotDependencies,
      releaseStepCommand("mimaReportBinaryIssues"),
      inquireVersions,
      runClean,
      runTest,
      releaseStepCommand("docs/tut"), // annoying that we have to do this twice
      // setReleaseVersion,
      // commitReleaseVersion,
      // tagRelease,
      releaseStepCommandAndRemaining("+publishLocalSigned")//,
      // releaseStepCommand("sonatypeReleaseAll"),
      // releaseStepCommand("docs/publishMicrosite"), // TODO: re-enable for 0.7.0 final
      // setNextVersion,
      // commitNextVersion,
      // pushChanges
    )
```

tpolecat:
Ok yeah `releaseStepCommandAndRemaining("+publishLocalSigned")` isn't the exact equivalent of `publishArtifacts`, which I think probably won't work. I may end up with `      releaseStepCommandAndRemaining("+publishSigned")`. 

I really have no idea what the difference is between `releaseStepCommand` and `releaseStepCommandAndRemaining`, do you?

ChristopherDavenport:
I believe `releaseStepCommmand` executes a single build step. like `test` on 1 version, whereas `releaseStepCommandAndRemaining` will execute the entire list of build steps that a command may expand to.

tpolecat:
Yeah that's the impression I got from the sbt-release source but it still makes no sense to me.

Issue #857, opened 2019-04-11
rawTrans/rawTransP don't close the connection


nigredo-tori:
After tracking down a connection leak in our project, we've found a culprit - a `Transactor.rawTrans` call. The whole purpose of `rawTrans` is to run without the transaction logic (setup/commit/rollback), which is useful in many cases. However, the current `rawTrans` implementation also skips `Strategy.always`. This means that the connection is not properly closed - which in our case resulted in it not being returned to the pool!

The root of the issue here is that `Connection.close` is not a part of the transaction logic, but rather a part of the connection lifetime logic. As such, it should be a part of the `Transactor` itself, and not of its `Strategy`. Ideally, I'd like for `Transaction.connect` to become an `A => Resource[M, Connection]`, so that there is less space to mess up connection handling in the future.

/cc @wedens

wedens:
I'd expand this issue to converting all resource handling (`ResultSet`, `PreparedStatement`, `Connection`, etc) to `Resource` which'll make leaks less likely to occur.

tpolecat:
Yeah if we're allocating the connection we should also close it.

nigredo-tori:
Closing this since `Transactor` and `Strategy` now use `Resource` (#858). @wedens, "resourcifying" the algebras should be its own issue, IMO.

Issue #856, opened 2019-04-09
Fix compatibility with fs2 1.0.4


nigredo-tori:
[fs2 1.0.4](https://github.com/functional-streams-for-scala/fs2/releases/tag/v1.0.4) relaxes the guarantees for order of finalizers. This, for one, means that in expressions like `a ++ b` we can't rely on `a` being finalized before `b` starts. This breaks our [streaming transaction logic](https://github.com/tpolecat/doobie/blob/v0.7.0-M3/modules/core/src/main/scala/doobie/util/transactor.scala#L227), among other things - reading a PostgreSQL large object now leads to an exception since the `BlobInputStream` is closed after the transaction is commited.

fs2 release mentions manually calling `.scope` on streams to provide stronger finalization guarantees - that might be a start.

tpolecat:
Let's fix this for 0.7

PR #858, opened 2019-04-13
Transactor and catchsql changes

### LINKED ISSUE #855: catchsql: except* functions break stream chunking
Some functions in `doobie.util.catchsql` can be reasonably implemented using `.handleErrorWith`/`.recoverWith`, but [use `.attempt` instead](https://github.com/tpolecat/doobie/blob/dd5d681fc31b59fa3352ce2db0c890eb6a4b5e99/modules/core/src/main/scala/doobie/util/catchsql.scala#L33-L34). This makes them unsuitable for use with `fs2.io.Stream`, since `.attempt` breaks chunking:

```scala
scala> def getChunks[A](ma: Stream[SyncIO, A]) = ma.chunks.compile.toVector.unsafeRunSync
getChunks: [A](ma: fs2.Stream[cats.effect.SyncIO,A])Vector[fs2.Chunk[A]]

scala> val ma = Stream.emits(1 to 10).covary[SyncIO]
ma: fs2.Stream[cats.effect.SyncIO,Int] = Stream(..)

scala> getChunks(ma)
res19: Vector[fs2.Chunk[Int]] = Vector(Chunk(1, 2, 3, 4, 5, 6, 7, 8, 9, 10))

scala> getChunks(exceptSql(ma)(_ => Stream.empty))
res20: Vector[fs2.Chunk[Int]] = Vector(Chunk(1), Chunk(2), Chunk(3), Chunk(4), Chunk(5), Chunk(6), Chunk(7), Chunk(8), Chunk(9), Chunk(10))

scala> getChunks(ma.handleErrorWith(_ => Stream.empty))
res21: Vector[fs2.Chunk[Int]] = Vector(Chunk(1, 2, 3, 4, 5, 6, 7, 8, 9, 10))
```

This hurts a lot when using the blob streaming in PostgreSQL module.

I think the implementations of these functions should be changed to use more direct/performant approach.

guymers:
Update scalatest and zio to versions that have 2.11, 2.12 and 2.13.0-M5 released versions.

`Transactor.connect` now takes an `A` and returns a `Connection` wrapped in a `Resource`. The `Resource` is expected to close the connection when it completes. A `Strategy` is now also defined as a `Resource` and ensures that the order of execution is: `before` -> program -> `after`/`oops` -> `always`

Update all methods in `catchsql` to only require an instance of  `ApplicativeError`. The `guarantee` and `onError` methods have been removed as they are provided via cats. Due to the relaxation the `monaderror` syntax has been renamed to `applicativeerror`.

These commits should resolve #855, #856, #857


tpolecat:
This looks great, thanks! I'll give it a closer read later this weekend.

Issue #861, opened 2019-04-24
Example NOTIFY/LISTEN program blocks execution


m4dc4p:
The example postgres listen/notify blocks the current thread of execution by sleeping while polling for notifications. This is by design, of course. However, I used the code w/o really understanding what it was doing which led to some nasty concurrency issues. 

I think it would make sense to update the example so polling happens on a dedicated thread pool.

tpolecat:
Yeah it's an old example written before `Timer` existed. I'll take a PR to fix it.

PR #862, opened 2019-04-24
Update example to use a thread pool for polling.

m4dc4p:
Updates the example to use a thread pool for getting notifications, so anyone copying the code doesn't give themselves a concurrency problem.

For issue #861.

### DIFF
--- a/modules/example/src/main/scala/example/PostgresNotify.scala
+++ b/modules/example/src/main/scala/example/PostgresNotify.scala
@@ -13,6 +13,8 @@ import doobie.postgres._
 import org.postgresql._
 import fs2.Stream
 import fs2.Stream._
+import java.util.concurrent._
+import scala.concurrent._
 
 /**
   * Example exposing PostrgreSQL NOTIFY as a Process[ConnectionIO, PGNotification]. This will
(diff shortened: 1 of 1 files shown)


tpolecat:
I was thinking this would use `Timer.sleep`, lifted into `ConnectionIO`. Something like

```scala
Timer[IO].sleep(1.second).to[ConnectionIO]
```

This way there is no blocking.

m4dc4p:
Thats a lot simpler. Does `Timer` use its own thread pool behind the scenes?

tpolecat:
> Does `Timer` use its own thread pool behind the scenes?

No, it's nonblocking so it doesn't need to.

m4dc4p:
Is `Timer...to[ConnectionIO]` a thing?

tpolecat:
> Is `Timer...to[ConnectionIO]` a thing?

Yes, via `LiftIO`. You need `cats.effect.implicits._` for that syntax.

m4dc4p:
I can't quite get this working.

tpolecat:
Ok no worries, thanks for giving it a shot. I'll see if I can figure it out.

PR #879, opened 2019-04-26
Fix late connection closing on streaming requests

LMnet:
In the recent [1.0.4 update of the fs2](https://github.com/functional-streams-for-scala/fs2/releases/tag/v1.0.4) resource handling is [slightly changed](https://github.com/functional-streams-for-scala/fs2/pull/1417). In the case of doobie this change led to a problem: connection was not returning to a connection pool. Eventually, this led to full connection leakage.

In this pull request, I added `.scope` in the stream natural transformations. With this fix, connections will be released earlier and will not leak.

FYI @mpilquist


vpavkin:
I can confirm the problem, this fixes a problem in one of our concurrent tests as well.
A release with a fix would be much appreciated

tpolecat:
Is there an obvious way to write an accompanying test?

LMnet:
I don't know a proper way to write a unit test for this case. 

But I can create an integration test with the real db and hikari pool. In my case, I faced with this issue when I repeated a stream queries one after one, and when the number of queries became bigger than the pool size application stuck, because all connections were busy. 

I don't know where should I place integration test: the fix itself is in the core module, but it doesn't know about hikari pool.

vpavkin:
I think it's ok to have a test-containers based test in doobie-postgres module, is it?
I can give it a shot if there're no objections :)

tpolecat:
The CI build initializes a database that's used for testing. You can make hikari depend on the postgres module in the test configuration and add a test in the hikari module. If this is too much let me know and I can do it, I just need you to explain how to make the leak happen.

vpavkin:
@tpolecat I made a separate PR with a test case #887

PR #887, opened 2019-04-28
Reproduction test for leaking connections in 0.7.0-M4

vpavkin:
Managed to write quite a simple test case for #879.
Fails in 0.7.0-M4, but if you apply changes from #879, it becomes green.

/cc @LMnet


tpolecat:
Cool, thanks. Can you cherry-pick that commit into this PR or vice-versa so we can get the fix and test in a single PR?

LMnet:
@tpolecat I merged this test into my branch. This pull request could be closed.

PR #893, opened 2019-05-06
updates

tpolecat:
this consolidates all the scala-steward updates


tpolecat:
cancelled the CI build to unblock the org queue, will restart in a bit

PR #894, opened 2019-05-07
2.13.0 Support

ChristopherDavenport:
2.13.0-M5 Support In Place

Currently having an issue with the PostGIS build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,10 @@ jobs:
       before_install:
         - export PATH=${PATH}:./vendor/bundle
       addons:
-      postgresql: "9.4"
-      apt:
-        packages:
-          - postgresql-9.4-postgis-2.3 # This might not be necessary in the future. If the Travis people update the trusty env to include this package whenever it sees pg getting installed then this can go away.
+        postgresql: "9.4"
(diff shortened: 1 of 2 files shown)


tpolecat:
Wow. I'll have a look in the morning, thanks!

tpolecat:
Wow I don't understand that Travis build at all, but it looks great! Are there any issues as far as you know?

ChristopherDavenport:
Your biggest gotcha would be if you run `sbt ++2.13.0-M5 test` or similar rather than `sbt ++2.13.0-M5 thirteen/test` you will run into issues with docs and quill and the missing dependencies you saw when you were fighting it earlier. But we move around it entirely with the project, and let the crossVersioning on each of the sub-projects control the release process with there being no cross-versions for any of the top level projects. 

So in general I think this is good to go. But it will be very nice to remove when all your dependencies have releases.

ChristopherDavenport:
Also I had to force trusty, because the postGIS issue was due to something changing with the Xenial gis process. So that means when you change that base image there will be some changes needed to your before_travis script.

tpolecat:
Sounds great, thanks.

PR #899, opened 2019-05-08
modernize NOTIFY example

### LINKED ISSUE #861: Example NOTIFY/LISTEN program blocks execution

tpolecat:
resolves #861 

/cc @m4dc4p

### DIFF
--- a/modules/example/src/main/scala/example/PostgresNotify.scala
+++ b/modules/example/src/main/scala/example/PostgresNotify.scala
@@ -5,7 +5,8 @@
 // relies on streaming, so no cats for now
 package example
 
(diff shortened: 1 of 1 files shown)


jbaileyashe:
Awesome! Can't wait to port this into my code.

m4dc4p:
Fantastic! Thank you.

PR #903, opened 2019-05-22
Make `and`/`or` & co. composable and add `comma` and `orderBy` `fragments` helpers

sideeffffect:


tpolecat:
Thanks! It looks like you need to suppress the overloading warning. Search for `@suppresswarning` in the codebase and you'll find some examples.

```
[error] /home/travis/build/tpolecat/doobie/modules/core/src/main/scala/doobie/util/fragments.scala:74:7: [wartremover:Overloading] Overloading is disabled
[error]   def combineWith[A](f: (Fragment*) => Fragment)(
[error]       ^
[error] /home/travis/build/tpolecat/doobie/modules/core/src/main/scala/doobie/util/fragments.scala:78:7: [wartremover:Overloading] Overloading is disabled
[error]   def combineWith[A](f: (Option[Fragment]*) => Fragment)(
[error]       ^
[error] two errors found
```

sideeffffect:
why is the build still yellow? it shows as green on Travis: https://travis-ci.org/tpolecat/doobie/builds/535950468

modules/core/src/main/scala/doobie/util/fragments.scala
```diff
@@ -20,41 +20,41 @@ object fragments {
   def notIn[F[_]: Reducible, A: Put](f: Fragment, fs: F[A]): Fragment =
     fs.toList.map(a => fr0"$a").foldSmash1(f ++ fr0"NOT IN (", fr",", fr")")
 
-  /** Returns `(f1) AND (f2) AND ... (fn)`. */
-  def and(fs: Fragment*): Fragment =
-    fs.toList.map(parentheses).intercalate(fr"AND")
+  /** Returns `(f1 AND f2 AND ... fn)`, or nothing for empty `fs`. */
+  def and(fs: Fragment*): Option[Fragment] =
+    if (fs.isEmpty) None else Some(parentheses(fs.toList.intercalate(fr"AND")))
```
tpolecat: We want each fragment parenthesized here; i..e, `(a) AND (b) AND (c)` so things like `(a OR b) AND (c or d)` don't turn into `a OR b AND c OR d` which will be parsed as ((a OR b) AND c) or d`.

I wonder if we want to return `true` when `fs` is empty, and `false` for `or` ... ?

sideeffffect: > each fragment parenthesized here; i..e, `(a) AND (b) AND (c)`

that would be tackling the issue from the wrong end. `and` should itself return a "single" expression, ie something parenthesized
so that `a OR b AND c OR d` is not possible to construct. `and(or(a, b), or(c, d))` will always create `((a OR b) AND (c or d))`
see the newly added tests to see how this makes these functions (`and`, `or`,...) composable and how the current approach would fail here.

tpolecat: Ok, that makes sense. I will look at the tests. I'm a little concerned that this will quietly break existing code so we should add a release note about it.

sideeffffect: thanks! this (we could even consider it a bug) has bitten me. I tried to come up with a solution which wouldn't be as disruptive, but couldn't.
please ask any question you might have about my proposed fix

sideeffffect:
I tried to describe the change in `CHANGELOG`

/cc @augi

CHANGELOG.md
```diff
@@ -29,7 +29,7 @@ And an assortment of other improvements:
 - `Text[Char]` instance was added for Postgres `COPY FROM STDIN`. Thanks **Dermot Haughey**!
 - You can now create a `HikariTransactor` from a `HikariConfig`. Thanks **Yuriy Badalyantc**!
 - Thanks to **Harry Laoulakos** and **Daan Hoogenboezem** for documentation updates!
-- `Fragment` combinators now parenthesize arguments so associativity in Scala is reflected in generated SQL. Thanks **Katrix**!
+- `Fragment` combinators now return parenthesized expressions so associativity in Scala is reflected in generated SQL. `and` and `or` & co. return an `Option[Fragment]`, `None` in case the input list is empty. This ensures that a valid SQL is always produced, i.e. no `... WHERE true AND ( ) `. Thanks **Katrix**!
```
tpolecat: What's the argument against returning `true` and `false` for empty `and` and `or`, respectively?

sideeffffect: 1. I'm not sure how well is the emptiness established (or could be) in the first place (let alone equality)
2. even if we had a reliable `isEmpty` method, it doesn't help me very well with composing/nesting `and`s or `or`s -- we actually, `andOpt`s and `orOpt`s to be precise
3 let's take a specific example:

```
val list1 = List()
val list2 = List(fr1, fr2)
...
andOpt(or(list1: _*), or(list2: _*))
```
this will generate exactly what I want
```
((fr1 OR fr2))
```
so not `() AND ((fr1) OR (fr2))` which is not a valid SQL

tpolecat: Emptiness is well-defined for sequences and equality isn't relevant. Maybe I misunderstand.
 
My suggestion is that `and()` return `true` and `or()` return false, consistent with the boolean monoids, thus your `and(or(), or(a, b))` would return `(false and (a or b))`.

sideeffffect: > Emptiness is well-defined for sequences

I was talking about emptiness/equality of `Fragment`s, but only because I misunderstood you original suggestion _"returning true and false for empty and..."_, sorry :)
I now get what you mean!

> `and(or(), or(a, b))` would return `(false and (a or b))`
which is just `false`, btw.

as a user of these helper functions, I would consider that quite a surprising and undesirable behaviour -- opposite of what I want to achieve.
wouldn't also you as a newcomer be surprised by this?

I intuitively expect `and(or(), or(a, b))` to be essentially equivalent to `or(a, b)`

---

to use the language of monoid, the empty `or` in an `and` shouldn't be replaced by a neutral element for `or`, but by a neutral element of the enclosing function. which is in our case `and` and so the empty `or()` should be replaced by `true`

`X(Y(a, b), Y(c, d, e), Y())` `=>` `X(Y(a, b), Y(c, d, e), neutralX)`

Katrix:
If the combinators will now return Option, can we add an overload that takes a NonEmptyList to avoid long chains of map, flatmap and so on?

sideeffffect:
@Katrix good point, added

sideeffffect:
@tpolecat unfortunately the PR is red, because my changes are not backwards compatible and a new series (`0.7.x`) has been released in the meantime :(
can you please create the `series/0.8.x` branch?

sideeffffect:
Hello @tpolecat I would like to bring this PR up to date. Are you still open to improving the `fragments` helper functions in the spirit of this PR?

sideeffffect:
@tpolecat gentle ping :)

tpolecat:
> gentle ping :)

noted, you're on the list!

tpolecat:
This can't be merged into 0.8 so you need to cherry-pick that commit over to a new branch and re-open.

PR #915, opened 2019-06-14
Update to Scala 2.13.0

guymers:
Use cats `show` to avoid wartremover `StringPlusAny` errors on Scala 2.13

Wondering if this PR should wait until there are released versions of cats, circe and fs2.


tpolecat:
Cool, thanks! Something we could do is put this on a 0.8.x branch and release it as 0.8.0-M1 … that way people can mess with it and we can do 0.8.0 final when the libraries are all caught up. What do you think?

guymers:
Sounds good, if you can create that branch I will update this PR to point at it.

PR #920, opened 2019-06-19
Add EXPLAIN and EXPLAIN ANALYZE output to Query.

m4dc4p:


m4dc4p:
"[error] java.lang.RuntimeException: doobie-core: Binary compatibility check failed!"

Do I need to bump a version number somewhere?

m4dc4p:
Wrong branch.

tpolecat:
Hi, thanks for this. Unfortunately `EXPLAIN` is vendor-specific so we can't really add it to `Query/Query0` directly, but you could get there via a generic method that lets you inspect the SQL and the program that sets the statement parameters. Something like this maybe? (off the top of my head, may not even typecheck)

```scala

// on Query
def inspect[B](a: A)(f: (String, PreparedStatementIO[Unit]) => ConnectionIO[B]): ConnectionIO[B] =
  f(sql, HPS.set(a))

// Somewhere in the postgres module
implicit class PostgresQueryOps[A](self: Query[A]) {

  def explain(a: A): ConnectionIO[List[String]] =
    self.inspect { (sql, prepare) =>
      HC.prepareStatement(s"EXPLAIN $sql")(prepare *> HPS.executeQuery(a, HRS.build[List, String]))
    }

}
```

m4dc4p:
Thanks for the guidance! I added `explainAnalyze` and `explain` to `Query0` and `Query` when `postgres.implicits._` is in scope.

I added some info to the documentation, but all I was able to do was run `docs/tut` - I'm not sure if the HTML generated looks good or not ...

Please check out the updated implementation.

tpolecat:
Can you rebase this onto 0.8.x?

m4dc4p:
> 
> 
> Can you rebase this onto 0.8.x?

Done! See #942 (branch names got too confusing here.)

PR #921, opened 2019-06-19
Add EXPLAIN and EXPLAIN ANALYZE output to Query.

m4dc4p:


m4dc4p:
This does break binary compat with `0.6.0` (by adding methods to `Query` and `Query0`. Is that acceptable or ?

Not sure if how to do with an extension method or implicit but I suspect there is a way.  May involve cloning the private `executeQuery` method from `Query` tho.

m4dc4p:
Closed in favor of https://github.com/tpolecat/doobie/pull/920 (would rather target 0.7.x)

PR #957, opened 2019-07-22
Update to Quill 3.3.0

deusaquilus:
Okay... let's see if this works. In Quill 3.3.0 we're just changing the SQL `RETURNING` API to allow passing multiple columns etc... so `column:String` in `JdbcContextBase` is replaced with `ReturningAction` which can be:
1. `ReturnColumns` which contains a list of column strings but assumes that the underlying SQL query dialect does not have a `RETURNING` clause so the columns need to be specified in the JDBC.
2. `ReturningRecord` which assumes the underlying Dialect *does* have a `RETURNING` syntax so the columns do not have to be specified in the JDBC (but `Statement.RETURN_GENERATED_KEYS` does need to be set in `prepareStatement`).
3. `ReturnNothing` this is a rare case for SQL Dialects that don't really support any Returning semantics at all but somehow were called with the `.returning` API (so `Statement.RETURN_GENERATED_KEYS` should not be set). I don't really think this case needs to be considered.


tpolecat:
Cool, thanks for getting the ball rolling on this. I'll have a look.

What's the status of 2.13 support in Quill? Right now we only compile this module for 2.11/12.

deusaquilus:
We've got a PR for 2.13 but there's a bit of a hang-up because Spark 2.13 is not out yet. Probably the solution is to disable that part of the build for 2.13 but I haven't managed to get to it yet.

tpolecat:
I didn't know there was a Spark module. Kind of guessing that one is going to drag for a while.

deusaquilus:
We'll figure it out.

deusaquilus:
There's something I think I'm doing wrong here. Do columns have to specified even when there is a `RETURNING` clause directly in the SQL? I.e:
```scala
HC.prepareStatement[A](sql)(_) // I do this when there is a RETURNING clause
HC.prepareStatementS[A](sql, columns)(_) // I do this when there is no RETURNING clause (e.g. for databases that don't support it) but I want to get columns returned anyway
```
This is based on the behavior I saw with JDBC. Does doobie do it differently?

tpolecat:
If you have a RETURNING clause in your statement you can executeQuery and get a ResultSet directly. Otherwise you have to executeUpdate and then getUniqueKeys. So it may be the followup after prepare that's the issue ... ? Just guessing.

deusaquilus:
I'm pretty sure you can use `getUniqueKeys` even in the former case. This is what I do with jdbc:
```scala
protected def prepareWithReturning(sql: String, conn: Connection, returningBehavior: ReturnAction) =
    returningBehavior match {
      // I.e. SQL has a RETURNING clause
      case ReturnRecord           => conn.prepareStatement(sql, Statement.RETURN_GENERATED_KEYS)
      // SQL Does not have a returning clause
      case ReturnColumns(columns) => conn.prepareStatement(sql, columns.toArray)
      // One-off
      case ReturnNothing          => conn.prepareStatement(sql)
    }
```

deusaquilus:
Okay. Let's try it with `executeQuery`. If it works...

deusaquilus:
Okay. I tried `executeQuery` and it's still not working. Any ideas?
```
0 != 1
Expected :1
Actual   :0
 <Click to see difference>

java.lang.Exception: 0 != 1
	at org.specs2.matcher.MatchResultStackTrace.setStacktrace(Expectations.scala:57)
	at org.specs2.matcher.MatchResultStackTrace.setStacktrace$(Expectations.scala:55)
	at org.specs2.mutable.Specification.setStacktrace(Specification.scala:15)
	at org.specs2.matcher.ExpectationsCreation.checkFailure(Expectations.scala:37)
	at org.specs2.matcher.ExpectationsCreation.checkFailure$(Expectations.scala:36)
	at org.specs2.mutable.Specification.checkFailure(Specification.scala:15)
	at org.specs2.matcher.ShouldThrownExpectables$$anon$4.check(ShouldExpectations.scala:57)
	at org.specs2.matcher.Expectable.applyMatcher(Expectable.scala:51)
	at org.specs2.matcher.ShouldThrownExpectables$$anon$4.applyMatcher(ShouldExpectations.scala:56)
	at org.specs2.matcher.ShouldExpectable.should_$eq$eq(ShouldExpectable.scala:19)
	at doobie.postgres.PostgresDoobieContextSpec$.$anonfun$new$27(PostgresDoobieContextSpec.scala:101)
```

tpolecat:
Well at least it compiled. Sorry, I'm kind of swamped and won't be able to look at it until next week. Push your WIP and I'll have a look when I can.

deusaquilus:
Already pushed :)

deusaquilus:
I tried using both `getGeneratedKeys` and `executeQuery` which I am doing now. The `getGeneratedKeys` version looked like this:

```scala
  override def executeActionReturning[A](
    sql:               String,
    prepare:           Prepare = identityPrepare,
    extractor:         Extractor[A],
    returningBehavior: ReturnAction
  ): ConnectionIO[A] = {
    prepareConnections[A](returningBehavior)(sql) {
      FPS.raw(prepare)  *>
        FPS.executeUpdate *>
        HPS.getGeneratedKeys(HRS.getUnique(extractor))
    }
  }a

  private def prepareConnections[A](returningBehavior: ReturnAction) =
    returningBehavior match {
      case ReturnColumns(columns) =>
        (sql:String) => HC.prepareStatementS[A](sql, columns)(_)
      case _ =>
        (sql:String) => HC.prepareStatement[A](sql)(_)
    }

  override def executeBatchActionReturning[A](
    groups:    List[BatchGroupReturning],
    extractor: Extractor[A]
  ): ConnectionIO[List[A]] = {
    groups.flatTraverse { case BatchGroupReturning(sql, returningBehavior, preps) =>
      prepareConnections(returningBehavior)(sql) {
        preps.traverse(FPS.raw(_) *> FPS.addBatch) *>
        HPS.executeBatch *>
        HPS.getGeneratedKeys(HRS.list(extractor))
      }
    }
  }
```

I had exactly the same test failures with the code above as I am having now.

deusaquilus:
It works. Yay! I'm going back to initial implementation using `executeUpdate`/`executeBatch` + `getGeneratedKeys` instead of `executeQuery` since there's no batch equivalent to the latter that I know of (although when I implement `ReturnRecord` for SQL we'll have to figure that out). This is good to go.

tpolecat:
Brilliant, thanks!

PR #978, opened 2019-08-16
Added syntax for converting to a fragment

Billzabob:
Not sure if I need to put that Copyright stuff at the top or if it would be autogenerated anyways.

Also this is my first "big" PR so be gentle :)


Billzabob:
I'm curious if there's a reason for using implicit defs instead of implicit classes for all the syntax stuff. I used implicit defs here to be consistent.

PR #979, opened 2019-08-16
Update cross-transactor stream example to allow different schemas

dantb:
I needed to do a streamed DB migration (very helpful example, thanks!) but the schemas were different in my case. Ended up adapting this example with some processing in the middle - I thought it might be useful to someone else someday. Please think of a better name though!


tpolecat:
Thanks!

PR #1001, opened 2019-09-11
update circe, fs2; remove 2.11 support

tpolecat:
Like it says on the tin. Circe isn't available for 2.11 anymore so I'm dropping support here as well.

Note that Postgres JDBC driver 42.2.7 is out but I'm skipping it because the newly-added mapping from `inet` ⟷ `InetAddress` is buggy and the existing mapping via `PGObject` no longer works. There's a fix planned for 42.2.8 so I'll wait for that and try again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ dist: xenial
 stages:
   - name: test
 
-scala_version_211: &scala_version_211 "2.11.12"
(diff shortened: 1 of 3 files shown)


tpolecat:
Thanks @ChristopherDavenport looks like that worked!

PR #1004, opened 2019-09-13
remove sbt-release, add sbt-ci-release

tpolecat:
eh this messes with the build a lot, let's see if it still builds


tpolecat:
ok this seems right. it won't release until we create a `master` branch

PR #1044, opened 2019-10-09
Removed instructions for partial-unification in docs, as it is now included by default

tuckerbuchy:
Had a minor issue with the build.sbt described in the docs, as `partial-unification` seems to be included by default now 

https://github.com/typelevel/cats/issues/2948


tpolecat:
It's on by default in 2.13 but is still required in earlier versions. We should probably clarify. Also 2.11 support has been dropped.

tuckerbuchy:
yeah true. I've rephrased it to be a bit clearer. Hopefully this helps

jatcwang:
@tuckerbuchy Can you rebase on top of latest master? Then we can merge this :)

PR #1045, opened 2019-10-11
 Fix #844, #610, #778, #652 add interpolation of Fragments

### LINKED ISSUE #610: Composition of Fragments by interpolation.
Is it possible to implement interpolation of fragments? I find that use fragments quite a lot, for example when defining which columns a resource I'm interested in has. Or when using `IN` queries, etc.

An example of how I'm currently composing fragments
```scala
import cats._
import cats.data._
import cats.implicits._
import doobie.imports._
import fs2.interop.cats._

case class Car(carId: Long, color: String, manufacturer: String, year: Int)

object CarQueries{
  def columns: Fragment = fr"cars.car_id, cars.color, cars.manufacturer, cars.year"
  def blueCarsByIdsQuery(carIds: NonEmptyList[Long]): Query0[Car] = (
    fr"SELECT" ++ CarQueries.columns ++
    fr"""FROM cars""" ++ 
    Fragments.whereAnd(Fragments.in(fr"car_id", carIds), fr"color = 'blue'"))
   ).query[Car]
}
```

What I'd love to do is write 
```scala
  def blueCarsByIdsQuery(carIds: NonEmptyList[Long]): Query0[Car] =
    fr"""SELECT ${CarQueries.columns}
          FROM cars
          ${Fragments.whereAnd(Fragments.in(fr"cars.car_id", carIds), fr"color = 'blue'")}""".query[Car]
```
Which in my opinion is quite a lot cleaner.

I took a stab at a sketch here: https://github.com/tpolecat/doobie/compare/series/0.5.x...halvorgb:fragment-interpolation-sketch

I'm convinced that my sketch is terrible, I don't need any help to convince me of that,  but how hard would it be to implement something like this? How would you send the parameters in the correct order to JDBC for "nested" fragments?

### LINKED ISSUE #652: Add string interpolation like slick's #$stringInterpolate
Example usage:
```
val table = "foo"
val a = sql"select 1 from #$table"
```

### LINKED ISSUE #778: Literal value splicing in fragments?
Currently, to splice in literal values to `Fragment`s we use `Fragment.const` and concatenation as recommended [in the guide](https://tpolecat.github.io/doobie/docs/08-Fragments.html#composing-sql-literals), which is a little less smooth than say, [what Slick offers](http://slick.lightbend.com/doc/3.2.3/sql.html#splicing-literal-values) in its plain SQL experience.

e.g, currently:
```scala
import doobie._
import doobie.implicits._
import cats.implicits._

val myTable = "my_table"
val fragment = fr"CREATE TABLE" ++ Fragment.const(myTable)
```

Maybe taking the literal syntax from Slick (would make it easier to port plain SQL usage from Slick if someone wanted to):
```scala
val myTable = "my_table"
val fragment = fr"CREATE TABLE #$myTable"
```

Perhaps it would be nice to have similar syntax in doobie, probably with the requisite warnings about injection...

3ba

### LINKED ISSUE #844: Better SQL literals syntax
### Motivation
Slick: `sqlu"CREATE TABLE #$tableName (id integer PRIMARY KEY)"`
doobie: `(sql"CREATE TABLE" ++ Fragment.const(tableName) ++ fr"(id integer PRIMARY KEY)").update.run`

#### Slick-like
Use same syntax as Slick uses, so probably check [here](https://github.com/tpolecat/doobie/blob/series/0.7.x/modules/core/src/main/scala/doobie/syntax/string.scala#L20) for parts ending with `#`, and then don't create the parameter and append the raw value directly to the sql string.
#### Fragments injection
Allow to inject Fragments using interpolation, so final syntax would be: `sql"CREATE TABLE ${Fragment(tableName)} (id integer PRIMARY KEY)"`
This would require a lot of work IMHO, because Fragment can contains parameters...
#### Raw type
Introduce a new `Raw` type and handle it somehow in the `SqlInterpolator` class (simplified version of previous solution): `sql"CREATE TABLE ${Raw(tableName)} (id integer PRIMARY KEY)"`

Unfortunately, I'm not able to implement none of proposed solutions because I don't understand what't actually happening in the `SqlInterpolator` class :(

neko-kai:
This bases on [Nuke Param PR](https://github.com/tpolecat/doobie/pull/1035), but can be done with `Param` too if Param can't go... Please tell me what you think!

[interpolation commit](https://github.com/tpolecat/doobie/pull/1045/commits/20e213adb1c5a11b0787d701c149fdb4d6b1e78f)

Fix #844, Fix #610, Fix #778, Fix #652 add interpolation of Fragments


neko-kai:
@tpolecat 🙏

neko-kai:
@tpolecat is there anything i can do for this PR, or any PR to add fragment interpolations or similar to be merged?

tpolecat:
Hi, sorry, if you rebase this I can get have a look. I think it's good. Please add "resolves ..." text to the description so we close everything on merge.

neko-kai:
@tpolecat 
Rebased. I've changed both commit messages & description to add `Fix` text for all issues (github seems to highlight it, same as "resolves", and it worked for me before)

tpolecat:
Thanks!

neko-kai:
Thanks!

PR #1057, opened 2019-10-24
Produce a 2.13 artifact for doobie-quill

crispywalrus:
this includes the scala-steward PR to upgrade quill to 3.4.10 which includes 2.13 artifacts and then removes the block on the doobie-quill_2.13 artifact.


tpolecat:
I think the Travis build requires more changes. There's some janky stuff in there.

crispywalrus:
sass failed to install. that seems totally unrelated but I can't find a similar failure.

crispywalrus:
After a fair bit of testing I'm pretty confident that the failure in the build was due to a (likely) one off. My forked repo builds fine and now with a whitespace change, because I don't have the ability to rerun the build by hand, it builds fine in this repo as well. I took me longer than it should have but this seems ready to merge now.

tpolecat:
The quill artifact is published as of 0.8.6. Forgot to close this. Sorry!

Issue #1056, opened 2019-10-24
[Feature request] streaming API for PostgreSQL's COPY


Z1kkurat:
Currently Doobie supports PostgreSQL's COPY for any `F: Foldable` which severely limits its usage, since COPY is perfect for inserting large datasets, and large dataset, embedded in `Foldable` would probably overflow application memory.
It would be nice to have Doobie API that can work with `Stream[F, A]` or `Stream[ConnectionIO, A]` to load large amounts of data into Postgres tables.

Z1kkurat:
Current JDBC API for this is [here](https://jdbc.postgresql.org/documentation/publicapi/org/postgresql/copy/CopyManager.html#copyIn-java.lang.String-java.io.InputStream-).
I guess our only choice is turning FS2's Stream into Java's InputStream by some means to be consumer by this API.

tpolecat:
This is the method in question

https://github.com/tpolecat/doobie/blob/master/modules/postgres/src/main/scala/doobie/postgres/syntax/FragmentSyntax.scala#L16-L32

It needs to take a stream instead of a foldable, but it's unclear what the effect type of the stream needs to be. It may need to be `G[_]: ConcurrentEffect` or something like that, depending on what the Java iostream stuff requires.

Daenyth:
It *seems* like this should be able to be implemented using `fs2.io.toInputStreamResource` which indeed requires `ConcurrentEffect`

PR #1058, opened 2019-10-25
Implement Postgres COPY stream support

Z1kkurat:
This resolves https://github.com/tpolecat/doobie/issues/1056.
Points to discuss:

- should old function be left for compatibility/etc.
- should `minChunkSize` param have some default? (it breaks type inference if it has, thus it is connected to n.1)
- is it possible to detect empty stream and not fail, returning `0` (as it is in the `Foldable` implementation)

I have run JMH benchmarks on my local machine: i7 6700k/16gb and got the following results
```
[info] Benchmark                    Mode  Cnt       Score       Error  Units
[info] d.b.t.fast_copyin_foldable  thrpt   20  384070.994 ± 44968.142  ops/s
[info] d.b.t.fast_copyin_stream    thrpt   20  377594.270 ± 40039.823  ops/s
[info] d.b.t.jdbc_copyin           thrpt   20  125470.964 ±  2952.709  ops/s
[info] d.b.t.naive_copyin          thrpt   20   41790.710 ±  2020.370  ops/s
```
Judging from these results, stream version (with `minChunkSize` = 10000) is quite as performant as `Foldable` version, but allowing large amounts of data to be copied.


Z1kkurat:
@tpolecat Hi, Rob! Is there anything I can improve on this PR? Would appreciate any feedback 🙏

tpolecat:
Having a look now, sorry for the delay.

tpolecat:
I made a few changes to avoid `Resource.allocate`. Continued at https://github.com/tpolecat/doobie/pull/1123

PR #1076, opened 2019-11-25
tut: fix psql command

giuscri:
This commit fix a SQL command that's going to fail for most users.

If not specified the command will use $USER as the db default. Yet in
general the only present db is `postgres`.


tpolecat:
I did not know that, thanks!

PR #1090, opened 2019-12-13
Add new Meta instances and tests for JSR-310 data types

### LINKED ISSUE #1070: new date issue in latest h2
dates are getting parsed in a timezone, looks like

https://github.com/tpolecat/doobie/pull/1069/files#diff-b430b84e498e097bc391705fedeb983d

zsambek:
This PR resolves #1070

### DIFF
--- a/modules/core/src/main/scala/doobie/enum/jdbctype.scala
+++ b/modules/core/src/main/scala/doobie/enum/jdbctype.scala
@@ -49,6 +49,7 @@ object JdbcType {
   /** @group Values */ case object SqlXml                 extends JdbcType(SQLXML)
   /** @group Values */ case object Struct                 extends JdbcType(STRUCT)
   /** @group Values */ case object Time                   extends JdbcType(TIME)
+  /** @group Values */ case object TimeWithTimezone       extends JdbcType(TIME_WITH_TIMEZONE)
   /** @group Values */ case object Timestamp              extends JdbcType(TIMESTAMP)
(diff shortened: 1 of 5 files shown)


jozic:
would be nice to update docs reflecting those changes

https://github.com/tpolecat/doobie/blob/master/modules/docs/src/main/tut/docs/12-Custom-Mappings.md

zsambek:
@jozic Good idea! Will do later today.

zsambek:
@jozic Updated the docs.

tpolecat:
Tagged as v0.8.7 … should be on Sonatype in a few hours.

zsambek:
Thanks!

jozic:
I have a `mysql` db with a table containing field `start` with `datetime` type.
I use `java.time.Instant` on the code side, and with doobie 0.8.6 all was good, but after upgrading to 0.8.7 my tests inserting data started to fail with 
```
com.mysql.cj.jdbc.exceptions.MysqlDataTruncation, with message: 
Data truncation: Incorrect datetime value: 
'\xAC\xED\x00\x05sr\x00\x0Djava.time.Ser\x95]\x84\xBA\x1B"H\xB2\x0C\x00\x00xpw\x0D\x02\x00\x00\x00\x00^\x1B\x0A@\x00\x00\x00\x00x' for column 'start' at row 1.
```

not sure if I use wrong combination of mysql/java types, but it used to work

zsambek:
@jozic I think it is the same as #1097. Until the fix is released please use 0.8.6

Issue #1095, opened 2019-12-16
PSQLException: Can't infer the SQL type to use for an instance of java.time.Instant.


Fristi:
Since upgrading to 0.8.7 I get

```
Exception: org.postgresql.util.PSQLException: Can't infer the SQL type to use for an instance of java.time.Instant. Use setObject() with an explicit Types value to specify the type to use.
```

mlopes:
I'm getting a similar one which I think might be related, also when upgrading from `0.8.6` to `0.8.7`.

`com.microsoft.sqlserver.jdbc.SQLServerException: The conversion to class java.time.Instant is unsupported.`

zsambek:
@Fristi Since doobie is only a wrapper around the JDBC if the underlying JDBC does not support the given type then, unfortunately, we could not do anything. 

From Pg JDBC Driver docs:
> Note that ZonedDateTime, Instant and OffsetTime / TIME [ WITHOUT TIMEZONE ] are not supported.

`java.time.Instant` only worked previously because it was converted to `java.sql.Timestamp` before passing to the underlying JDBC library. 

What I can advise is to use a data type that is supported by the JDBC driver you are using or implement a custom mapping for `java.time.Instant`. You can learn more about Custom Mappings: [here](https://tpolecat.github.io/doobie/docs/12-Custom-Mappings.html).

mlopes:
@zsambek thanks for the clarification, I guess the confusion came from the BC break on a patch version change.

zsambek:
@mlopes We are providing instances, not for just `java.time.Instant`, but all other JSR-310 data types too. 
Providing instance means that we provide a Meta that knows to which methods should we call on the PreparedStatement and on the ResultSet to serialize and deserialize the type. 
The problem here is that the Pg JDBC does not support `java.time.Instant` serialization and deserialization. You can read more about JSR-310 types which are supported by PostgreSQL JDBC: [here](https://jdbc.postgresql.org/documentation/head/8-date-time.html)

`java.time.Instant` serialization only worked before because doobie did not use the recommended method to serialize and deserialize the type using `_.setObject(col, instant)` and `_.getObject(col, classOf[java.time.Instant])` instead it first converted it to a `java.sql.Timestamp` and serialized as a `java.sql.Timestamp`.

tpolecat:
We may have been too hasty removing the old instances. You can add the instance back if it's working ok for you. `Meta[Timestamp].imap(...`, etc.

[...]

PR #1097, opened 2019-12-16
Revert to use java.sql.Timestamp to serialize java.time.Instant

### LINKED ISSUE #1096: Inserting Instant failed
Since the version 0.8.7 my queries are failing when trying to insert an instant like `sql"insert into test_instant(my_instant) values (${Instant.now()})".update.run`.
It seems that that value is not retrieved:
```
Data truncation: Incorrect datetime value: '\xAC\xED\x00\x05sr\x00\x0Djava.time.Ser\x95]\x84\xBA\x1B"H\xB2\x0C\x00\x00xpw\x0D\x02\x00\x00\x00\x00]\xF7\x9A\xC4;m\x03@x' for column 'my_instant' at row 1
```

zsambek:
Fixes #1096 #1095

### DIFF
--- a/modules/core/src/main/scala/doobie/util/meta.scala
+++ b/modules/core/src/main/scala/doobie/util/meta.scala
@@ -260,10 +260,7 @@ trait MetaInstances { this: MetaConstructors =>
 
   /** @group Instances */
   implicit val JavaTimeInstantMeta: Meta[java.time.Instant] =
-    Basic.one[java.time.Instant](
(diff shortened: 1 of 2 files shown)


max5599:
Should a test be added to avoid this regression?

zsambek:
@max5599 it is already in there. We have two tests, one for Pg and one for H2

zsambek:
@tpolecat May you know why did it fail?

PR #1100, opened 2019-12-17
java.sql._ and java.time._ explicit importing

zsambek:
`import doobie.implicits.javasql._`
`import doobie.implicits.javatime._`
`import doobie.implicits.legacy.instant._`
`import doobie.implicits.legacy.localdate._`


tpolecat:
This looks good. We need to update the doc (or at least release notes) but can do that as a followup.

AiiR42:
@zsambek @tpolecat Hi!

As I see in JDBC 4.2 specification (https://jcp.org/en/jsr/detail?id=221) support for `Instant` and `ZonedDateTime` is not required, so it looks a bit confusing that we have implicits for `Instant` moved to `legacy`.

We can't be sure that driver supports `Instant` or `ZonedDateTime` even if it supports JDBC 4.2 (e.g. latest PostgreSQL driver doesn't support it), so I think we shouldn't have it in `doobie.implicits.javatime`. But I guess we can support these types in a way `doobie.implicits.legacy` works (and place implicits in a separate object without `legacy` in name).

What do you think?

tpolecat:
I'm not quite sure what you're proposing. Can you lay out what would go where? I don't feel strongly about it.

AiiR42:
Something like:

```
doobie.implicits.javatime.jdbc42._ (classes supported by JDBC 4.2)
doobie.implicits.javatime.all._ (jdbc42 plus `Instant` and `ZonedDateTime`)

doobie.implicits.javatime.viasql.instant._ (former doobie.implicits.legacy.instant._)
doobie.implicits.javatime.viasql.localdate._ (former doobie.implicits.legacy.localdate._)
doobie.implicits.javatime.viasql.all._ (support for all java.time.* via conversions with java.sql.*; not sure we need that)
```

That will allow us to use direct Meta (which works via `getObject`) for java.time.* classes supported by JDBC 4.2 and combine it with Instant Meta (which works via conversions with java.sql.*) if we need to.

Second part with "renaming" `legacy` to `viasql` is not needed if you plan to get rid of `doobie.implicits.legacy` completely in future releases (and do not have "working in most cases" Meta for Instant).

PR #1113, opened 2019-12-25
add query toMap function

keiSunagawa:
`to[F]` do can't convert to `Map[K, V]`.
this PR added function `toMap` , that can only call if type `Query0[(K, V)]`


modules/core/src/test/scala/doobie/util/query.scala
```diff
@@ -23,11 +23,15 @@ class queryspec extends Specification {
...
 
   val q = Query[String,Int]("select 123 where ? = 'foo'", None)
+  val pairQuery = Query[String, (String, Int)]("select 'xxx', 123 where ? = 'foo'", None)
 
   "Query (non-empty)" >> {
     "to" in {
       q.to[List]("foo").transact(xa).unsafeRunSync must_=== List(123)
     }
+    "toMap" in {
+      pairQuery.toMap[String, Int]("foo").transact(xa).unsafeRunSync must_=== Map("xxx" -> 123)
```
keiSunagawa: type annotation not needs if only scala 2.13.

tpolecat:
I keep hoping there's some way to do this by generalizing `.to` but this seems fine. Will need to do another merge but you're next in line.

keiSunagawa:
I did merge  succeed, I think.
please check :bow: .

keiSunagawa:
@tpolecat 
Sorry. I may not be able grasp your intention.
> I keep hoping there's some way to do this by generalizing .to but this seems fine. Will need to do another merge but you're next in line.

I think "another merge" for resolved, because I see "All checks have passed" on Github web ui.
but, this PR was not merge in this 1 week.
If I'm misunderstanding, you able to close this PR. 🙏

tpolecat:
Thanks!

PR #1122, opened 2020-01-13
bump Scala version to 2.13.1 (was 2.13.0)

SethTisue:
context: I'm trying to get doobie green in the Scala 2.13 community build, over at scala/community-build#1033

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,7 +9,7 @@ lazy val collCompatVersion    = "2.1.2"
 lazy val fs2Version           = "2.1.0"
 lazy val h2Version            = "1.4.200"
 lazy val hikariVersion        = "3.4.1"
-lazy val kindProjectorVersion = "0.10.3"
+lazy val kindProjectorVersion = "0.11.0"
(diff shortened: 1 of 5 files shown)


SethTisue:
```
[error] /Users/tisue/doobie/modules/core/src/main/scala/doobie/util/fragments.scala:16:30: reference to Put is ambiguous;
[error] it is both defined in trait Types and available as class Put in package util
[error]   def in[F[_]: Reducible, A: Put](f: Fragment, fs: F[A]): Fragment =
[error]                              ^
```

which is the same thing I'm seeing in the community build, so that's good

offhand, my guess as the root cause would be https://github.com/scala/scala/pull/7609

SethTisue:
okay, CI is green now. (I hope the 7609-induced source tweaks I blindly made are sensical, but they are ugly; you might prefer to redo this a different way before merging)

tpolecat:
This is fine. It only happens internally when I do `package doobie; package util` which is just me being lazy. Thanks!

PR #1123, opened 2020-01-14
postgresql copy stream (2)

tpolecat:
This continues the work from #1058


tpolecat:
@Z1kkurat do these changes make sense to you? I think it's good to go.

modules/postgres/src/main/scala/doobie/postgres/syntax/FragmentSyntax.scala
```diff
@@ -23,14 +28,39 @@ class FragmentOps(f: Fragment) {
...
   }
 
+  /**
+   * Given a fragment of the form `COPY table (col, ...) FROM STDIN` construct a
+   * `ConnectionIO` that inserts the values provided by `stream`, returning the number of affected
+   * rows. Chunks input `stream` for more efficient sending to `STDIN` with `minChunkSize`.
+   */
+  def copyIn[F[_]: ConcurrentEffect, A: Text](
+    stream: Stream[F, A],
+    minChunkSizeBytes: Int
```
Z1kkurat: @tpolecat I'm not quite sure about `Bytes` part in the name, since it can be misleading. We chunk `Stream` of `A` before converting to string and then to bytes.

Z1kkurat:
@tpolecat Except minor naming issue, everything else looks good to me.

PR #1145, opened 2020-02-16
du -h -d 6 $HOME/.cache/coursier/

sideeffffect:


tpolecat:
Hm why is it there instead of `~/.coursier`?

sideeffffect:
Coursier uses `~/.cache/coursier` as cache directory

tpolecat:
Closing, no longer using Travis.

Issue #1147, opened 2020-02-18
check example is wrong


tpolecat:
> hi Rob, just a heads up. in the book of doobie, chapter 6, https://tpolecat.github.io/doobie/docs/06-Checking.html, the first .check output seems not to be the right one: it should show errors but it’s showing everything OK

PR #1152, opened 2020-02-24
Fix Typechecking Queries example's output

heraklos:
Fix for #1147 
It seemed that the first `.check` called the second `biggerThan`, so I just changed the second function's name (naming followed the pattern used for ` Country` and `Country2`)


tpolecat:
Thanks!

PR #1155, opened 2020-02-26
Replace deprecated ? with *

takayahilton:
https://github.com/typelevel/kind-projector

>?, +?, and -? (deprecated syntax; same meaning as the above)


tpolecat:
It's going to take me a while to get used to this. Thanks!

PR #1167, opened 2020-03-20
Fix docsite

calvellido:
This PR basically creates a menu for the book sidebar, and fix main page horizontal menu positions.

If you feel like styles are misbehaving on some pages, check that you executing locally something like this (form the project root)

`jekyll serve -s ./modules/docs/target/site -b /doobie`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,6 @@ bin/
 # Metals
 .metals/
 project/metals.sbt
+
+# Jekyll
+_site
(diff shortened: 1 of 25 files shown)


tpolecat:
Man I appreciate this so much, thank you!

tpolecat:
Hm do you have any idea why Introduction comes last in the TOC instead of first?

image

calvellido:
Weird, can't think on anything that could cause that :thinking: 

I disabled all `mdoc` to generate this, but shouldn't change the outcome:

image

Do you have any more local changes beyond what it's in this branch?

calvellido:
@tpolecat try this though: https://github.com/tpolecat/doobie/pull/1169

PR #1171, opened 2020-03-23
circe 0.13.0

sideeffffect:
I presume we don't want to upgrade to a (breaking) new version of Circe in the `0.8.x` series, so I've open this PR targeting `master` branch


tpolecat:
thanks, closed via #1180

PR #1176, opened 2020-03-30
WIP: switch to paradox

tpolecat:
I have had it with Jekyll so this switches to Paradox.

- [x] update travis config
- [x] document how to build and release the doc because i already forgot


tpolecat:
you can do it travis

tpolecat:
Ok this is passing in Travis but we're getting no status back. I'm overriding and merging.

PR #1188, opened 2020-04-16
sql-interpolation for Update and Query

teigen:
why:
sql interpolation is limited to binding concrete values only, typically as a way of producing Update0 and Query0 instances.
Having to switch to Update or Query with plain strings and explicit `?` to use updateMany is confusing
for beginners, and a bit annoying for everyone else. This change adds support for also building
Update and Query instances through sql-interpolation.

how:
`Fragment.Elem` gets two more cases `FunArg` and `FunOpt` for functions.
`Fragment.Elem`, `Fragment` and `SingleFragment` all gets a contravariant type parameter for keeping track of the input type to these functions.

To provide fairly good source compatibility with existing code, all methods on
Fragment touching the contravariant type is moved to two new implicit syntax-classes
Fragment0Syntax for Fragment[Any] providing the same api as today, yielding the same Update0 and Query0 instances,
and FragmentSyntax yielding Update and Query instances.

notes:
FirstExampleLens demonstrates interpolated Update using doobie lenses.
FirsExampleMonocle demonstrates how to add monocle support to reduce boilerplate
I also added docker-compose file in a separate commit to easily set up a working environment to run the tests.

warning:
this probably needs a lot more testing, but I wanted to see if there was any interest in this change before I pour more hours into this.

### DIFF
--- a/modules/core/src/main/scala/doobie/util/fragment.scala
+++ b/modules/core/src/main/scala/doobie/util/fragment.scala
@@ -7,9 +7,8 @@ package doobie.util
 import cats._
 import cats.data.Chain
 import cats.implicits._
-
-import doobie._, doobie.implicits._
-import doobie.util.pos.Pos
+import doobie._
+import doobie.implicits._
 import doobie.enum.Nullability._
 import doobie.util.pos.Pos
 import java.sql.{ PreparedStatement, ResultSet }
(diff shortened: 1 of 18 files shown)


hamnis:
This would be great to have indeed.

tpolecat:
Hi, sorry for the delay. This is very interesting. I had tried to figure out a way to do this and had not considered interpolating lenses (really all you here is a getter). I need to think about this some more but it's promising. Thanks!

kareblak:
In absolute support of this feature. Has been bugging me for years.

modules/core/src/main/scala/doobie/util/fragment.scala
```diff
@@ -24,156 +23,209 @@ object fragment {
...
+    val empty: Fragment[Any] =
       const0("")
 
     /** Statement fragments form a monoid. */
-    implicit val FragmentMonoid: Monoid[Fragment] =
-      new Monoid[Fragment] {
-        val empty = Fragment.empty
-        def combine(a: Fragment, b: Fragment) = a ++ b
+    implicit def FragmentMonoid[R]: Monoid[Fragment[R]] =
+      new Monoid[Fragment[R]] {
```
Daenyth, line 217: Since `R` is in a phantom position, might be worth caching+casting this instance to avoid constructing more than one.

Daenyth (commented):
This is pretty exciting :)

PR #1197, opened 2020-04-25
Turn strategy into a polymorphic class.

diesalbla:


tpolecat:
Hi, sorry I haven't had a chance to look at this. What's the use case for this generalization?

diesalbla:
> Hi, sorry I haven't had a chance to look at this. What's the use case for this generalization?

I must be sorry as well... had forgotten about this. At the time, it crossed my mind if we could build `doobie` into a direct interpretation, so-called "tagless", without allocating the `Free` objects. Hence the idea of decoupling from `ConnectionIO`.

tpolecat:
> At the time, it crossed my mind if we could build doobie into a direct interpretation, so-called "tagless", without allocating the Free objects.

Yes, I have looked into this as well and it's a huge simplification in the implementation but a catastrophic change from the user's point-of-view, so it's a doobie-2 kind of thing. I'm going to close this for now because I don't think there's much immediate benefit but I may come back to it later.

Thanks!

PR #1202, opened 2020-04-30
fail renamed to raiseError

jacobm:
It looks like `fail` was renamed to `raiseError` to be consistent:

https://github.com/typelevel/cats-effect/issues/32

(and thanks for doobie, it is such a pleasure to work with).


anthonyjoeseph:
shouldn't it be `.raiseError`, not `raiseError`?

anthonyjoeseph:
Maybe also mention `doobie.free.connection.raiseError`? It's hard to find but I think it's super useful

tpolecat:
Thanks!

PR #1233, opened 2020-07-13
Add doobie-munit package to integrate Doobie with the shiny new MUnit

Kazark:
[MUnit](https://scalameta.org/munit/) is a [the new Scala testing framework on the block](https://www.youtube.com/watch?v=_2qdZNKlNbM). Seems reasonable for Doobie to support it along with the other two it already supports.

I copied off `scalatest` because it is fairly similar to MUnit. Hopefully I didn't forget to change anything after the copy and paste, particularly in the documentation (the test is green, at least).


modules/docs/src/main/mdoc/docs/06-Checking.md
```diff
@@ -99,7 +99,7 @@ def biggerThan2(minPop: Int) =
 biggerThan2(0).check.unsafeRunSync
 ```
 
-**doobie** supports `check` for queries and updates in three ways: programmatically, via YOLO mode in the REPL, and via the `doobie-specs2` and `doobie-scalatest` packages, which allow checking to become part of your unit test suite. We will investigate this in the chapter on testing.
+**doobie** supports `check` for queries and updates in four ways: programmatically, via YOLO mode in the REPL, and via the `doobie-specs2`, `doobie-scalatest` and `doobie-munit` packages, which allow checking to become part of your unit test suite. We will investigate this in the chapter on testing.
```
kevinmeredith, line 102: Take or leave it: might a bulleted list be easier to view?

modules/docs/src/main/mdoc/docs/13-Unit-Testing.md
```diff
@@ -129,3 +129,25 @@ Details are shown for failing tests.
...
 ```
+
+### The MUnit Package
+
+The `doobie-munit` add-on provides a mix-in trait that we can add to any `Assertions` implementation (like `FunSuite`) much like the ScalaTest package above.
+
+```scala mdoc:silent
+import munit._
+
+class AnalysisTestScalaCheck extends FunSuite with doobie.munit.IOChecker {
```
kevinmeredith, line 140: Does `FunSuite` come from munit, @Kazark ?

Kazark: Yes.

modules/docs/src/main/mdoc/docs/06-Checking.md
Kazark: I think that's a reasonable idea; the general principle of the PR was to try to stick closely with what's there. We'll see what @tpolecat says.

alejandrohdezma:
Hi @tpolecat, any chance on taking a look at this? I've been copying @Kazark implementation around and it will be super if we could have this integration out of the box 😸

tpolecat:
This looks fine, sorry for the delay. I'll merge it if you bring it up to date.

Kazark:
Thanks @alejandrohdezma

alejandrohdezma:
@tpolecat done! I've included and update of the `setup-scala` Github Action since the build was broken without it

alejandrohdezma:
Hey @tpolecat, I have conform `doobie-munit` with the latest changes in `doobie-specs2`. Do you think this is ready to merge now?

modules/docs/src/main/mdoc/docs/06-Checking.md
jatcwang: idm what's currently here, since it's featured prominantly in 13-Unit-Testing.md :)

modules/munit/src/test/scala/doobie/munit/CheckerTests.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  implicit def contextShift: ContextShift[M]
+
+  lazy val transactor = Transactor.fromDriverManager[M](
+    "org.h2.Driver",
+    "jdbc:h2:mem:queryspec;DB_CLOSE_DELAY=-1",
+    "sa", ""
+  )
+
+  test("trivial") { check(sql"select 1".query[Int]) }
```
jatcwang, line 23: Can we assert a failing test as well? You know, just in case ;)

alejandrohdezma: Done https://github.com/tpolecat/doobie/pull/1233/commits/539e1da3896193400ddfdaf458130c27eeea9b57

jatcwang (approved):

FunFunFine:
So when can we expect this one released?

jatcwang:
There are a few thing left that we need to release 0.11 so I've asked Rob to look at a few important PRs (including this one). No ETA sorry.

modules/munit/src/main/scala/doobie/munit/analysisspec.scala
```diff
@@ -0,0 +1,67 @@
...
+
+    private def checkImpl(args: AnalysisArgs)(implicit loc: Location) = {
+      val report = analyzeIO(args, transactor).unsafeRunSync()
+      if (!report.succeeded) {
+        fail(
+          formatReport(args, report, colors)
+            .padLeft("  ")
+            .toString
+        )
+      }
```
ValdemarGr, line 58: Wouldn't https://github.com/typelevel/munit-cats-effect be better instead?
```suggestion
      val report = analyzeIO(args, transactor)
      report.flatMap{
          case r if !r.succeeded => 
            IO(fail(
              formatReport(args, report, colors)
                .padLeft("  ")
                .toString
            ))
          case _ => IO.unit
       }
```

ValdemarGr: I have implemented the munit-cats-effect integration at 
https://github.com/ValdemarGr/doobie/tree/munit
Can someone give any insight or consideration for the idea?

jatcwang: Sounds like a good idea. It'll also make CE3 migration easier

ValdemarGr: This implementation also changes the semantics of check; it is no longer eager, is this desired?
The scalatest and specs2 test utilities seem to be eager.

jatcwang: If execution thread gets to `check(..)` then we're running the test already right? (This PR's impl seems eager to me, unless I'm missing something)

munit (and all other test frameworks that I know) all use call-by-name to allow skipping tests, so `test("asdf") { check(..) }` is still skippable even if `check` is eager.

ValdemarGr: I mean the evaluation of the check effect.
In the initially proposed solution check runs eagerly which fails the following test:
https://github.com/ValdemarGr/doobie/blob/6ad3f9414bc7cb6f1a1bbe1e65789315bc38dd08/modules/munit/src/test/scala/doobie/munit/CheckerTests.scala#L31-L34
In the version I published the test combinator is wrapped here:
https://github.com/typelevel/munit-cats-effect/blob/b0845550c337a21d72c46f9e2619a02a4479ad88/common/shared/src/main/scala/munit/CatsEffectFunFixtures.scala#L68-L79
Thus the test succeeds in the munit-cats-effect variant:
https://github.com/ValdemarGr/doobie/blob/ebbd0ccc475bf21dd5ebee4967741c4ff0263992/modules/munit/src/test/scala/doobie/munit/CheckerTests.scala#L33-L36

jatcwang: Ah yeah I think the other integrations (scalatest/specs) run it eagerly. Perhaps it is worth keeping that behaviour to make it easier for people to migrate between test libs. (I know that in my codebase I have a few `check(..)` being called as if they're statements)

So I guess cats-effect munit won't help us here. Sorry!

hamnis:
at $WORK we would really like to leave scalatest in the dust, so this would help a lot. Can we get this going again?

jatcwang:
Sorry just saw that I needed to approve the CI.
@tpolecat I've approved the PR. Merge it if you're happy with it too :)

jatcwang:
Thanks for the PR and sorry for the delay! I'll get this merged into the cats-effect 3 (1.x) branch too.

PR #1246, opened 2020-08-18
Query cancelation

mikkka:
Main reason of statements to be uncancelable is using Statement#query in acquire section of Bracket to guarantee ResultSet#close. But we don't have to do this: https://docs.oracle.com/javase/7/docs/api/java/sql/Statement.html#close(), close on Statement also calls close on ResultSet. So it's possible to use Concurrent + Statement#cancel() to achieve query execution cancelation behaviour.


modules/core/src/test/scala/doobie/util/strategy.scala
```diff
@@ -165,22 +167,6 @@ class strategyspec extends Specification {
 
   }
 
-  "ResultSet safety" >> {
```
Daenyth, line 168: It doesn't seem safe to me to remove this test - why remove it?

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
```diff
@@ -827,47 +831,70 @@ trait KleisliInterpreter[M[_]] { outer =>
...
+        blocker.blockingContext.execute(new Runnable {
+          def run(): Unit = 
+            try {
+              cb(Right(f(a)))
+            } catch {
+              case scala.util.control.NonFatal(e) => cb(Left(e))
+            }
+        })
+        concurrent.delay{
+          a.cancel()
```
Daenyth, line 867: ❓ Should the call to `cancel` be moved to the `Blocker`?

modules/core/src/test/scala/doobie/util/strategy.scala
mikkka: Because I removed ReslutSet#close, and it's safe to that, according to jdbc documentation.

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
mikkka: Yep, I'll fix this.

mikkka: Fixed

mikkka:
So, will it be merged in near future? I'd like to see this functionality in my work projects without building my own fork artefact.

Daenyth:
@mikkka Thanks for reminding us. I'll try to get it merged soon.

tpolecat (changes requested):
Unfortunately `KleisliInterpreter` is generated code so these changes need to be folded into `FreeGen2.scala` somehow. This will be complex because only statements are cancelable.

Also can you explain your changes re: `ResultSet.close`?

mikkka:
> Unfortunately `KleisliInterpreter` is generated code so these changes need to be folded into `FreeGen2.scala` somehow. This will be complex because only statements are cancelable.

Yes, it's harder than I thought. I'am trying to modify codegen.

> Also can you explain your changes re: `ResultSet.close`?

Bracketing query to close ResultSet prevents query to be cancelable, so I made `close` with guarantee, in very exceptional cases ResultSet `close` will be triggered with Statement/PreparedStatement `close`.

mikkka:
> Unfortunately `KleisliInterpreter` is generated code so these changes need to be folded into `FreeGen2.scala` somehow. This will be complex because only statements are cancelable.

I've modified code gen.

mikkka:
@tpolecat @Daenyth Would you kindly review the changes?

project/FreeGen2.scala
```diff
@@ -425,7 +431,28 @@ class FreeGen2(managed: List[Class[_]], pkg: String, renames: Map[Class[_], Stri
...
+   def commonMethods = s"""
+      |  // Some methods are common to all interpreters and can be overridden to change behavior globally.
+      |  def primitive[J, A](f: J => A): Kleisli[M, J, A] = Kleisli { a =>
+      |    // primitive JDBC methods throw exceptions and so do we when reading values
+      |    // so catch any non-fatal exceptions and lift them into the effect
+      |    blocker.blockOn[M, A](try {
+      |      asyncM.delay(f(a))
+      |    } catch {
+      |      case scala.util.control.NonFatal(e) => asyncM.raiseError(e)
+      |    })(contextShiftM)
```
Daenyth, line 445: Is this right? I don't think it is - `delay` takes a by-name argument, and so `f(a)` is not evaluated inside of a `try`, so the `try` will never catch any errors. Further, `delay` itself catches `NonFatal`, so why do this explicitly? It's very unclear to me

project/FreeGen2.scala
```diff
@@ -521,7 +603,8 @@ class FreeGen2(managed: List[Class[_]], pkg: String, renames: Map[Class[_], Stri
     val ki = {
       val file = new File(base, s"kleisliinterpreter.scala")
       val pw = new PrintWriter(file)
-      pw.println(kleisliInterpreter)
+      if (base.getPath().contains("postgres")) pw.println(kleisliInterpreter)
+      else pw.println(kleisliInterpreterCancelable)
```
Daenyth, line 607: What's the specific check here for?

Daenyth (commented):
- It's likely that we'll want to have tests explicitly covering cancelation to ensure that it does what we want. I didn't see any (I might have missed them; it's a large PR)
- It's possible that we want to see if it's possible to have cancelable and non-cancelable Transactor construction methods, to support backwards compatibility
- It's possible we actually want to just wait for cats-effect 3, because I expect the implementation gets much simpler, and we may want to avoid people needing to do *two* codebase migrations in short order for a marginal feature.

Beyond those thoughts, I can't add much - I'm just not familiar enough with the deep parts of the code here. Rob's feedback will likely be more valuable.

project/FreeGen2.scala
mikkka: `doobie.postgres.free.KleisliInterpreter` doesn't need cancel feature at all, so we can generate ordinary interpreter.

project/FreeGen2.scala
mikkka: It's copypaste from previous implementation - I just splitted big string into smaller pieces. So, I assume it's correct.

mikkka:
> * It's likely that we'll want to have tests explicitly covering cancelation to ensure that it does what we want. I didn't see any (I might have missed them; it's a large PR)

I agree, so I'll try to create tests for these.

> * It's possible that we want to see if it's possible to have cancelable and non-cancelable Transactor construction methods, to support backwards compatibility

I think there is no need in splitting, because usually one construct transactor on edge of application, where all needed instances exist. 

> * It's possible we actually want to just wait for cats-effect 3, because I expect the implementation gets much simpler, and we may want to avoid people needing to do _two_ codebase migrations in short order for a marginal feature.

I disagree with this point, it will be to long to wait for cats-effect 3.

> 
> Beyond those thoughts, I can't add much - I'm just not familiar enough with the deep parts of the code here. Rob's feedback will likely be more valuable.

Thank you for feedback.

mikkka:
It turned out that solution is much simpler, so I'am closing this and going to open another.

unoexperto:
@mikkka Please post url to new PR here so I can subscribe.

mikkka:
> @mikkka Please post url to new PR here so I can subscribe.

https://github.com/tpolecat/doobie/pull/1357

PR #1253, opened 2020-08-25
Doc for using fragments inside fr interpolator

jatcwang:
Documentation for #1045


modules/docs/src/main/mdoc/docs/08-Fragments.md
```diff
@@ -76,6 +76,16 @@ count("city").quick.unsafeRunSync()
 ```
 
 > Note that `Fragment.const` performs no escaping of passed strings. Passing user-supplied data is an **injection risk**.
+>
+
+You can also use Fragments inside `fr` interpolator directly. Both parameters and SQL literals will be
```
Daenyth, line 81: I think it's worth including a "note" section on sql injection, and in what ways this is and is not safe, injection-wise

jatcwang: I think this it is only a concern for Fragment.const right? Perhaps we need more emphesis there?

Daenyth: I think just to mention that as long as the interpolated fragment is constructed safely, then this is safe.

jatcwang: Sure I'll see how I can fit it in. (Though I'm not sure how it is different from `++`)

jatcwang: done

jatcwang:
Force of habit :) Fixed

PR #1276, opened 2020-09-25
Add CopyManager example

rzeigler:
Example of COPY ... FROM STDIN with csv.

Per discussion at https://gitter.im/tpolecat/doobie?at=5f6ccc03417bf140aa0e7e13


jatcwang:
Thanks. Think we should add a FAQ to make it more discoverable?

rzeigler:
> 
> 
> Thanks. Think we should add a FAQ to make it more discoverable?

Happy to add whatever ya'll decide.

Daenyth (approved):
Happy to merge as-is.

Extras if you feel inclined:
- Add some doc comments explaining the step by step
- Add an example using fs2, taking a `Stream[F, Byte]` to `InputStream` (with fs2-io), and then using that with the copy manager, since that's a much more common example. I have some code that would do this, I'll see if I can pull it out of my git history in the relevant project (long since deleted iirc)

rzeigler:
Let me know how you feel about this copy.

modules/example/src/main/scala/example/PostgresCopyInCsv.scala
```diff
@@ -0,0 +1,61 @@
...
+  )
+
+  val csv = """name,food
+                |piglet,haycorns
+                |eeyore,thistles
+                |pooh,honey
+                |tigger,extract of malt""".stripMargin
+
+  // The postgres driver expects and InputStream containing the data to load
+  val is: InputStream = new ByteArrayInputStream(csv.getBytes(StandardCharsets.UTF_8))
```
Daenyth, line 30: 🤔 Does it matter to properly wrap this with `IO.delay`? Normally you would, because BAIS contains state, so allocating a new instance isn't referentially transparent. I'm not sure it matters enough in this example. I don't have a good sense of where people would be at, when reading the docs. If they know `IO` they'd know that they need to wrap it, and in many code samples it would end up being *probably* ok by coincidence if they don't. But it is incorrect.

Up to you I think - not a blocker to merge, at all. We could clean it after the fact if need be

modules/example/src/main/scala/example/PostgresCopyInCsv.scala
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  def run(args: List[String]): IO[ExitCode] = {
+    // Construct a copy from the input stream we have lying around
+    val simple = (createTable >> copyIn(is)).transact(xa)
+    // Construct a copy from a Stream[IO, Byte] using fs2.io
+    val fromByteStream = io.toInputStreamResource(byteStream)
+      .use(is => (createTable >> copyIn(is)).transact(xa))
+
+    (simple |+| fromByteStream)
```
Daenyth, line 55: For what it's worth, I found it surprising to use `|+|` here, since more often to sequence `IO` operations together you'd see `>>`.

If this is intentionally to merge the results, it may help to have a comment to that effect.

I expect that what I'd see here is that `ct` below is 2*(`csv` line count) - is that right?

Daenyth (approved):
I'll leave this open to Monday if you'd prefer to make further revisions, but I'm also happy to have it as is.

modules/example/src/main/scala/example/PostgresCopyInCsv.scala
rzeigler: That’s probably a fair point. It doesn’t add clutter and it’s more correct.

modules/example/src/main/scala/example/PostgresCopyInCsv.scala
rzeigler: I was adding the column counts to see that both copies inserted 4 records. I will attach the print to each action instead.

rzeigler:
> I'll leave this open to Monday if you'd prefer to make further revisions, but I'm also happy to have it as is.

I will make a few more revisions to address your points.

modules/example/src/main/scala/example/PostgresCopyInCsv.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  val fromByteStreamExample = 
+    // Construct a copy by converting a Stream[IO, Byte] into an InputStream
+    io.toInputStreamResource(byteStream)
+      .use(is => (createTable >> copyIn(is)).transact(xa))
+      .map(_.toString)
+      .flatMap(ct => putStrLn(show"loaded $ct from Stream[IO, Byte]"))
+
+  def run(args: List[String]): IO[ExitCode] = {
+    // Should print 4 twice
```
Daenyth, line 65:
```suggestion
    // Should print "loaded 4 from Stream[IO, Byte]" twice
```

Daenyth:
@rzeigler Thanks very much for the contribution! 👍

PR #1280, opened 2020-10-06
cats-effect 3 upgrade

rkrzewski:
I would like to start the ball rolling on the upgrade to Cats Effect 3.

I made made some fairly mechanical changes by chasing compilation errors, but when I was fixing `kleisliinterpreter.scala` I've realized that I need to step back and adjust all the free algebras to the updated `Async` interface.

In order to start with a clean slate I added `wip.scala` that has a free algebra with no domain functionality - it just encodes `Async` in a free monad, and provides an interpreter back to `Async`. I was able to get it to compile but I would rally appreciate a review of that file from @djspiewak or @SystemFw before I continue. If I'm doing something dumb, I'd like to fix it now and not after I smear if all over the umpteen algebras :smiley_cat:


build.sbt
```diff
@@ -2,7 +2,7 @@ import FreeGen2._
 
 // Library versions all in one place, for convenience and sanity.
 lazy val catsVersion          = "2.1.1"
-lazy val catsEffectVersion    = "2.1.4"
+lazy val catsEffectVersion    = "3.0-805b021"
 lazy val circeVersion         = "0.13.0"
 lazy val collCompatVersion    = "2.1.6"
 lazy val fs2Version           = "2.4.2"
```
mpilquist, line 8: `3.0-9d83065` is compatible with this version of CE3

djspiewak: Also I'm working on getting actual-M1 out right now.

djspiewak:
> it just encodes Async in a free monad

First off, I'm a little uncertain whether this is 100% possible. Second, I know Doobie is almost entirely within `Free`, but is this kind of thing really needed? :-)

build.sbt
rkrzewski: @mpilquist you mean fs2 version right?

rkrzewski:
> First off, I'm a little uncertain whether this is 100% possible. Second, I know Doobie is almost entirely within `Free`, but is this kind of thing really needed? :-)

No idea if it's possible or necessary, I'm just trying to replicate what was done in Doobie for CE2 :)

build.sbt
mpilquist: Yep

rkrzewski:
After moving code from `wip.scala` into `blob.scala` and updating `kleisliiterpreter.scala` with the new common operations for all algebras it's more clear to me what @djspiewak meant saying encoding `Async` as free monad may be impossible.
There's a problem with the methods on `Async[F]` that refer to `F[_]` more than once in their signature: `uncancelable`, `start`, `racePair`, `ref`, `deferred` and `cont`. When trying to define the `BlobOp ~> F` I ended up with weird mixed signatures like
```
def uncancelable[A](body: Poll[F] => BlobIO[A]): F[A]
```
which in turn make the `Op` classes for those elements of the algebra unimplementable :disappointed:

SystemFw:
I don't think it makes sense semantically for `ConnectionIO` to be `Concurrent`, and since in ce3 `Async <: Concurrent`, I don't think it can be `Async` either. I would think that `Sync` + `LiftIO` should be enough for practical use cases, and afaik the general guidance is for users to use `LiftIO` to embed things in `ConnectionIO`.

Maybe @tpolecat can weigh in on the need for `Async` in the current version?

rkrzewski:
I'm happy to report that I've made good progress with this pull request.

Right now all preexisting tests are passing for me locally except the following:
```
[error]   x write+read DATE as NonEmptyList(Some(java.sql.Date))
[error]    Falsified after 12 passed tests.
[error]    > ARG_0: 1900-01-01
[error]    The seed is EB65_-MlwbYfwP16wYkCn3DkJDkAU2w_ITfV-EaJZtL=
[error]    
[error]    > Right(1899-12-31) != Right(1900-01-01) (h2types.scala:57)
[error]   x write+read DATE as Option[NonEmptyList(Some(java.sql.Date))] (Some)
[error]    Falsified after 16 passed tests.
[error]    > ARG_0: 1900-01-01
[error]    The seed is KeIkRoIGSYfZUhqnC3etYCySZNqt9gfb6618-USalhA=
[error]    
[error]    > Right(Some(1899-12-31)) != Right(Some(1900-01-01)) (h2types.scala:60)
```
but those two are also failing for me on `master` branch, so I didn't dig into that.

The tests seem to hang on Github since a few builds. Does anyone knows what's up with that? They run to completion on my machine every time.

I still need to propagate several `Async` operations (`uncancelable`, `ref` and `deferred`) to all algebras, because right now I have them only in places when they were necessary to get the tests to pass.

I need @SystemFw to take a look at the encoding for `uncancelable` I came up with - you said it was impossible, but I'm just a simple code monkey so I kept mashing things together until they worked :rofl: The important bits are here:
https://github.com/rkrzewski/doobie/blob/0fd2e101794e30779fef58143bc236f33adee57e/modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala#L131
https://github.com/rkrzewski/doobie/blob/0fd2e101794e30779fef58143bc236f33adee57e/modules/free/src/main/scala/doobie/free/connection.scala#L376

I'm looking forward to https://github.com/typelevel/cats-effect/issues/1299 because I needed a `F ~> ConnectionIO` given `Async[F]` in a couple places. I came up with helper function for that:
https://github.com/rkrzewski/doobie/blob/cats-effect-3/modules/core/src/main/scala/doobie/util/effects.scala 
but I'd love to get rid of the `UnsafeRun` constraint there.

djspiewak:
I have a relatively dumb version of it locally right now. I'm working on encoding a smarter non-blocking version. Expect it sometime this week.

SystemFw:
That's great news!
Although I'm still not sure what something like `start` means in `ConnectionIO`

rkrzewski:
> Although I'm still not sure what something like `start` means in `ConnectionIO`

Oh, right. I've left `start` and `racePair`as `rairseError(new Exception("Unimplemented")`. But I haven't been able the get the tests to run without implementing `uncancelable` and `ref` for `ConnectionIO`. The latter was straightforward, the former - not so much :slightly_smiling_face:

tpolecat:
> I don't think it makes sense semantically for ConnectionIO to be Concurrent, and since in ce3 Async <: Concurrent, I don't think it can be Async either. I would think that Sync + LiftIO should be enough for practical use cases, and afaik the general guidance is for users to use LiftIO to embed things in ConnectionIO.

@SystemFw I think `MonadError` + `LiftIO` is probably the minimum we can get away with. It would be nice to have `MonadCancel` because there are a handful of operations that can in fact be canceled. I think I prefer not to have `Sync` at all, I don't think it's necessary,

SystemFw:
>  think MonadError + LiftIO is probably the minimum we can get away with.

This should be fine

>  It would be nice to have MonadCancel

There is hope for this I think, my worries were more for the Spawn and Concurrent operations.

> Apologies for not weighing in earlier, I haven't had much attention to spare in the past few weeks.

ah no worries! From the messages on scala Gitter I figured you were probably in the weeds with something :)

rkrzewski:
@SystemFw 
> Although I'm still not sure what something like `start` means in `ConnectionIO`

I pondered on it a bit, and came to conclusion that `start` and `racePair` would compile to a program doing concurrent operations against a single `Connection` (or a `ResultSet` etc). This is definitely not useful and would break in "interesting" ways, because JDBC objects are not threadsafe per specification. So we don't want that, but as you said `Async <: Concurrent` now and I was unable to get the tests to run without implementing most of the `Async[ConnectionIO]` machinery. I can show the pieces that break without `async`, `ref`, and so on. Maybe that'll give some ideas how to move forward?

>  the general guidance is for users to use LiftIO to embed things in ConnectionIO

Limiting the API surface area like that sounds great, but how that work with polymorphic effects aka tagless final style? `LiftIO[ConnectionIO]` allows converting concrete `IO` values into `ConnectionIO`. How do we convert an opaque `F[_]` to `IO`? What constraints on `F` would allow that?

rkrzewski:
Here are some examples what happens when I stub out some pieces of `Async[ConnectionIO]` instance:

without `Sync.suspend`

in `doobie.util.transactorspec`
```
[info] Connection lifecycle (streaming)
[error]   ! Connection.close should be called on success
[error]    java.lang.Exception: haha nope (connection.scala:457)
[error] doobie.free.connection$$anon$3.suspend(connection.scala:457)
[error] doobie.free.connection$$anon$3.suspend(connection.scala:448)
[error] cats.effect.kernel.Sync.delay(Sync.scala:40)
[error] cats.effect.kernel.Sync.delay$(Sync.scala:39)
[error] doobie.free.connection$$anon$3.delay(connection.scala:448)
[error] cats.effect.kernel.Ref$SyncRef.modify(Ref.scala:348)
[error] fs2.internal.CompileScope.$anonfun$open$5(CompileScope.scala:143)
[error] cats.free.Free.$anonfun$step$1(Free.scala:52)
[error] cats.free.Free.step(Free.scala:53)
[error] cats.free.Free.$anonfun$foldMap$1(Free.scala:153)
[error] cats.data.KleisliFlatMap.$anonfun$tailRecM$2(Kleisli.scala:624)
[error] cats.StackSafeMonad.tailRecM(StackSafeMonad.scala:16)
[error] cats.StackSafeMonad.tailRecM$(StackSafeMonad.scala:15)
[error] cats.effect.IO$$anon$1.tailRecM(IO.scala:496)
[error] cats.StackSafeMonad.$anonfun$tailRecM$1(StackSafeMonad.scala:17)
[error] cats.effect.IOFiber.flatMapK(IOFiber.scala:993)
[error] cats.effect.IOFiber.succeeded(IOFiber.scala:797)
[error] cats.effect.IOFiber.mapK(IOFiber.scala:985)
[error] cats.effect.IOFiber.succeeded(IOFiber.scala:796)
[error] cats.effect.IOFiber.runLoop(IOFiber.scala:230)
[error] cats.effect.IOFiber.afterBlockingSuccessfulR(IOFiber.scala:947)
[error] cats.effect.IOFiber.run(IOFiber.scala:139)
[error] cats.effect.unsafe.WorkerThread.runFiber(WorkerThread.scala:138)
[error] cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:271)
```

without `GenConcurrent.ref`

in `doobie.util.transactorspec`
```
[info] Connection lifecycle (streaming)
[error]   ! Connection.close should be called on success
[error]    java.lang.Exception: haha nope (connection.scala:465)
[error] doobie.free.connection$$anon$3.ref(connection.scala:465)
[error] doobie.free.connection$$anon$3.ref(connection.scala:448)
[error] fs2.Compiler$Target$ConcurrentTarget.ref(Compiler.scala:139)
[error] fs2.internal.CompileScope$.fs2$internal$CompileScope$$apply(CompileScope.scala:416)
[error] fs2.internal.CompileScope$.$anonfun$newRoot$1(CompileScope.scala:421)
[error] cats.free.Free.$anonfun$step$1(Free.scala:52)
[error] cats.free.Free.step(Free.scala:53)
[error] cats.free.Free.$anonfun$foldMap$1(Free.scala:153)
[error] cats.data.KleisliFlatMap.$anonfun$tailRecM$2(Kleisli.scala:624)
[error] cats.StackSafeMonad.tailRecM(StackSafeMonad.scala:16)
[error] cats.StackSafeMonad.tailRecM$(StackSafeMonad.scala:15)
[error] cats.effect.IO$$anon$1.tailRecM(IO.scala:496)
[error] cats.data.KleisliFlatMap.$anonfun$tailRecM$1(Kleisli.scala:624)
[error] doobie.free.KleisliInterpreter.$anonfun$uncancelable$2(kleisliinterpreter.scala:132)
[error] cats.effect.IOFiber.runLoop(IOFiber.scala:319)
[error] cats.effect.IOFiber.afterBlockingSuccessfulR(IOFiber.scala:947)
[error] cats.effect.IOFiber.run(IOFiber.scala:139)
[error] cats.effect.unsafe.WorkerThread.runFiber(WorkerThread.scala:138)
[error] cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:271)
```

without `Async.async`: 

in `doobie.postgres.textspec`

```
[info] copyIn should
[info]   + correctly insert batches of rows
[error]   ! correctly insert batches of rows via Stream
[error]    java.lang.Exception: Exception raised on property evaluation.> ARG_0: List()
[error]    > ... > Exception: java.lang.Exception: haha nope
[error]    The seed is BfyN3NoT5Sd4U5vOoLhtv1etV7J9-r15-TerVcNK0iI= (connection.scala:470)
[error] doobie.free.connection$$anon$3.async(connection.scala:470)
[error] doobie.free.connection$$anon$3.async(connection.scala:448)
[error] doobie.util.effects$$anon$1.$anonfun$apply$3(effects.scala:17)
[error] cats.free.Free.$anonfun$step$1(Free.scala:52)
[error] cats.free.Free.$anonfun$foldMap$3(Free.scala:156)
[error] cats.effect.IOFiber.mapK(IOFiber.scala:976)
[error] cats.effect.IOFiber.succeeded(IOFiber.scala:796)
[error] cats.effect.IOFiber.runLoop(IOFiber.scala:230)
[error] cats.effect.IOFiber.afterBlockingSuccessfulR(IOFiber.scala:947)
[error] cats.effect.IOFiber.run(IOFiber.scala:139)
[error] cats.effect.unsafe.WorkerThread.runFiber(WorkerThread.scala:138)
[error] cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:271)
```

In the last one uses the gadget I wrote to compensate for removal of `AsyncEffect`:
https://github.com/rkrzewski/doobie/blob/cats-effect-3/modules/core/src/main/scala/doobie/util/effects.scala#L14
It relies on `UnsafeRun` for the source effect and `Async` for target effect. This will change again when `UnsafeRun` is replaced by `Dispatcher`. Anyway it ties into the `LiftIO` question from my previous post. How will we tie effects together when `Effect` and `UnsafeRun` are gone?

SystemFw:
Ok so I think you need `Sync`, otherwise fs2 won't be able to `compile`, with `Sync` you will also get `Ref` for free (the instances for Ref work for Concurrent or Sync).

> In the last one uses the gadget I wrote to compensate for removal of AsyncEffect:


can you show a place where this is actually used?

SystemFw:
> Limiting the API surface area like that sounds great, but how that work with polymorphic effects aka tagless final style? LiftIO[ConnectionIO] allows converting concrete IO values into ConnectionIO. How do we convert an opaque F[_] to IO? What constraints on F would allow that?

This is a problem we encountered a few times now (ConnectionIO, ciris ConfigValue, and cats-effect-slick). There is no great answer here:
- you can either go for `ConnectionIO[F, A]`, which is out of the question

rkrzewski:
> > In the last one uses the gadget I wrote to compensate for removal of AsyncEffect:
> 
> can you show a place where this is actually used?

here https://github.com/rkrzewski/doobie/blob/cats-effect-3/modules/postgres/src/main/scala/doobie/postgres/syntax/FragmentSyntax.scala#L52
and here https://github.com/rkrzewski/doobie/blob/cats-effect-3/modules/hikari/src/test/scala/doobie/hikari/issue/824.scala#L53

SystemFw:
mm ok

@djspiewak we have an interesting counterexample of the `UnsafeRun` issue, I think we need it here

SystemFw:
>  https://github.com/rkrzewski/doobie/blob/cats-effect-3/modules/hikari/src/test/scala/doobie/hikari/issue/824.scala#L53

I think this one is covered much more simply by LiftIO

rkrzewski:
> > https://github.com/rkrzewski/doobie/blob/cats-effect-3/modules/hikari/src/test/scala/doobie/hikari/issue/824.scala#L53
> 
> I think this one is covered much more simply by LiftIO

Sure. How do I implement `LiftIO[ConnectionIO]` though? I need to ask for `LiftIO` for the target effect we are compiling into here: https://github.com/rkrzewski/doobie/blob/cats-effect-3/modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala#L71 right? Or is there another way?

rkrzewski:
@SystemFw a few weeks ago I was trying to write a piece of code that would pull data from a network socket and insert it into DB in batches. I ended up needing `Effect` constraint for the effect of the source stream, which is exactly the same situation as in `FragmentSyntax`
https://gist.github.com/rkrzewski/a85aaae0dc16b6b43923a08d7f62fdb0#file-streaming-scala-L50
I thought that I might be missing something, but it appears to be fundamental.

rkrzewski:
> > https://github.com/rkrzewski/doobie/blob/cats-effect-3/modules/hikari/src/test/scala/doobie/hikari/issue/824.scala#L53
> 
> I think this one is covered much more simply by LiftIO

I've added `LiftIO[ConnectionIO]` piggybacking on the `Transactor`'s target effect. The test case looks much cleaner now.

rkrzewski:
I'm trying to replace `Async` instance for `ConnectionIO` with `Sync` and `MonadCancel` according to earlier suggestion, but having them both in scope this leads to ambiguous `MonadError` instances which leads to pretty miserable end user experience. Any suggestions?

kubukoz:
Wouldn't you use `Sync[F] with MonadCancel[F, E]`? The only problem is that this doesn't look nice in context bounds. Maybe we should consider having a `With1[Sync, MonadCancel[*[_], E]]` in CE?

rkrzewski:
> Sync[F] with MonadCancel[F, E]

You mean that I could provide an implicit value of type `Sync[F] with MonadCancel[F, Throwable]` and Scalac would use it to satisfy `implicit F: Sync[F]` and `implicit F: MonadCancel[*[_], Throwable]` requirements? I vaguely recall there's compiler plugin that does something like that. I never tried that before. When does it / doesn't it work?

kubukoz:
> > Sync[F] with MonadCancel[F, E]
> 
> You mean that I could provide an implicit value of type `Sync[F] with MonadCancel[F, Throwable]` and Scalac would use it to satisfy `implicit F: Sync[F]` and `implicit F: MonadCancel[*[_], Throwable]` requirements? I vaguely recall there's compiler plugin that does something like that. I never tried that before. When does it / doesn't it work?

I mean that you would require `(implicit F: Sync[F] with MonadCancel[F, Throwable])`. As long as you have separate values for these, you'll have ambiguous `MonadError` instances.

Unfortunately I'm not sure all implementations of, say, Sync, will also implement MonadCancel in the same value. @djspiewak is that the case for monad transformers etc.?

rkrzewski:
I can easily provide `Sync[ConnectionIO] with MonadCancel[ConnectionIO, Throwable]` value because I'll wire all the methods via the free structure into the target effect `ConnectionIO` is compiled to, which was `Async` requirement.

I just tried it a particular use site that required `[F[_]: Async]` earlier and supplying it with separate `Sync` and `MonadCancel` values led to ambiguities. Changing the signature to `[F[_]](implicit F: Sync[F] with MonadCancel[F, Throwable])` resolved everything nicely.

I think I'm going to go for it. But it'll take a while, because updating 21 algebras is a _bit_ tedious :joy:

djspiewak:
Btw it should be relatively rare that something is both Sync and MonadCancel. Unfortunately, none of the inductive derivations take this situation into account, so if you do this it will be ambiguous within any transformer. That might be solvable with some shenanigans, but I'm not sure.

If this is a common scenario (as SyncIO will make it), then we probably need another typeclass under Sync which inherits both.

tpolecat:
@rkrzewski can you provide a sumary of the state of things right now? I think really all we can do here is `MonadError` (or, better, `MonadCancel`) since `LiftIO` isn't supported anymore.

rkrzewski:
@tpolecat I got distracted for a couple of days and I have some half cooked changes that I need to finish. 

@djspiewak 's comment above prompted me to try to remove the `Sync` instances and leave only `MonadCancel` ones. I came to the conclusion that it's possible for most of the algebras, but `ConnectionIO` and Postgres `LargeObjectIO` need `Sync`  to support certain streaming operations.

`Sync` and `MonadCancel` requirements usually appear separately, but here's a bit of code that expressly needs both:
https://github.com/tpolecat/doobie/blob/76324b1060de38337535b13e8441aa6de78ce1d4/modules/core/src/main/scala/doobie/util/io.scala#L23 

But I understand Daniel's remark correctly, it's the *providing* the implicit with compound type that is problematic, because it doesn't play well with transformers.

I'm still not sure if removing `Sync` instances from the algebras that don't need them (to the extent exercised by existing tests) is a good idea. I have the changes that remove them on a separate branch at the moment. Please advise.

Regarding  removal of `LiftIO`, I have a solution for the that allows lifting arbitrary `F[_]: Async` into `ConnectionIO` that is based on CE3 `Dispatcher`. It's more clunky on the use side (the shape is `Resource[F, F ~> ConnectionIO]`) but I believe it supports all the use cases of CE2 `Effect` in Doobie.
https://github.com/rkrzewski/doobie/blob/cats-effect-3/modules/core/src/main/scala/doobie/util/transactor.scala#L196-L207
It probably needs further tweaking, but at least the tests are passing.

rkrzewski:
OK, I've finished the changes I had in mind earlier. I think the PR is ready for review now. I plan to go over the changes myself and add comments in places that especially need attention IMO.

Some things to do / discuss / decide:
-  there to problems with tests: h2 tests consistently fail for me locally (also on the `master` branch) and the tests appear to hang in CI, but never hang for me locally. I haven't looked into the former and don't know how to debug the latter problem.
- should we keep `Sync` instances for algebras other that `ConnectionIO` and `LargeObjectIO`?
- what do we do about compound instances `Sync with MonadCancel`? A fix on `cats-effect` side appears to be necessary.
- there are API changes that need to be reflected in documentation.
- release notes / migration guide is needed

tpolecat:
> should we keep Sync instances for algebras other that ConnectionIO and LargeObjectIO?

Yes, all algebras should be uniform. Did you change them in-place? Everything in the `free` project is generated by `project/FreeGen2.scala`. I can make that change if you did them in-place.

rkrzewski:
LOL yes I did it all by hand. I wish I knew about this earlier! I think I can figure out this FreeGen2 thingy.

rkrzewski:
@tpolecat I've pulled the algebra changes into `FreeGent2.scala` but this uncovered the following problem, that I sidestepped by making changes by hand:
`ConnectionIO` does not have an CE3 `Async` instance because it doesn't support concurrency as discussed before. In order to implement lifting `IO` into `ConnectionIO` I needed to implement a part of `Async` interface, namely `async` method or it's equivalent. I've picked `fromFuture` because it was the easiest to hook into CE3 `Dispatcher`. In order to interpret that operation, `KliesliInterpreter` takes an `Async` constraint on it's target effect. So we have an asymmetry: we require `Async` but offer `Sync` and `MonadCancel` instances only. 
So far so good, but a problem arises in postgres module. It's `KleislitInterpreter` interprets into `ConnectionIO` (`doobie.postres.hi.connection.defaultInterpreter`) and since it's now generated from the same template as the core interpreter it requires `Async[ConnectionIO]` which is not available.
In the code written by hand, `ConnectionIO` was the only algebra of all with `fromFuture` operation, so postgres interpreter  could be implemented using `Sync` and `MonadCancel` constraints.
I guess I could modify the code of `FreeGen2.scala` to generate slightly different code for core and postgres interpreters. I would also need to remove `fromFuture` operation from postgres algebras, leaving it on the core ones. Would that be acceptable? What are the alternatives?

rkrzewski:
Other things I've noticed after tweaking `FreeGen2.scala`
-  some unnecessary imports were generated in the interpreters in the `// Types referenced in the JDBC API` section. I'm wondering if cleaning them up by hand is the usual practice, or is there some sort of regression?
- a couple new operations were added to postgres algebras. It seems the code was not regenerated after a recent postgres dependency update?
- I got a weird compilation error in tests:
```
[error] /home/rafal/workspace/scala/typelevel/doobie/modules/core/src/test/scala/doobie/util/strategy.scala:43:23: incompatible type in overriding
[error] def getTypeMap: cats.data.Kleisli[cats.effect.IO,java.sql.Connection,java.util.Map[String,Class[_]]] (defined in trait Visitor)
[error]   with override def getTypeMap: cats.data.Kleisli[cats.effect.IO,java.sql.Connection,java.util.Map[String,Class[_ <: Object]]] (defined in trait ConnectionInterpreter);
[error]  found   : cats.data.Kleisli[cats.effect.IO,java.sql.Connection,java.util.Map[String,Class[_ <: Object]]]
[error]  required: cats.data.Kleisli[cats.effect.IO,java.sql.Connection,java.util.Map[String,Class[_]]]
[error]     override lazy val ConnectionInterpreter = new ConnectionInterpreter {
```
I got around that by spelling out the return type in `KleisliInterpreter` by hand
```
override def getTypeMap: Kleisli[M, Connection, Map[String, Class[_]]] = primitive(_.getTypeMap)
```
Is this a known issue?

rkrzewski:
I've just had an idea for solving the instance asymmetry / algebra uniformity problem:
- Define a Doobie-specific typeclass in the `free` module called `WeakAsync` that would support all operations in `Sync`, `MonadCancel` and additionally `fromFuture` operation (I could also add other operations defined in `Async`). 
- Provide derivation of `WeakAsync` from `Async` and also derivations of `Sync` and `MonadCancel` from `WeakAsync`. I would prioritize the latter two using `LowPriorityImplicits` technique to avoid ambiguities at call sites that require `MonadError`. 
- Provide `WeakAsync` for all Doobie core and Postgres algebras, making the uniform.
- Change the constraint on `Transactor` from `Async` to `WeakAsync` which would make the core and Postgres interpreters uniform and allow piggy-backing one on the other.

I'm going to try it out later today :slightly_smiling_face:

rkrzewski:
Update: it worked really well.

modules/core/src/main/scala/doobie/util/transactor.scala
```diff
@@ -102,7 +101,7 @@ object transactor  {
     def strategy: Strategy
 
     /** Construct a [[Yolo]] for REPL experimentation. */
-    def yolo(implicit ev: MonadCancel[M, Throwable]): Yolo[M] = new Yolo(this)
+    def yolo(implicit ev: Async[M]): Yolo[M] = new Yolo(this)
```
hamnis, line 104: should this also be WeakAsync ?

hamnis: oh, never mind. I was looking at the commit this appeared in.

tpolecat:
I just merged doobie support which means this is going to have to be reworked a bit. But I can do it. Thanks for your work on this!

rkrzewski:
You mean Scala 3 support #1319? Oooh, exciting! I'll be happy to work on it, because I haven't hat a chance to play with Scala 3 yet.

rkrzewski:
I was able to get the code to compile and the tests to run on Scala 2.13 and Scala 3-M2.
There are compilation problems on 2.12 though. I didn't investigate them yet.

rkrzewski:
@tpolecat the latest CI build failed with an error in `paradoxMarkdownToHtml` task. I'm a bit stumped because it passes on my machine. Also, I don't think I touched anything related to documentation structure...

modules/docs/src/main/mdoc/docs/14-Managing-Connections.md
```diff
@@ -121,7 +121,7 @@ object HikariApp extends IOApp {
 And running this program gives us the desired result.
 
 ```scala mdoc
-HikariApp.main(Array())
+// HikariApp.main(Array()) // https://github.com/typelevel/cats-effect/issues/1560
```
djspiewak, line 124: Is this now resolved?

rkrzewski: I just did a quick test and unfortunately after uncommenting this line rendering the mdoc does not terminate.

from stack dump:
```
"run-main-0" #486 prio=5 os_prio=0 cpu=18472,29ms elapsed=1078,35s tid=0x00007f083403e800 nid=0x48714 waiting on condition  [0x00007f07c6ffa000]
   java.lang.Thread.State: TIMED_WAITING (parking)
        at jdk.internal.misc.Unsafe.park(java.base@11.0.8/Native Method)
        - parking to wait for  <0x00000004246a7340> (a java.util.concurrent.CountDownLatch$Sync)
        at java.util.concurrent.locks.LockSupport.parkNanos(java.base@11.0.8/LockSupport.java:234)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedNanos(java.base@11.0.8/AbstractQueuedSynchronizer.java:1079)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireSharedNanos(java.base@11.0.8/AbstractQueuedSynchronizer.java:1369)
        at java.util.concurrent.CountDownLatch.await(java.base@11.0.8/CountDownLatch.java:278)
        at cats.effect.IOPlatform.unsafeRunTimed(IOPlatform.scala:40)
        at cats.effect.IOPlatform.unsafeRunSync(IOPlatform.scala:27)
        at repl.MdocSession$App.<init>(15-Extensions-PostgreSQL.md:78)
        at repl.MdocSession$.app(15-Extensions-PostgreSQL.md:3)
        at mdoc.internal.document.DocumentBuilder$$doc$.$anonfun$build$2(DocumentBuilder.scala:89)
```
No other thread appears to be holding `0x00000004282929d8`

rkrzewski: Wait, I just noticed that the file name in the stack trace is wrong? An also line numbers! lines 3 and 78 in `15-Extensions-PostgreSQL.md` do not contain runable code and I'm positive that it's the line 124 in `14-Managing-Connections.md` that makes the rendering hang or not.

rkrzewski:
build failed because some dependencies are missing for Scala 3.0.0-M2. Tried switching from M2 + M3 to M3 + RC1 but unfortunately _other_ dependencies are missing for RC1.

kubukoz:
Any chance we could merge this somewhere and make a release?

tpolecat:
> Any chance we could merge this somewhere and make a milestone release?

I'll try to do this on Friday. I need to review it closely, which I haven't done.

kubukoz:
Sounds good to me ;) thanks

rkrzewski:
@tpolecat If you have questions / requests for changes let me know.

kubukoz:
Music to my ears :D

bplommer:
Might this be a good time to join the rest of the world in calling type constructor parameters `F[_]` rather than `M[_]`?

modules/docs/src/main/mdoc/docs/03-Connecting.md
```diff
@@ -32,18 +32,17 @@ This is a perfectly respectable **doobie** program, but we can't run it as-is; w
 ```scala mdoc:silent
 import doobie.util.ExecutionContexts
 
-// We need a ContextShift[IO] before we can construct a Transactor[IO]. The passed ExecutionContext
-// is where nonblocking operations will be executed. For testing here we're using a synchronous EC.
-implicit val cs = IO.contextShift(ExecutionContexts.synchronous)
+// This is just for testing. Consider using cats.effect.IOApp instead of calling calling
```
CremboC, line 35: This file an bunch of other `.md` files with the same comment have a typo `calling calling`

modules/core/src/test/scala/doobie/util/StrategySuite.scala
```diff
@@ -45,18 +41,18 @@ class StrategySuite extends munit.FunSuite {
     }
 
     override lazy val ConnectionInterpreter = new ConnectionInterpreter {
-      override val close = delay(() => Connection.close = Some(())) *> super.close
-      override val rollback = delay(() => Connection.rollback = Some(())) *> super.rollback
-      override val commit = delay(() => Connection.commit = Some(())) *> super.commit
-      override def setAutoCommit(b: Boolean) = delay(() => Connection.autoCommit = Option(b)) *> super.setAutoCommit(b)
+      override val close = delay(Connection.close = Some(())) *> super.close
```
CremboC, line 44: Still feel like we shouldn't be even using `*>` with IO and always prefering `>>`, see https://gitter.im/typelevel/cats-effect?at=6062ef06eb800601df0adf70

modules/free/src/main/scala/doobie/WeakAsync.scala
```diff
@@ -0,0 +1,57 @@
...
+import cats.effect.implicits._
+import cats.effect.std.Dispatcher
+import scala.concurrent.Future
+import scala.concurrent.duration.FiniteDuration
+
+trait WeakAsync[F[_]] extends Sync[F] {
+  def fromFuture[A](fut: F[Future[A]]): F[A]
+}
+
+object WeakAsync {
```
CremboC, line 19: Would it be worth to put a comment in the code explaining why `WeakAsync` is necessary? From what I can see it's `Sync` but with `fromFuture` and nothing else. Why not just use `Async` directly?

modules/core/src/test/scala/doobie/util/StrategySuite.scala
tpolecat: They're identical up to termination. I always use the weaker operator unless I know I need the stronger one.

CremboC:
I'm not really maintainer of doobie but thought another pair of eyes through the code would be useful, so put some comments in.

One more thing that I noticed that has nothing to do with PR but in general doobie is the amount of repetition we have to do with `free` module - lots of files that look 99% similar. Seems like the large diff of this PR is mainly to do with update those files.

modules/free/src/main/scala/doobie/WeakAsync.scala
rkrzewski: Absolutely. The short explanation is that `Async` cannot be implemented for `ConnectionIO` because it implies concurrency. You cannot concurrently access a `Connection` or a `ResultSet` from different fibers / threads.

One the other hand `Sync` is not sufficient, because it would preclude streaming data produced with `IO` effects into the database. In order to do that we need a `FunctionK` from `IO` to `ConnectionIO`. Cats Effect 3 `Dispatcher` can be uses do that, but only if `ConnectionIO` can absorb a value that is produced asynchronously. Hence `WeakAsync.fromFuture`.

See also https://github.com/rkrzewski/doobie/blob/cats-effect-3/modules/free/src/main/scala/doobie/WeakAsync.scala#L45

rkrzewski:
@CremboC 
> Seems like the large diff of this PR is mainly to do with update those files.

Those files are generated, see: https://github.com/rkrzewski/doobie/blob/cats-effect-3/project/FreeGen2.scala

CremboC:
Ahhh sorry, that makes sense, should have checked it myself. 😅  Thanks!

rkrzewski:
No worries, I've learned this the hard way, editing 20+ files by hand first...

modules/docs/src/main/mdoc/docs/03-Connecting.md
rkrzewski: Good catch. Fixed.

Gonzih:
I assume that doing sensible review of such massive change is problematic. So based on this, would it be possible to merge those changes and put them out as RC release and work from there based on feedback from people trying out RC release on their codebases?

modules/core/src/test/scala/doobie/util/StrategySuite.scala
kubukoz: Yeah, Fabio (in an answer before the message you linked to) also said to use it when recursion is involved, which isn't always the case ;)

tpolecat:
N.B. I have retargeted this PR for `series/1.0`.

tpolecat:
**THANK YOU** for your hard work on this, and to everyone who has commented. I have some other things I need to get into master and synced up here but I'll try to get a milestone out today or tomorrow. This is an outstanding contribution.

kubukoz:
@tpolecat wanna do an organization name change now that we're at it? wink wink

PR #1281, opened 2020-10-07
fixes NullPointer if nested exception has null msg

HlibRuk:
I got an error `SQLFeatureNotSupportedException` in a test, but it got covered by NullPointerException. This commit should fix it.


HlibRuk:
Also, when I tried to run `sbt core/test` I got 3 tests failed:
 x allow margin stripping
 x allow margin stripping with custom margin
 x not affect margin characters in middle outside of margin position
Stayed that way even after I reverted my changes. But that can be just that I ran them on windows/cygwin environment.

jatcwang (approved):

Daenyth (approved):

Daenyth:
@HlibRuk Thank you for the ☼contribution☼ ;)

Issue #1288, opened 2020-10-13
Book Of Doobie - Text doesn't correspond to code example


LLCampos:
In [here](https://tpolecat.github.io/doobie/docs/04-Selecting.html#reading-rows-into-collections), there's the following snippet of code:

```scala
sql"select name from country"
  .query[String]    // Query0[String]
  .to[List]         // ConnectionIO[List[String]]
  .transact(xa)     // IO[List[String]]
  .unsafeRunSync    // List[String]
  .take(5)          // List[String]
  .foreach(println) // Unit
// Afghanistan
// Netherlands
// Netherlands Antilles
// Albania
// Algeria
```

But the text that follows does **not** correspond to this example, e.g., it mentions `.unique`, which is not called in the snippet.

jatcwang:
Hi thanks for reporting an issue, but I think in this case it's intentional. You can see that mention of `unique` etc follows "Similar methods are:"

Perhaps it'd be clearer if those were sublist items? Would you like to make a PR to help make it clearer? :)

LLCampos:
Ok, checked this out and it seems that in [the markdown file](https://github.com/tpolecat/doobie/blob/f04a7a3cab5aecb50be0d1ad10fbdae6b8db5ec2/modules/docs/src/main/mdoc/docs/04-Selecting.md#reading-rows-into-collections), those are sublist items indeed!

I guess the issue then is with the docs generation (through [mdoc](https://github.com/scalameta/mdoc))?

jatcwang:
Possibly a bug in Paradox. I'm using mdoc for my other [project documentation](https://jatcwang.github.io/doobieroll/docs/assembler) and sublists work fine there.

LLCampos:
Yes, it's a bug in Paradox. Already raised [an issue](https://github.com/lightbend/paradox/issues/452) on its repo, so I guess this one can be closed.

jatcwang:
Thank you @LLCampos  :)

PR #1293, opened 2020-10-20
docs - use 4 spaces instead of 2 for sub-lists

LLCampos:
As reported in https://github.com/tpolecat/doobie/issues/1288, sub-lists are not being shown as expected in Doobie documentation. This is because when using paradox, [4 spaces or a tab of indentation are needed for sub-lists to be rendered as expected](https://github.com/lightbend/paradox/issues/452#issuecomment-711138155), but Doobie docs are using 2 spaces indentation for sub-lists.

This PR fixes that. I've build the docs locally and confirmed that the sub-lists now are shown as expected.


Daenyth (approved):
Might be worth setting up a markdownlint config file later

Daenyth:
@LLCampos Thank you very much for the contribution!

PR #1310, opened 2020-11-21
`Read` can be `Applicative`, not just `Apply`

stephenjudkins:


stephenjudkins:
Hmm. Looks like failure is due to issue fixed in https://github.com/tpolecat/doobie/pull/1312

tpolecat (approved):
Thanks! Will merge after https://github.com/tpolecat/doobie/pull/1312/files

PR #1343, opened 2021-01-08
Replace Shapeless with Magnolia

oyvindberg:
Hi there

I'm posting this early to see if you can see a path forward with this.

Typeclass derivation with shapeless is terribly slow, and magnolia speeds it up significantly. It also has significantly better error messages. 

As you might know there are a few drawbacks as well.
- Magnolia doesn't currently support unary products (AFAIK)
- Magnolia isn't shapeless, so it doesn't have HLists.

### Where can we go further with this?

- If we manage to extract derivation out of core, we can support all of magnolia (possibly for both scala 2 and 3 if and when the port is completed), shapeless and scala 3 derivation
- Ideally we can have auto and semi-auto derivation.

### Existing usage

This code is derived from code which has been in production for a year (though it is rewritten and hence should be tested well).
We estimate that it has cut compilation time by almost half for this particular project, which has heavy doobie usage.


oyvindberg:
This PR has the same problem with optional things as #1344 showed, but I'll fix that soon.

Still interested in feedback on this :)

jatcwang (approved):
LGTM! There are still a few references to HList / shapeless in the docs / scala docs if you can reword or remove them :) 

I'm also wondering whether we should add support for HLists instances as well if it doesn't already work for compatibility reasons. Not all queries are done on types that are statically available (e.g. imagine some generic graphql query engine). Can be done separately though

propensive:
What requirements did you have in mind for unary products? I'm not aware of any limitations in that area, but possibly I just don't know what Doobie needs...

oyvindberg:
Thanks @jatcwang . I'll spend a bit more time on this on friday and will update the docs then

@propensive I'm not currently aware of a way to use magnolia to limit a product to be unary, case in point:
```scala
  /** @group Instances */
  implicit def unaryProductPut[A: TypeTag, L <: HList, H, T <: HList](
     implicit G: Generic.Aux[A, L],
              C: IsHCons.Aux[L, H, T],
              H: Lazy[Put[H]],
              E: (H :: HNil) =:= L
  ): Put[A] = {
    void(E) // E is a necessary constraint but isn't used directly
    H.value.contramap[A](a => G.to(a).head)
  }
```

Have I missed something?


So about this change in general, I feel we should really try to separate derivation out of core. I'll see if it seems possible on friday as well. That way we can mimic what circe does, and I think that has been pretty great, including that we use [circe-magnolia](https://github.com/circe/circe-magnolia)

propensive:
This probably wouldn't help, but would it be possible for you to implement two `gen` methods, one with a `<: Product` constraint on the type parameter, and one without, which both call the macro? One would probably need to go into a supertrait to give it lower priority...

But I think you're probably right: I think there's no way to _statically_ disambiguate between a one-parameter case class and a multi-parameter case class. If the return type needs to be different in each case, you might be out of luck. If you don't need different types, then a runtime check in the `combine` method would be sufficient (but I guess you've already explored that avenue...)

oyvindberg:
Yeah, in doobie there are different type classes for this. For instance (following my understanding) `Put` describes how to take a (non-composite) value and apply it to a cell in a `PreparedStatement` or `ResultSet`, while `Write` describes how to do the same for for a whole row of data. As you can imagine it was easy to write type class derivation for the latter, but I'm stuck with the former.

oyvindberg:
also @propensive thank you very much for magnolia. I'm using it (close to) everywhere now and I'm incredibly happy with it :1st_place_medal:

tpolecat:
👍🏻 for `magnoobie`

tpolecat:
Tagged `v0.11.0-M1` … I want to give people a chance to try it out before a final release. Thanks!

propensive:
Absolutely. This change to derivation might need a lot of testing with downstream libraries, and there are enough weird implicit resolution scenarios that the only real way to test it is for most users to try it for themselves. So I'd take some time over this step...

oyvindberg:
Thanks for merging @tpolecat :fireworks: 

So yes, like @propensive said, let's definitely give users a bit of time to report back how it works.

Like I said in the PR message this code is rewritten/refactored from what has been battle-tested, so we also need to confirm that everything that compiles still works. I'll do my part and spend some time on Friday to upgrade a new $WORK codebase, and my colleagues will upgrade the codebase I mentioned initially.

We also need to document the change properly, including that users now need to manually create instances for wrapper types.

tpolecat:
The issues with unary products are preventing us from moving forward here, so I'm going to revert this for now and maybe we can revisit if it gets sorted out in Magnolia.

PR #1344, opened 2021-01-08
Simplify implicits for Option instance of Read and Write

hamnis:
We do not need to have complex shapeless / scala 3 derivations for Option instances


hamnis:
hm, i dont understand why the docs build fails.

jatcwang:
```
 error: /home/runner/work/doobie/doobie/modules/docs/src/main/mdoc/docs/18-FAQ.md:137:1: SQL `NULL` read at column 3 (JDBC type Char) but mapping is to a non-Option type; use Option here. Note that JDBC column indexing is 1-based.
{
^
doobie.util.invariant$NonNullableColumnRead: SQL `NULL` read at column 3 (JDBC type Char) but mapping is to a non-Option type; use Option here. Note that JDBC column indexing is 1-based.
```

Random guess: the derivation you wrote may not handle Option of product types (case class)?

hamnis:
this works 
using types in `ReadSuite`
`util.Read[Option[(Int, LenStr1, Option[(LenStr1, String, LenStr2)])]]`

hamnis:
Eck. I guess shapeless generates this for Option as well, so will investigate this a bit more

hamnis:
dammnit, this does not work. Closing.

PR #1350, opened 2021-01-21
Reintroduce automatic derivation of `Get` and `Put` 

oyvindberg:
Followup from #1343 .

I made a proposal in https://github.com/propensive/magnolia/pull/278 , and this implements it on the doobie side. 

I fully expect some discussion about what it'll look like in the end, just opening this in a draft state this to show that it might be easy to fix.


oyvindberg:
This is a concrete proposal now that doesn't change anything in magnolia anymore. I'd still wait a bit on feedback on the PR I opened there as this could still be better solved on that side I think.

It's obviously not entirely optimal to define local macro in doobie, but it's only for scala 2, and it's not that complex I think.


An alternative to this solution is to resurrect the shapeless dependency and use that only to infer `Get` and `Put.

modules/core/src/test/scala/doobie/util/PutSuite.scala
```diff
@@ -9,6 +9,8 @@ import doobie._
 import scala.concurrent.ExecutionContext
 
 class PutSuite extends munit.FunSuite with PutSuitePlatform {
+  case class X(x: Int)
+  case class Q(x: String)
```
jatcwang, line 13: Can we add tests for ones with private constructors as well :)

oyvindberg: added a few more tests now

modules/core/src/test/scala/doobie/util/GetSuite.scala
```diff
@@ -13,18 +13,32 @@ import scala.concurrent.ExecutionContext
...
 class GetSuite extends munit.FunSuite with GetSuitePlatform {
 
-  case class Z(i: Int, s: String)
-  object S
+  case class Yes1(x: Int)
+  case class Yes2(x: Yes1)
+
+  case class No1(i: Int, s: String)
+  object No2
+  case class Y private (x: String)
```
jatcwang, line 21: This isn't used (and it shouldn't be possible, since constructor is private)

tpolecat:
I'm losing confidence that this is the way to go. For now I'm going to revert #1343 and maybe we can revisit if it's fixed in Magnolia.

PR #1351, opened 2021-01-23
Remove specialized option implicits and a `Write#toList` problem

oyvindberg:
So I'm going over the magnolia changes one more time. 

There are two specialized implicits for `Option` I now think are unnecessary:
- `Read.fromGetOption`
- `Write.fromPutOption`

When removing the latter, `FragmentSuite` started failing because of a `ClassCastException` in the values returned fom `Write#toList` ("String can not be cast to Option"). It's typed as `List[Any]` so it's not entirely obvious to me what should go in there. 

Anyways something is (and was) wrong, because given a `p: Put[Foo]`, `Write.opt(Write.fromPut(p))` is not the same as `Write.fromPutOption(p)`

I tried patching `toList` in WriteLowerPriorityImplicits#opt to something I think is more correct (and it fixes the test)

Does this look good?

The PR also avoids some closures where we can just pass existing functions


szymon-rd:
It would be great to merge it soon, as it also fixes an error with a match error in Put for optional column vectors.

modules/core/src/main/scala/doobie/util/write.scala
```diff
@@ -98,29 +98,22 @@ object Write extends WritePlatform with WriteLowerPriorityImplicits {
...
 
 sealed trait WriteLowerPriorityImplicits {
   implicit def opt[A](implicit A: Write[A]): Write[Option[A]] = new Write[Option[A]](
     puts = A.puts.map { case (p, _) => (p, Nullable) },
-    toList = oa => oa.map(A.toList).getOrElse(A.puts.map(_ => null)),
-    unsafeSet = (ps, i, a) => A.unsafeSetOption(ps, i, a),
-    unsafeUpdate = (ps, i, a) => A.unsafeUpdateOption(ps, i, a),
+    toList = {
+      case None => A.puts.map(_ => None)
+      case Some(a) => A.toList(a).map(Some(_))
```
nigredo-tori, line 113: Hm. So if we have an `Option` field in a case class, and we wrap that case class in another `Option`, then we'll have `Some(Some(...))` here (possibly up to unlimited depth)? I'm not sure if this is correct.

nigredo-tori: We have a pattern-match [here](https://github.com/tpolecat/doobie/blob/d7c42833cf5aa3d8c26716d33ac5f76c1a5c531b/modules/core/src/main/scala/doobie/util/write.scala#L69-L70) that provides some evidence as to what the contract of `toList` is (a raw type element for each `NoNulls` column, an `Option` for each `Nullable`). Based on what happens to `puts` above, I'd say we want something like this:
```suggestion
      case Some(a) =>
        A.puts.zip(A.toList(a)).map {
          case ((_, NoNulls), x) => Some(x)
          case ((_, Nullable), ox) => ox
        }
```

oyvindberg: > Hm. So if we have an Option field in a case class, and we wrap that case class in another Option, then we'll have Some(Some(...)) here (possibly up to unlimited depth)

Yes, this exactly was my understanding when I wrote this code. 


Thanks for the link and the suggestion. Based on the pattern match you linked I think I convinced myself that it's correct. so I'll accept it and see the tests agree :)

Of course it would be preferable to capture this relationship in a type, and/or to produce some edge case tests and documentation for this thing. Any ideas in that direction @nigredo-tori ?

nigredo-tori: Regarding capturing the relationship in a type - there might a way to capture that with `HList`s, but it won't be easy, and it will certainly require an overhaul of `Write`.

Regarding tests - I think that adding a type like this
```scala
class Foo(
  int: Int,
  intOpt: Option[Int]
)
```
, writing `Foo`, `Option[Foo]`, `Option[Option[Foo]]` to the database and reading them back would be a good enough test for this particular instance. For a more general approach, we can `zip` `puts`  from an arbitrary `Write` instance with its `toList` output, and check that we have `Option`s for (and only for) `Nullable` columns, and that we don't have nested options.

jatcwang:
Thanks @oyvindberg. This looks fine to me. Can you please add a test or two (maybe in modules/postgres/src/test/scala/doobie/postgres/TypesSuite.scala) to cover the case you've fixed?

oyvindberg:
Just to report back @jatcwang , I just sat down to write tests to cover all variants of optionality I could come up with (more or less what @nigredo-tori suggested), and we're not passing, before or after this PR. 

On the bright side this will be a very useful set of tests to have, since it seems to be completely uncovered for now.

Code and analysis soon, but I reached bedtime before I could wrap anything up. I'll also compare with latest release.

tpolecat:
This seems related to the magnolia change (which is on hold for now) so I'm closing this. Feel free to re-open if this is in error.

PR #1358, opened 2021-01-28
docs: Clarify wording regarding ContextShift in IOApp

pektinasen:
I think the sentence makes it sound like IOApp implements a ContextShift based off of ExecutionContext.global. Which is not the case: [PoolUtils.scala](https://github.com/typelevel/cats-effect/blob/8ae7479753d6673bf261e797c073bcfd54dd3050/core/jvm/src/main/scala/cats/effect/internals/PoolUtils.scala#L28)

Not sure if this new wording expresses the same what you initially intended to say, though.


jatcwang (approved):
I think this is an improvement & more accurate. Thanks :)

PR #1379, opened 2021-02-20
Make Update#apply take logHandler implicitly

jozic:
This is to follow up a short conversation in gittter
The issue (or rather usability issue) I'm trying to solve:

II have a need to use `updateMany` which is available on `Update` which when constructed using `apply` is the only place (at least in my code) that requires an explicit `LogHandler` (as well as `Option[Pos]`). And because there's a default value set to `LogHandler. nop` available in scope implicit `LogHandler` is silently ignored


jozic:
@jatcwang Do you mind checking this PR?

jatcwang (approved):
:+1: for this but normally it's up to Rob to merge the code changes.

PR #1389, opened 2021-03-08
Revert magnolia

tpolecat:
This reverts PR #1353 which swapped out Shapeless for Magnolia. Lingering issues with unary products are blocking further development so I'm rolling this back for now. See discussions in #1343  and #1350 for context. 

Other simplifications introduced in #1343 prior to the Shapeless/Magnolia swap are also removed because one of the tests doesn't compile in 2.12, but I would consider a PR re-introducing them.


oyvindberg:
Alright, that's a shame, but I think a good decision for now - let's keep the Scala 3 release train running instead. Meanwhile I'll chase the necessary changes upstream :+1:

Issue #1400, opened 2021-03-23
Select a list of values without a table


dmitriibundin:
Is there an existing functionality to select a `List` of some values without a table in order to join it with some existing table?

Example:
```
def executeQuery(ids: List[Int]) = 
   sql"select * from (values $ids) s(a) join some_tbl on a=some_tbl.id"
```

jatcwang:
`doobie.util.fragments#in` is probably what you're looking for?
Note that in SQL `IN` requires at least one value, so you have to use NonEmptyList or any other container with a `Reducible` typeclass instance

dmitriibundin:
@jatcwang Thanks for the prompt reply. Unfortunately `doobie.util.fragments#in` is not exactly what I'm looking for. Would it make sense to add something like the following:
 
    def values[F[_]: Reducible, A: util.Put](f: Fragment, fs: F[A]): Fragment =
      fs.toList.map(parentheses).foldSmash1(f ++ fr0"VALUES (", fr",", fr")")

?

jatcwang:
@dmitriibundin Sorry I misread the original example. Yes I think it's a good idea - we can extend it to support tuples e.g.
`SELECT * FROM (VALUES (?,?),(?,?))`.

https://www.postgresql.org/docs/9.5/queries-values.html

Would you like to raise a PR? :)

PR #1401, opened 2021-03-23
Add tableValues fragment #1400

dmitriibundin:


modules/core/src/main/scala/doobie/util/fragments.scala
```diff
@@ -12,6 +12,14 @@ import cats.syntax.all._
 /** Module of `Fragment` constructors. */
 object fragments {
 
+  /** Returns `VALUES (fs0, fs1, ...)`. */
+  def tableValues[F[_]: Reducible, A: util.Put](fs: F[A]): Fragment =
```
jatcwang, line 16: * If you use `(implicit w: Write[A])` instead and call `w.toFragment`, you won't need the method below. This will work for single column values & product types too.
* I might be wrong but the extra paren after `VALUES` seems unnecessary. `SELECT * FROM (VALUES (3),(2)) a;` seems to work fine for me in Postgres. (MySQL has very different syntax for table values. @tpolecat do we tend to only put standard SQL in this file?)
* Is there an issue calling this `values` following the the names of other functions in this file?

dmitriibundin: - I agree that `Write` can be derived for tuple/product types, but is it possible to derive it for variable-length data structures like `NonEmptyList` as well which is what my initial point was about?
- Agree, will fix.

jatcwang: No, Write doesn't get derived for List/NEL so it's fine.

dmitriibundin: So why don't we need the method? With the current implementation it's possible to write the following:

```
def query(ids: NonEmptyList[Int]) = 
    sql"select * from (${fragments.values(ids)}) s(a) join some_tbl on a=some_tbl.id"
```

`fragments.values` here provides the `VALUES (?), (?), (?)` which is supposed to be written by hand otherwise.

dmitriibundin: Hi, @jatcwang . Are there any updates from your side?

jatcwang: @dmitriibundin Hi sry missed your msg. I'll take a look later

jatcwang: Looking through the thread it seems like there might be some miscommunication of what I meant 
and why I suggetsted using `Write`.

Basically I'm trying to generalize this to allow any F[A] as long as `A` has a `Write` instance

```
case clsas MyObj(a: Int, s: String, d: Double)

val myObjs: NonEmptyList[MyObj] = ...

sql"SELECT * from (${fragments.values(myObj)}) s(a) join ..."
```

So the signature of `values` that I'm suggesting is this

```
def values[F[_]: Reducible, A](fs: F[A])(implicit w: Write[A]): Fragment = ...
```

This will work whether `A` is `Int` (single column) or a case class / tuple. It's more general :)

Implementation-wise you can use `w.toFragment`. Let me know if you need more help.

dmitriibundin: Hi @jatcwang, thanks for the detailed clarification. Updated the PR according to your latest comment.

jatcwang (approved):

PR #1418, opened 2021-04-11
Typo

gruggiero:
Must be a long connection with three 'n' :)


jatcwang:
Yeah long connections are bad. We don't like timeouts ;) Thanks

PR #1445, opened 2021-05-13
Add Discord badge

paualarco:
Making the Discord support channel more visible by adding a badge with a badge to #doobie that never expires :)

This is how it looks like:

Screen Shot 2021-05-14 at 00 25 25


tpolecat:
Thanks!

PR #1451, opened 2021-05-15
Scala 3.0.0 for 1.x series (CE3)

jatcwang:


build.sbt
```diff
@@ -80,8 +79,8 @@ lazy val commonSettings =
 
     // MUnit
     libraryDependencies ++= Seq(
-      "org.typelevel"     %% "scalacheck-effect-munit" % "1.0.1"  % Test,
-      "org.typelevel"     %% "munit-cats-effect-2"     % "1.0.2" % Test,
+      "org.typelevel"     %% "scalacheck-effect-munit" % "1.0.2"  % Test,
+      "org.typelevel"     %% "munit-cats-effect-2"     % "1.0.3" % Test,
```
jatcwang, line 83: @tpolecat should this be munit-cats-effect-3 instead? :thinking:

tpolecat: Yes, curious that it works at all.

tpolecat: We're good to go if you can fix this.

ahjohannessen:
:shipit:

PR #1456, opened 2021-05-18
Avoid narrowing the return type of the Quill probe() method

guilgaly:
**Context:** We're using Quill with Doobie and want to use Quill's query probing functionality. `probe()` is not implemented in Doobie's Quill context, so we tried overriding it with an implementation that would work for our app.

**Issue:** when overriding `probe()`, we can never return a `Failure` because its return type gets narrowed down to `Success[Unit]` through type inference.

**Proposed fix:** with the explicit return type, even if it's not implemented out of the box, the user can override it in a way that works for them.


guilgaly:
For what it's worth, our implementation looks something like this:

```scala
object MyContext extends DoobieContext.Postgres(Literal) with QueryProbing {

  // this datasource doesn't rely on Doobie at all, and is only used for query probing
  private lazy val dataSource = JdbcContextConfig(LoadConfig("ctx")).dataSource
  
  private val queryProbing: Boolean = ??? // we enable this with an env variable (not shown here to simplify) because we don't always want to compile with query probing enabled; plus it should be disabled at runtime to avoid instantiating the above datasource

  override def probe(sql: String): Try[Unit] =
    if (queryProbing) {
      Using(dataSource.getConnection) { connection =>
        connection.createStatement.execute(sql)
      }
    } else {
      Success(())
    }

  override def close(): Unit =
    if (queryProbing) {
      dataSource.close()
    }
}
```

tpetillot (approved):

PR #1457, opened 2021-05-19
Use HikariCP 4

hamnis:
This is an interesting one.
Hikari depends on slf4j-api 1.7 according to the POM, unless you
take profiles into account, which apparently coursier does.

Hikari changes the slf4j-api dependency to 2.0-alphaN when
you are running Java 11 and above, this also is reflected when you
resolve using Coursier in Java 11 or Java 8.
The only reason to do this is to use Java module support.

Most Scala libraries do not care about Java Modules, so we can safely
ignore this.

Added a dependency to Hikari4, but adds an exclusion to slf4j-api.
Added a direct dependency to slf4j-api 1.7 to fullfill the contract.


tpolecat:
Huh, ok. Thanks for figuring this out!

PR #1463, opened 2021-05-24
add more helper cases for in and notIn

massimosiani:
Hi,
thanks for this project!

I have a couple of helper methods in my codebase, maybe they can help someone else.

I'm not sure how they can be generalized to arbitrary length.


tpolecat:
You can interpolate `Write` now (since #1519, not yet published) which would allow you to generalize this to any arity. Closing for now, feel free to re-open if you'd like to make that change.

PR #1483, opened 2021-06-29
Update 12-Custom-Mappings.md

yisraelU:
added clarification for pulling into scope java.time meta instances


jatcwang:
Thanks @yisraelU!

PR #1485, opened 2021-07-01
Fix order of gets in `Read.ap`

mdmoss:
I'm a little uncertain about this, but I think the order of gets isn't right in `Read.ap`.

I've been constructing some `Read` instances by hand (using `.product`) to handle a table with both `json` and `jsonb` columns. The resulting query has the expected type and runs successfully, but fails checks. The values in the check output are reversed compared to what's expected (see example below).

After a little playing around, I've found a change that seems to fix the issue - queries check and work as expected. I've spent some time trying to follow how the gets are used in analysis, but I couldn't find an obvious reason why this is happening. I normally don't like to PR without understanding what's actually going on, but here we are.

Using the world database (from https://tpolecat.github.io/doobie/docs/01-Introduction.html):

```
def selectAllWithReadFromProduct = {
  def readTuples: Read[(String, (Int, (Option[BigDecimal])))] = {
    doobie.Read.fromGet(implicitly[doobie.Get[String]]).product(
      doobie.Read.fromGet(implicitly[doobie.Get[Int]]).product(
        doobie.Read.fromGetOption(implicitly[doobie.Get[BigDecimal]])
      )
    )
  }
  sql"""select name, population, gnp from country""".query[(String, (Int, Option[BigDecimal]))]
}

// Fails, with column order reversed - see output below
selectAllWithReadFromProduct.check.unsafeRunSync()

// Works as expected - prints "239"
selectAllWithReadFromProduct.to[Vector].map(_.length).quick.unsafeRunSync()
```

The output from `selectAllWithReadFromProduct.check.unsafeRunSync()` is:

```
Query0[(String, (Int, Option[BigDecimal]))] defined at <console>:46
  select name, population, gnp from country
  ✓ SQL Compiles and TypeChecks
  ✕ C01 name       VARCHAR (varchar) NOT NULL  →  Option[BigDecimal]
    VARCHAR (varchar) is ostensibly coercible to Option[BigDecimal]
    according to the JDBC specification but is not a recommended
    target type. Expected schema type was DECIMAL or NUMERIC.
  ✓ C02 population INTEGER (int4)    NOT NULL  →  Int
  ✕ C03 gnp        NUMERIC (numeric) NULL      →  String
    NUMERIC (numeric) is ostensibly coercible to String according to
    the JDBC specification but is not a recommended target type.
    Expected schema type was CHAR or VARCHAR or LONGVARCHAR or NCHAR
    or NVARCHAR or LONGNVARCHAR.
    Reading a NULL value into String will result in a runtime failure.
    Fix this by making the schema type NOT NULL or by changing the
    Scala type to Option[String]
```

Swapping the order of gets in the new `Read` instance from `Read.ap` fixes the issue:

```
implicit class ReadFixes[A](base: Read[A]) {
  def apFixed[B](ff: Read[A => B]): Read[B] = {
    // Change here - current `master` has base.gets + ff.gets for the first argument
    new Read(ff.gets ++ base.gets, (rs, n) => ff.unsafeGet(rs, n)(base.unsafeGet(rs, n + ff.length)))
  }
  
  // Same as `.product` from cats, except spaced out a little and using `.apFixed`
  def productFixed[B](f: Read[B]): Read[(A, B)] = {
    val ff: Read[B => (A, B)] = base.map(a => (b: B) => (a, b))
    f.apFixed(ff)
  }
}

def selectAllWithReadFromFixedProduct = {
  def readTuples: Read[(String, (Int, (Option[BigDecimal])))] = {
    doobie.Read.fromGet(implicitly[doobie.Get[String]]).productFixed(
      doobie.Read.fromGet(implicitly[doobie.Get[Int]]).productFixed(
        doobie.Read.fromGetOption(implicitly[doobie.Get[BigDecimal]])
      )
    )
  }

  sql"""select name, population, gnp from country""".query[(String, (Int, Option[BigDecimal]))]
}

// Works as expected - check passes.
selectAllWithReadFromFixedProduct.check.unsafeRunSync()

// Works as expected - prints "239".
selectAllWithReadFromFixedProduct.to[Vector].map(_.length).quick.unsafeRunSync()
```

(Here's the equivalent with everything implicit, just for reference).
```
def selectAll = {
  sql"""select name, population, gnp from country""".query[(String, (Int, Option[BigDecimal]))]
}

// Works as expected - check passes.
selectAll.check.unsafeRunSync()

// Works as expected - prints "239".
selectAll.to[Vector].map(_.length).quick.unsafeRunSync()
```


jatcwang:
Thanks for the report. I don't have my head wrapper around the `ap` implementation yet. I'm thinking about implementing `product` first (which is much easier to think about) and then implement `ap` in terms of it. Going along that line, the fact that we're using `ff.length` seems off to me.

Would appreciate the help if you can play around with my suggestion and see if that solves the problem. Otherwise I'll look into it.

mdmoss:
> using `ff.length` seems off to me - shouldn't we be using `base.length` instead since that's the offset we should be using to fetch the columns for `B`

That seemed off to me as well, but after trying a few things all I achieved was to make running the query throw exceptions.

Thinking aloud here, when `A.product(B)` is called, we actually end up with `B.ap(B => (A, B))`. It makes sense, then, that we want to `unsafeGet` the argument that's passed to `.ap` with the lower `n` (because it's the first of the two arguments to .`product`, and because it needs to get the A from somewhere first before we get the B). The eventual execution gets the `A` from `ff` first, and then gets the `B` from `base`.

I think I've convinced myself - when the Read is used, the `gets` from the argument to `.ap` should come first, because they're actually for earlier columns. The call to `A.product(B)` ends up calling `B.ap(B => (A, B))`, so for everything to end up in the right order we need `ff.gets ++ base.gets`, which gives you `A.gets ++ B.gets`.

jatcwang:
I can reproduce the issue @mdmoss. I had a play around with defining `ap` using `product` and it seems to fix the issue as well. I've raised #1490 with some tests as well. See if that addressed your scenario as well :)

jatcwang:
I'm also ok with your fix here too, I think I finally understood what's going on after hitting the same issue (manifested in a different form) in my own branch. Just need to add some tests to exercise both product and ap. (Feel free to steal mine)

mdmoss:
Hi @jatcwang, I've been out of action for a few days - sorry for the long gap here.

I've tested your PR and it also resolves the issue I was having, so it seems like we've arrived at the same point.

My reasoning for changing `ap` rather than `product` was only that `ap` seems more fundamental (because `product` is implemented in terms of `ap` in https://github.com/typelevel/cats/blob/4869a8874c8abaab8ffd1f1cca06dbe86c1d49d9/core/src/main/scala/cats/Apply.scala#L112).

I've brought your test across, just to make sure it passes (which it seems to). Ideally we'd also test the output of check (as that was broken) but I couldn't find anywhere in tests that it's done at present, so I'll leave it for now.

If it's good practice to optimise by implementing `product` directly (rather than calling through to `ap`), that's also easily done.

jatcwang:
I added a few more tests in modules/munit/src/test/scala/doobie/munit/CheckerTests.scala to test the behaviour of Read after using product/ap. I tried testing Write's `product` as well, but h2 seems to think every input must be of type VARCHAR :thinking:  

Feel free to incorporate them into this PR and then we're should good to go!
```
  test("Read should select correct columns when combined with `product`") {
    import cats.syntax.all._
    import doobie.implicits._

    val ri = Read[Int]
    val rs = Read[String]

    // tupled use product under the hood
    val combined: Read[(Int, String)] = (ri, rs).tupled

    check(sql"SELECT 1, '2'".query(combined))
  }

  test("Read should select correct columns for checking when combined with `ap`") {
    val readInt = Read[(Int, Int)]
    val readIntToInt: Read[Tuple2[Int, Int] => String] =
      Read[(String, String)].map(i => k => s"$i,$k")

    val combined: Read[String] = readInt.ap(readIntToInt)

    check(sql"SELECT '1', '2', 3, 4".query(combined))
  }
```

mdmoss:
I've added your tests (confirming they fail without the `ap` fix, and pass with it). I've also added another test specifically for the order of gets produced by .product (seeing as that was the original issue).

jatcwang (approved):
@tpolecat if you're happy please merge this :)

PR #1490, opened 2021-07-04
Define Read's ap in terms of product (alternative fix of #1485)

jatcwang:
Fixes incorrect column ordering issue when using product

An alternative approach of #1485 to fixing the issue


mdmoss:
I've checked locally, and this seems to resolve the original issue :+1:

jatcwang:
Closing in favour of #1485

PR #1508, opened 2021-07-30
Add a paragraph about custom mappings for `Option[A]` in docs

danicheg:
According to the discussion in discord https://discord.com/channels/632277896739946517/632727524434247691/868065452109234177 it's worth adding a paragraph with caveats about trying to override the default behavior for mapping `Option[A]`, I suppose.

### DIFF
--- a/modules/docs/src/main/mdoc/docs/12-Custom-Mappings.md
+++ b/modules/docs/src/main/mdoc/docs/12-Custom-Mappings.md
@@ -177,6 +177,9 @@ implicit val jsonMeta: Meta[Json] =
 
 There are similar constructors for array types and other possibilities, but `other` is by far the most common in user code and we won't discuss the others here. See the Scaladoc for more information.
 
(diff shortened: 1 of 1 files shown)


danicheg:
@jatcwang it'd be great if you can look at this. That minor thing, but clarify some tricky moments IMO.

jatcwang:
@danicheg Thanks for the PR and sorry this fell off my radar. 
I think you can actually use a custom mapping but you'd have to do it via `Read`/`Write`. Something like `implicit val readMyType: Read[Mytype] = Read[Option[Int]].map(MyType.fromOptionInt)`.

The documentation does already state Get and Put instances must be non-nullalbe (under "Single-Column Type Mapping"), so I think it's worth documenting _how_ people can achieve custom mapping?

tpolecat:
I think the doc can be improved but I'm not sure this approach is helpful. Closing for now, please feel to re-open if you can figure out a more constructive way to address it.

PR #1519, opened 2021-08-08
Add SingleFragment.fromWrite

nigredo-tori:
Use `Write` over `Put` for interpolation, thus allowing interpolating values that are mapped to multiple parameters.


jatcwang:
Hi thanks @nigredo-tori. I'm a bit on the fence on this one since it makes it slightly more obscure what the resulting SQL will be (at least the number of `?`s). One scenario I can think of is a case class with JSON encoder/decoder instance + `import doobie.postgres.circe.implicits._`, which means the final SQL can be vastly different depending on whether the doobie import is present or not. (TBF, this wouldn't be a problem if Read/Write isn't automatically derived for case classes)

All in all, it is a pretty minor case against this so I'm ok with merging this in too. @tpolecat What do you think?

nigredo-tori:
> One scenario I can think of is a case class with JSON encoder/decoder instance + `import doobie.postgres.circe.implicits._`, which means the final SQL can be vastly different depending on whether the doobie import is present or not.

That import doesn't actually add implicit `Put`/`Get` for types with JSON codecs, only for `Json` itself. `Put` and `Get` instances for other types have to be explicitly created by the application using `pgDecoderGet` etc.

jatcwang:
You're right. That's better then :)

jatcwang (approved):

PR #1126, opened 2020-01-18
Add Config case class

sideeffffect:
Adds
 * a configuration case class that is suitable for PureConfig
  * helper method `Config.makeHikariConfig` for creating a `com.zaxxer.hikari.HikariConfig`
  * helper method `HikariTransactor.fromConfig` for creating a `doobie.hikari.HikariTransactor`

(credit goes to @jakubjanecek, the original author of [`scala-server-toolkit`](https://github.com/avast/scala-server-toolkit/) where this approach is used extensively)

PR #1541, opened 2021-08-31
Add Config case class

sideeffffect:
Revival of https://github.com/tpolecat/doobie/pull/1126

> Adds
> * a configuration case class that is suitable for PureConfig
>  * helper method `Config.makeHikariConfig` for creating a `com.zaxxer.hikari.HikariConfig`
>  * helper method `HikariTransactor.fromConfig` for creating a `doobie.hikari.HikariTransactor`


jatcwang:
@sideeffffect `main` is now CE3 so if you do need this in CE2 then you can target `series/0.13.x`, or you can update this PR to CE3 :)

sideeffffect:
The PR is based on `main` already. I just needed to fix the headers.

sideeffffect:
Oh, actually you're right :facepalm: I've made the necessary Cats Effect 3 changes.

modules/hikari/src/main/scala/doobie/hikari/Config.scala
```diff
@@ -0,0 +1,105 @@
...
+  dataSourceJNDI: Option[String] = None,
+  driverClassName: Option[String] = None,
+  initializationFailTimeout: Duration = Duration(1, TimeUnit.MILLISECONDS),
+  isolateInternalQueries: Boolean = false,
+  jdbcUrl: Option[String] = None,
+  readOnly: Boolean = false,
+  registerMbeans: Boolean = false,
+  schema: Option[String] = None,
+  transactionIsolation: Option[TransactionIsolation] = None,
+)
```
jatcwang, line 49: May I suggest that instead of default parameters we have a `val DEFAULT = Config(..)` in the companion object?

In the tests you can also construct a `Config` from `HikariConfig` (a bit like a round-trip test) which means we won't miss any new fields we add to `Config`.

sideeffffect: Default parameters are intentionally there. It works well with PureConfig -- which is the whole point of this config case class.
Otherwise I would go with the `val DEFAULT = Config(..)` myself from the start :smile_cat:

sideeffffect: You'll feel really dirty when doing that once you get used to PureConfig :joy_cat:

PR #1144, opened 2020-02-15
Composable fragments helpers

sideeffffect:
Make `and`/`or` & co. composable and add `comma` and `orderBy` `fragments` helpers

closes https://github.com/tpolecat/doobie/pull/903

PR #1553, opened 2021-09-12
Composable fragments helpers (new)

sideeffffect:
Make `and`/`or` & co. composable and add `comma` and `orderBy` fragments helpers

resuscitation of https://github.com/tpolecat/doobie/pull/1144
/cc @jatcwang


sideeffffect:
@jatcwang gentle ping :smile_cat:

modules/core/src/test/scala/doobie/util/FragmentsSuite.scala
```diff
@@ -25,87 +25,87 @@ class FragmentsSuite extends munit.FunSuite {
   val ofs  = List(1,2,3).map(n => Some(fr"$n").filter(_ => n % 2 =!= 0))
 
   test("values for one column") {
-    assertEquals(values(nel).query[Unit].sql, "VALUES (?), (?), (?) ")
+    assertEquals(values(nel).query[Unit].sql, "VALUES ?, ?, ? ")
```
jatcwang, line 28: Hmm I think this is a breaking change. I think the parenthesis is always needed?

https://www.db-fiddle.com/f/UQ1QzHQvkj1niuynfP4ob/0

modules/core/src/test/scala/doobie/util/FragmentsSuite.scala
```diff
@@ -25,87 +25,87 @@ class FragmentsSuite extends munit.FunSuite {
...
   }
 
   test("values for two columns") {
-    assertEquals(values(NonEmptyList.of((1, true), (2, false))).query[Unit].sql, "VALUES (?,?), (?,?) ")
+    assertEquals(values(NonEmptyList.of((1, true), (2, false))).query[Unit].sql, "VALUES ?,?, ?,? ")
```
jatcwang, line 32: Similar to above, we definitely want to keep `VALUES (?,?), (?,?)` because we're inserting two tuples no?

modules/core/src/main/scala/doobie/util/fragments.scala
```diff
@@ -5,68 +5,103 @@
...
+    notIn(f, fs0 :: fs1 :: fs.toList)
+
+  /** Returns `(f NOT IN (fs0, fs1, ...))`, or `true` for empty `fs`. */
+  def notIn[F[_]: Foldable, A: util.Put](f: Fragment, fs: F[A]): Fragment = fs.toList.toNel match {
+    case Some(fs) => parentheses(f ++ fr"NOT IN" ++ parentheses(comma(fs.map(a => fr"$a"))))
+    case None => fr"${true}"
+  }
+
+  /** Returns `(f1 AND f2 AND ... fn)`. */
+  def and(f1: Fragment, f2: Fragment, fs: Fragment*): Fragment =
```
jatcwang, line 47: Nit: I'm not too sure whether we need these 2+ param var args. If you statically have more than two fragments you can directly interpolate it in right?

```
fr"...WHERE $frag1 AND $frag2 AND $frag3"
```

which IMO is also more readable than


```
fr"...WHERE ${and(frag1, frag2, frag3)}"
```

modules/core/src/test/scala/doobie/util/FragmentsSuite.scala
sideeffffect: Of course! You're correct, I've fixed it. Thank you

modules/core/src/test/scala/doobie/util/FragmentsSuite.scala
sideeffffect: Fixed :+1:

modules/core/src/main/scala/doobie/util/fragments.scala
sideeffffect: Well, they were here before. But I can remove them, if you want :+1:

sideeffffect:
Thank you so much for the review @jatcwang and sorry for the delay.
Please let me know if you have any further concerns.

sideeffffect:
@jatcwang :pray:

sideeffffect:
@jatcwang :pray: 
Kindly please have a look once you have the time.

sideeffffect:
This PR refactors the helpers so that they're better composable among themselves. The key change is that the helpers always return something inside `(   )`. That ensures that piece of thing is worked with as an isolated unit. From your example, `or(fr"true", fr"false")` should always eagerly evaluate to `true` regardless of its surroundings/context (and `and(or(fr"true", fr"false"), fr"false")` to `false`). So in some cases, where the code incorrectly relied on the context, it may change the semantics, but the new behavior is correct and easier to use and compose. We can afford the change, because we're before a major version release, there won't (and hasn't been) a better time for this than now.

modules/core/src/main/scala/doobie/util/fragments.scala
jatcwang: True you're right.

jatcwang:
Thanks for the explanation @sideeffffect I think it's better to isolate the behaviour. I looked at my own usages of and/or and I think the scenario I raised is likely more rare than I initially thought - most of our use consists of `in` and `whereAnd/Or`. My other concern about and/or defaulting to `FALSE/TRUE` when provided collection is empty isn't a big deal as existing code couldn't use empty collections anyway.
Still have to think about some of the behaviour here (left a comment) and I'll raise another PR to make the generated SQL prettier too, but for now let's merge this in.
Thanks for the patience!

sideeffffect:
About the `FALSE/TRUE`, isn't that the most natural behavior for the `OR/AND` operator?

jatcwang:
@sideeffffect I think the correct default boolean value actually depends on the context where `or/and` is called (i.e. shouldn't be within)

See the following examples:
```
and(fr"c = 1", or(List.empty)) // I think it make sense for the output to just be "c = 1" instead of "c = 1 AND false"
```
Similarly
```
or(fr"c = 1", and(List.empty)) // Makes more sense to result in "c = 1" than "c =  1 OR true"
```

So in terms of composability I think it makes more sense for or/and (as well as other util functions) to filter out empty fragments from their arguments, and also return empty fragments if its input list is all empty.

I'll play around with it and see if it makes sense..

sideeffffect:
I think it's perfectly fine for `or(List.empty)` to evaluate to `false`. That's how most people think of using OR. If there isn't even a single true thing in `x OR y OR z OR ...`, then it should be `false`.

More formal mathematical/logical explanation:

For `AND`, the _neutral element_ is `true`, because `x AND true` = `x`.
For `OR`, the _neutral element_ is `false`, because `x OR false` = `x`.

And it is natural to substitute the _neutral element_ (or sometimes also called _identity element_) for `List.empty`.

jatcwang:
Yeah I think for `or` the default case when empty is quite intuitive. However I think for `and` it's not as clear cut and the desired default really depends on the context it's used.
I'm leaning towards updating the `and`/`or` helper method to require specifying a default if the supplied collection is empty. Let me see how it looks..

PR #1564, opened 2021-10-04
Add a cause to the munit Checker failure

brendanmaguire:
* Add an `ErrorItems` exception as the failure cause
* This allows clients to inspect and filter failures on their underlying cause `Item`s


jatcwang:
Just for my own learning, @brendanmaguire can you briefly explain what "client" you're using that this is helpful? :)

Issue #1151, opened 2020-02-21
Effectful logging


eugkhp:
Is there any way i could do logs as effects? Currently LogHanlder is  ```final case class LogHandler(unsafeRun: LogEvent => Unit)```, i was wondering if i could wrap it some to make it "safe"? ```final case class LogHandler(run: LogEvent => F[Unit])```.
And then use it somewhat like that: 
```
class SQLLogger[F[_]: Sync] { // or smth that i can run
    val effectfulLogger: LogHandler = {
      val myLogger: Logger = Logs.sync[F, F].byName("my sql logger") // https://github.com/TinkoffCreditSystems/tofu/tree/master/logging
      LogHandler {
        case Success(s, a, e1, e2) =>
          myLogger.info(s"${s}")
      }
    }
  }
```

tpolecat:
The current LogHandler is probably going away, to be replaced with a logging effect that enters via the transactor. For now you could hack this up with an `Effect` constraint since you can `.unsafeRunSync` anything with that constraint. Doesn't buy you much.

jatcwang:
In #1566 we now have effectful logging (through providing a LogHandlerM to the interpreter).

Which is also a good time to remove the old `LogHandler` (due to its less idea interface of `unsafeRun: LogEvent => Unit`).

If we go ahead with the removal of `LogHandler` we'll also remove the ability for users to change how each `Query`/`Update` is logged. I haven't really needed it, but would like to hear examples of the contrary.

Idea: Add a field `label: String` to `Query`/`Update` which is passed through the LogEvent interface. This makes it very easy for the LogHandlerM implementation to skip logging certain queries.

jozic:
We do use on demand logging extensively
it's very helpful for troubleshooting: 
a service call which results in a db call, has an additional param which when provided enables logging and service collects logs (including doobie logs) and return them as part of response
For majority of the the calls we don't need those logs, only for some sampling or debugging/troubleshooting

jatcwang:
@jozic Thanks for your input. Do you think my idea would work well for your usecase?

satorg:
Looks like this can be closed as completed in RC3.

jatcwang:
Thanks!

PR #1566, opened 2021-10-07
Provide loghandler to interpreters

oyvindberg:
(I've reworded and squashed the PR to make it easier to review)

Hey there. This PR seems to work as is,but I think it would benefit a lot from some discussion, especially regarding what the exposed API should be, naming, possible extensions if there are other things the interpreters could be more knowledgeable about.

### Problematic use case

We've been working on improving logging and tracing lately, and to that end we want to log mostly everything doobie does. We've actually been doing it for a while, and as a result we have hundreds of boilerplatey and convention-driven calls to `updateWithLogHandler` and friends, as I'm sure you can imagine. It gets worse when you need parameters to the `LogHandler`, because now all the DAOs are classes with parameters instead of static objects

### Proposed solution 

The idea behind this PR is to make that logging part of the interpreter rather than the queries and updates. Effectively, that means you can provide a `LogHandlerM` (an effectful copy of `LogHandler`) when creating an `xa`. Obviously it should have a better name, and maybe we would want to remove/deprecate the current query logging API.

### How

I generated a new node for all the algebras, `case class PerformLogging(logEvent: LogEvent) extends ...Op[LogHandler]`. 
This enables programs in all the generated algebras to log `LogEvent`s. The logging itself now runs in the output effect of the interpreter, so it can just delegate.

Currently it's added to all algebras even though it's likely only needed by `PreparedStatementIO`. 

This means that the interpreter, and in turn `xa` need a new optional parameter. This new parameter causes some friction in the API for creating an `xa`, since there are already many overloads and so default parameters are out of the picture. I sketched a builder-like pattern which I think we could grow in a binary-compatible way after 1.0 for `Transactor.fromDriverManager`

### Specific/General

Right now just a `LogHandlerM` is provided to the interpreter, but maybe it should be an `InterpreterParams` type instead, if there are more things like this which it could be useful to configure.

I also thought about adding it a bit later in the execution, so we would have interpreters with a signature like `NClobOp ~> Kleisli[M, (NClob, InterpreterParams), *]`, but didn't want to venture there unless that was seen to be clearly better.

### effectful `LogHandlerM`

`LogHandler` had to become effectful, the main driver for me other than it fitting better in with the rest, was that I'm experimenting with using `IOLocal` to thread through context/spans for logging. Running those effects disconnected from the rest of the IO runtime breaks that use case.

### DIFF
--- a/modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
+++ b/modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
@@ -75,12 +75,12 @@ import doobie.free.callablestatement.{ CallableStatementIO, CallableStatementOp 
 import doobie.free.resultset.{ ResultSetIO, ResultSetOp }
 
 object KleisliInterpreter {
-  def apply[M[_]: WeakAsync](logHandler: Option[LogHandlerM[M]]): KleisliInterpreter[M] =
+  def apply[M[_]: WeakAsync](logHandler: LogHandlerM[M]): KleisliInterpreter[M] =
     new KleisliInterpreter[M](logHandler)
 }
 
 // Family of interpreters into Kleisli arrows for some monad M.
-class KleisliInterpreter[M[_]](logHandler: Option[LogHandlerM[M]])(implicit val asyncM: WeakAsync[M]) { outer =>
+class KleisliInterpreter[M[_]](logHandler: LogHandlerM[M])(implicit val asyncM: WeakAsync[M]) { outer =>
   import WeakAsync._
 
   // The 14 interpreters, with definitions below. These can be overridden to customize behavior.
@@ -167,7 +167,7 @@ class KleisliInterpreter[M[_]](logHandler: Option[LogHandlerM[M]])(implicit val 
     override def suspend[A](hint: Sync.Type)(thunk: => A) = outer.suspend(hint)(thunk)
     override def canceled = outer.canceled[NClob]
 
-    override def performLogging(event: LogEvent) = Kleisli(_ => logHandler match {case Some(lh) => lh.run(event); case None => asyncM.pure(())})
+    override def performLogging(event: LogEvent) = Kleisli(_ => logHandler.run(event))
 
     // for operations using NClobIO we must call ourself recursively
     override def handleErrorWith[A](fa: NClobIO[A])(f: Throwable => NClobIO[A]) = outer.handleErrorWith(this)(fa)(f)
@@ -206,7 +206,7 @@ class KleisliInterpreter[M[_]](logHandler: Option[LogHandlerM[M]])(implicit val 
     override def suspend[A](hint: Sync.Type)(thunk: => A) = outer.suspend(hint)(thunk)
     override def canceled = outer.canceled[Blob]
 
-    override def performLogging(event: LogEvent) = Kleisli(_ => logHandler match {case Some(lh) => lh.run(event); case None => asyncM.pure(())})
+    override def performLogging(event: LogEvent) = Kleisli(_ => logHandler.run(event))
 
     // for operations using BlobIO we must call ourself recursively
     override def handleErrorWith[A](fa: BlobIO[A])(f: Throwable => BlobIO[A]) = outer.handleErrorWith(this)(fa)(f)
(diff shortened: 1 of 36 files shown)


modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
```diff
@@ -74,19 +75,12 @@ import doobie.free.callablestatement.{ CallableStatementIO, CallableStatementOp
...
-    }
+  def apply[M[_]: WeakAsync](logHandler: Option[LogHandlerM[M]]): KleisliInterpreter[M] =
+    new KleisliInterpreter[M](logHandler)
 }
 
 // Family of interpreters into Kleisli arrows for some monad M.
-trait KleisliInterpreter[M[_]] { outer =>
-
-  implicit val asyncM: WeakAsync[M]
+class KleisliInterpreter[M[_]](logHandler: Option[LogHandlerM[M]])(implicit val asyncM: WeakAsync[M]) { outer =>
```
jatcwang, line 83: Do you think we should take just a `LogHandlerM[M]` and define a NoopLogHandlerM?
This means that in `performLogging` we won't need to pay for a `match` every time.

jatcwang (commented):
One small comment but otherwise this looks great :)

oyvindberg:
Hey again @jatcwang . Sorry, didn't have the bandwidth before now. 

I added a commit with a no-op variant of `LogHandlerM` now, see if it's what you had in mind. 

I had to go through some gymnastics with the `Transactor` builder methods to capture the `Applicative` constraint late in order to maintain syntax. I suppose `newH2Transactor` should get the same builder treatment, maybe others.


Any name suggestions for `LogHandlerM`? Otherwise I'll go ahead and update documentation for this.

jatcwang:
Thanks that's exactly what I was thinking :)

I think LogHandlerM name is quite clear.

oyvindberg:
Great, let's go on then. I added an example of usage to the docs in the latest commit, any input there?

tpolecat:
I'll have a look at this soon. Let's sit on it for now.

oyvindberg:
friendly almost-end-of-month nudge @tpolecat :)

jatcwang (approved):
LGTM

jozic:
@oyvindberg  you mentioned about deprecating the current API, will the new API support logging on demand? Like if logging statements per request (if request has some flag set for example) is needed or logs should be part of some response (but not of every response)

oyvindberg:
@jozic Apologies for answering late. I haven't received any comments about the old API, so I suppose it stays like it is. 

Also yes! Here is a modified example of what I put in the docs which shows how you can leverage `IOLocal` to only log some subset of queries.

```scala
object Tester extends IOApp.Simple {

  def users = List.range(0, 4).map(n => s"user-$n")

  def transactorWithLogging(logHandlerM: LogHandlerM[IO]): Transactor[IO] =
    Transactor.fromDriverManager[IO]
      .withLogHandler(logHandlerM)( // thread through the logHandler here
        "org.h2.Driver",
        "jdbc:h2:mem:queryspec;DB_CLOSE_DELAY=-1",
        "sa", ""
      )

  override val run: IO[Unit] =
    for {
      // define an IOLocal where we store the user id if we enable logging for this user. This will be done as you start the app
      enableLoggingForUser <- IOLocal[Option[String]](None)
      xa = transactorWithLogging {
        case Success(sql, _, _, _) =>
          enableLoggingForUser.get.flatMap {
            case Some(user) => IO(println(s"sql for user $user: '$sql'"))
            case None => IO.unit
          }
        case _ => IO.unit
      }
      // run a bunch of queries
      _ <- users.parTraverse(user =>
        for {
          // determine if you want to enable logging. do this for instance in http middleware
          _ <- if (math.random() < 0.5) enableLoggingForUser.set(Some(user)) else IO.unit
          _ <- sql"select 1".query[Int].unique.transact(xa)
        } yield ()
      )
    } yield ()
}
```

hamnis:
Any chance of this getting merged?

Prillan:
This would be excellent to have! I'm currently doing a very convoluted `Dispatcher[F]` based approach that feels pretty brittle, native support in the library would fix that :)

jatcwang:
Hi all sorry for the long wait. I think this looks good.

We should remove the old LogHandler before 1.0 to reduce confusion/noise for users. I think it's worth discussing whether Doobie should support on/off logging for individual queries and we can do it in #1151

Thanks for the PR @oyvindberg!

PR #1567, opened 2021-10-07
Fix effectful `copyIn` (Fixes #1512)

### LINKED ISSUE #1512: Postgres COPY fails past minChunkSize when using non-Pure effects
We're seeing that the copy operation gets closed once we have a stream that's larger than the minChunkSize for the .copyIn() method. We only see this when using effects other than Pure.

This is true for both 1.0.0-M1 and 1.0.0-M5.

Failing test code follows:

```
package io.citrine.tbl.ingest

import cats.effect.IO
import cats.implicits.catsSyntaxApplicativeId
import doobie.ConnectionIO
import doobie.implicits._
import doobie.postgres.implicits._
import doobie.util.transactor.Transactor
import doobie.util.transactor.Transactor.Aux
import munit.CatsEffectSuite
import org.postgresql.ds.PGSimpleDataSource

import javax.sql.DataSource

class Fs2CopyTest extends CatsEffectSuite {

  val minChunkSize = 200

  val datasource: DataSource = {
    val ds = new PGSimpleDataSource
    ds.setURL(System.getenv("DB_URL"))
    ds.setUser(System.getenv("DB_USER"))
    ds.setPassword(System.getenv("DB_PASSWORD"))
    ds
  }
  val xa: Aux[IO, DataSource] = Transactor.fromDataSource[IO](datasource, scala.concurrent.ExecutionContext.global)

  test("A stream with a Pure effect inserts items properly") {

    sql"""
        DROP TABLE IF EXISTS demo;
        CREATE TABLE demo(id BIGSERIAL PRIMARY KEY NOT NULL, data BIGINT NOT NULL);
         """.update.run
      .transact(xa)
      .unsafeRunSync()

    // A pure stream is fine - can copy many items
    val count = 10000
    val stream = fs2.Stream.emits(1 to count)

    (sql"COPY demo(data) FROM STDIN").copyIn(stream, minChunkSize).transact(xa).unsafeRunSync()

    val queryCount =
      sql"SELECT count(*) from demo".query[Int].unique.transact(xa).unsafeRunSync()

    assertEquals(queryCount, count)
  }

  test("A stream with a ConnectionIO effect copies <= than minChunkSize items") {

    sql"""
        DROP TABLE IF EXISTS demo;
        CREATE TABLE demo(id BIGSERIAL PRIMARY KEY NOT NULL, data BIGINT NOT NULL);
         """.update.run
      .transact(xa)
      .unsafeRunSync()

    // Can copy up to minChunkSize just fine with ConnectionIO
    val inputs = (1 to minChunkSize)
    val stream = fs2.Stream.emits[ConnectionIO, Int](inputs)
      .evalMap(i => (i + 2).pure[ConnectionIO])

    (sql"COPY demo(data) FROM STDIN").copyIn(stream, minChunkSize).transact(xa).unsafeRunSync()

    val queryCount =
      sql"SELECT count(*) from demo".query[Int].unique.transact(xa).unsafeRunSync()

    assertEquals(queryCount, minChunkSize)
  }

  test("A stream with a ConnectionIO effect fails to copy items with count > minChunkSize") {

    sql"""
        DROP TABLE IF EXISTS demo;
        CREATE TABLE demo(id BIGSERIAL PRIMARY KEY NOT NULL, data BIGINT NOT NULL);
         """.update.run
      .transact(xa)
      .unsafeRunSync()

    // Can't copy over minChunkSize with ConnectionIO - copy operation gets closed
    val inputs = (1 to minChunkSize + 1)
    val stream = fs2.Stream.emits[ConnectionIO, Int](inputs)
      .evalMap(i => (i + 2).pure[ConnectionIO])

    (sql"COPY demo(data) FROM STDIN").copyIn(stream, minChunkSize).transact(xa).unsafeRunSync()

    val queryCount =
      sql"SELECT count(*) from demo".query[Int].unique.transact(xa).unsafeRunSync()

    assertEquals(queryCount, minChunkSize + 1)
  }
}
```

oyvindberg:
The challenge faced here was to capture the number of affected rows when ending a `COPY`.

To safely end a copy, it needs to be run in the finalizer of a `bracket`. We cannot retrieve values from a finalizer, so there was some logic in place to call `PFCI.endCopy` in two places and use `isActive` to determine if it should be called the second time. `isActive` seems to not be an `!isCopyEnded`, but rather "is connection reserved for this Copy operation?".

The fix is to replace the double closing with a reference

### DIFF
--- a/modules/postgres/src/main/scala/doobie/postgres/syntax/FragmentSyntax.scala
+++ b/modules/postgres/src/main/scala/doobie/postgres/syntax/FragmentSyntax.scala
@@ -5,6 +5,7 @@
 package doobie.postgres.syntax
 
 import cats.Foldable
+import cats.effect.Ref
 import cats.effect.kernel.Resource
 import cats.syntax.all._
 import doobie._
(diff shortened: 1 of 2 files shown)


oyvindberg:
@jatcwang When you have time :)

jatcwang (approved):
LGTM Thanks for the fix @oyvindberg!

PR #1530, opened 2021-08-29
remove sbt-dotty (and bump sbt to 1.5.5)

SethTisue:

PR #1573, opened 2021-10-12
remove sbt-dotty (and bump sbt to 1.5.5)

SethTisue:
like #1530, but for `main` branch


tpolecat:
I'm pulling in monix for one example in the docs. Weird that this didn't get caught earlier. It just needs an update probably.

SethTisue:
> It just needs an update probably

I don't think there's a Monix version on CE3: https://github.com/monix/monix/issues/1502

Monix seems not-very-maintained these days

tpolecat:
Thanks @SethTisue!

PR #1585, opened 2021-11-01
Managing Connections page no longer breaks

LazyBun:
Currently [Managing Connections page](https://tpolecat.github.io/doobie/docs/14-Managing-Connections.html#using-a-hikaricp-connection-pool) breaks after an empty code block. 

I've implemented this very complicated fix to resolve this issue.


jatcwang (approved):
Thanks :)

PR #1594, opened 2021-11-16
Enable support for `derives` in Scala 3 for `Read`, `Write` and `Text`

oyvindberg:
In order for this keyword to work, there needs to be a `derived` method available in the companion objects. This gives some of the `given`s the expected name.

This is what it looks like without this patch:
```
[error] -- [E008] Not Found Error: /home/.../doobie/modules/core/src/test/scala-3/doobie/util/ReadSuitePlatform.scala:25:51 
[error] 25 |    case class Foo(a: String, b: Int) derives util.Read
[error]    |                                                   ^
[error]    |value derived is not a member of object doobie.util.Read, but could be made available as an extension method.
[error]    |
[error]    |The following import might make progress towards fixing the problem:
[error]    |
[error]    |  import munit.Clue.generate
[error]    |
[error] one error found
```

### DIFF
--- a/modules/core/src/main/scala-3/util/ReadPlatform.scala
+++ b/modules/core/src/main/scala-3/util/ReadPlatform.scala
@@ -20,7 +20,7 @@ trait ReadPlatform:
     )
 
   // Generic Read for products.
-  given [P <: Product, A](
+  given derived [P <: Product, A](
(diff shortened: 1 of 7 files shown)


jatcwang:
Thanks @oyvindberg!

PR #1595, opened 2021-11-17
Make connectionIO syntax available without import

### LINKED ISSUE #1592: Make ConnectionIO syntax available without an import
If the `transact` extension syntax is defined in the `ConnectionOp` companion object then it should no longer require an import - http4s did something similar in https://github.com/http4s/http4s/pull/5083. This would require merging the `free` and `core` modules, `ConnectionOp` is defined in `free` and cannot currently reference `Transactor`, which is in `core` - would this be a significant downside in practice?

bplommer:
This requires combining the `free` and `core` modules to make Transactor visible from ConnectionOp. I added a `macros` module so the scala 2 macros would be usable from core.

Resolves #1592


jatcwang (commented):
Thanks @bplommer. @tpolecat are you ok with merging the two modules?

bplommer:
Is this still wanted? Happy to fix the conflicts it if so

jatcwang:
Thanks @bplommer. Sorry this has been lingering for a long time.

I think we can close this ticket for now. I'm thinking of transferring doobie to the typelevel github & maven org, which will be a good time to merge the modules as it minimizes confusion.

I think we can close this PR for now but leave the issue open. When we've done the necessary work we can implement this change again on top, to avoid having to deal with conflicts. Thanks!

bplommer:
No worries!

PR #1596, opened 2021-11-22
Fixed docs broken internal link for 12-Custom-Mappings

mbablok:
Hello,

I stumbled upon a broken link in the microsite and it was a quick fix. I hope I do things correctly, as I am not experienced in contributing.

Thanks for the great work!


tpolecat:
Thank you!

Issue #1597, opened 2021-11-23
Fix the name of the type classes associated with Get and Write type


rcardin:
Hi.
I was surfing into the Doobie code base, and I saw that the name associated with the functor instance for the `Get` type has the wrong name:

```scala
trait GetInstances extends GetPlatform {
  import Predef._ // for array ops

  /** @group Instances */
  implicit val FunctorPut: Functor[Get] =
    new Functor[Get] {
      def map[A, B](fa: Get[A])(f: A => B): Get[B] =
        fa.map(f)
    }
  // ...
}
```

I supposed it should be `FunctorGet`. If somebody confirms this, I can make the change quite easily :)

I found the same for the `Write`type class:

```scala
object Write extends WritePlatform {

  def apply[A](implicit A: Write[A]): Write[A] = A

  implicit val ReadContravariantSemigroupal: ContravariantSemigroupal[Write] =
    new ContravariantSemigroupal[Write] {
      def contramap[A, B](fa: Write[A])(f: B => A) = fa.contramap(f)
      def product[A, B](fa: Write[A], fb: Write[B]) = fa.product(fb)
    }
  //...
}
```

In this case, I think the right name is `WriteContravariantSemigroupal`

PR #1598, opened 2021-11-23
Fixed name of the implicit val defining the function for the Get and Write types

rcardin:
This PR addresses issue  #1597.

Let me know if I understood it wrong.


jatcwang (approved):

jatcwang (approved):
Thanks. This does break binary compatibility but since we're in RC I think this is fine.

PR #1599, opened 2021-11-25
Add support for JSON in H2 using circe

andrzejressel:
This PR introduces support for JSON type in H2.

Code and tests are based on Postgres implementation.

I've tried using native H2 type `ValueJson`, but H2 really insists on using VARCHAR/bytearray:

```scala
    implicit val jsonPut: Put[Json] =
      Put.Advanced.other[ValueJson](
        NonEmptyList.of("JSON", "json")
      ).tcontramap(a =>
        ValueJson.fromJson(a.noSpaces)
      )
	  
	implicit val jsonGet: Get[Json] =
      Get.Advanced.other[ValueJson](
        NonEmptyList.of("JSON", "json")
      ).temap(a =>
        parse(a.getString).leftMap(_.show)
      )
```

### DIFF
--- a/modules/docs/src/main/mdoc/docs/16-Extensions-H2.md
+++ b/modules/docs/src/main/mdoc/docs/16-Extensions-H2.md
@@ -10,6 +10,26 @@ libraryDependencies += "org.tpolecat" %% "doobie-h2" % "$version$"
 
 This library pulls in H2 as a transitive dependency.
 
+There are extensions available for dealing with JSON by using Circe, if you like to use those, include this dependency:
+
+@@@ vars
+
+```scala
+libraryDependencies += "org.tpolecat" %% "doobie-h2-circe" % "$version$"
+```
+
+@@@
+
+Then, you will be able to import the implicits for dealing with JSON:
+
+@@@ vars
+
+```scala
+import doobie.h2.circe.json.implicits
+```
+
+@@@
+
 ### Array Types
 
 **doobie** supports H2 arrays of the following types:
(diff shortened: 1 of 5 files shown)


modules/docs/src/main/mdoc/docs/16-Extensions-H2.md
```diff
@@ -10,6 +10,26 @@ libraryDependencies += "org.tpolecat" %% "doobie-h2" % "$version$"
...
+```
+
+@@@
+
+Then, you will be able to import the implicits for dealing with JSON:
+
+@@@ vars
+
+```scala
+import doobie.h2.circe.json.implicits
```
jatcwang, line 28: Do you mean
```suggestion
import doobie.h2.circe.json.implicits._
```
?

andrzejressel: To be honest I was thinking about adding `._`, but I haven't saw that in PostgreSQL implementation that I based my code on.

Regarding `h2DecoderGetT` and `h2EncoderGetT`, from what I understand they're used mostly for testing.

jatcwang: > Regarding h2DecoderGetT and h2EncoderGetT, from what I understand they're used mostly for testing.

They are used for creating instances of `Put[A]/Get[A]` provided that you have `Encoder[A]/Decoder[A]`. This isn't automatically derived (i.e. not `implicit def`) because that can often lead to the wrong behaviour. (Just because you have an Encoder[A] does not mean you want to write `A` as a JSON column in the database).

Anyway I can do the doc improvements in another PR. Thanks for your contribution :)

PR #1609, opened 2021-12-08
add sqlite3 information and Scala3 information in 01-Indroduction.md.…

benluo:
… Add sqlite information in 03-Connecting.md.

### DIFF
--- a/modules/docs/src/main/mdoc/docs/01-Introduction.md
+++ b/modules/docs/src/main/mdoc/docs/01-Introduction.md
@@ -79,7 +79,15 @@ libraryDependencies ++= Seq(
 
 The `-Ypartial-unification` compiler flag enables a bug fix that makes working with functional code significantly easier. See the Cats [Getting Started](https://github.com/typelevel/cats#getting-started) for more info on this if it interests you. **If you're using Scala 2.13+ the compiler no longer accepts that option**.
 
-If you are not using PostgreSQL you can omit `doobie-postgres` and will need to add the appropriate JDBC driver as a dependency. Note that there is a `doobie-h2` add-on if you happen to be using [H2](http://www.h2database.com/).
+If you want to use scala 3.x, you cannot use doobie-specs2 yet.
+
+If you are not using PostgreSQL you can omit `doobie-postgres` and will need to add the appropriate JDBC driver as a dependency. Note that there is a `doobie-h2` add-on if you happen to be using [H2](http://www.h2database.com/). 
+
+If you want to use [Sqlite](https://sqlite.com/index.html), you have to use a little different config in `libraryDependencies`.
+
+```scala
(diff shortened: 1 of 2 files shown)


modules/docs/src/main/mdoc/docs/03-Connecting.md
```diff
@@ -46,6 +46,23 @@ val xa = Transactor.fromDriverManager[IO](
 )
 ```
 
+To use sqlite, your code like
```
jatcwang, line 49: Thanks for the contribution :) Since there are no specific integration with sqlite I'd like to avoid adding too much info about it, as all that's different is the driver class name and the connection URL.

May I suggest that we change all of this to something like the following to be more generic and concise?

```
Note: If you're connection to anyother databases , make sure to change the JDBC driver class name and connection URL according to their documentation. 
```

modules/docs/src/main/mdoc/docs/01-Introduction.md
```diff
@@ -79,7 +79,15 @@ libraryDependencies ++= Seq(
...
-If you are not using PostgreSQL you can omit `doobie-postgres` and will need to add the appropriate JDBC driver as a dependency. Note that there is a `doobie-h2` add-on if you happen to be using [H2](http://www.h2database.com/).
+If you want to use scala 3.x, you cannot use doobie-specs2 yet.
+
+If you are not using PostgreSQL you can omit `doobie-postgres` and will need to add the appropriate JDBC driver as a dependency. Note that there is a `doobie-h2` add-on if you happen to be using [H2](http://www.h2database.com/). 
+
+If you want to use [Sqlite](https://sqlite.com/index.html), you have to use a little different config in `libraryDependencies`.
+
+```scala
+   "org.xerial" % "sqlite-jdbc" % "3.36.0.3",
+```
```
jatcwang, line 90: In similar spirit, I would change this to be something more generic:

```
For connecting to any other database, please make sure you have added their JDBC driver in your dependencies.
```

modules/docs/src/main/mdoc/docs/03-Connecting.md
benluo: I total agree with you. 

It may be my specific problem. I PR this because I never learn(in deeper way) java language. So it's very difficult for me to find the solution for sqlite.

Please close the PR.

modules/docs/src/main/mdoc/docs/01-Introduction.md
benluo: I total agree with you. 

It may be my specific problem. I PR this because I never learn(in deeper way) java language. So it's very difficult for me to find the solution for sqlite.

Please close the PR.

jatcwang:
Thanks for raising the PR all the same @benluo. I'll raise another PR as I have suggested to make it easier for users of other DB.

Issue #1264, opened 2020-09-10
No mapping for PostGIS 'geography' type


marko-asplund:
PostGIS Types support seems to be missing a mapping for the `geography` type:

```
 sql"select location from foo limit 1".query[org.postgis.Point].unique.transact(xa).unsafeRunSync 
doobie.util.invariant$InvalidObjectMapping: SQL object of class org.postgresql.util.PGobject cannot be cast to mapped class org.postgis.PGgeometry.
  doobie.util.Get$Advanced$.$anonfun$other$1(get.scala:163)
```

Database schema type is the following:
```
location     | geography(Point,4326)       |           |          |
```

More details on Gitter: https://gitter.im/tpolecat/doobie?at=5f590147dfaaed4ef53f65c1

PR #1625, opened 2022-01-04
Postgis: Add custom mapping for geography -> point postgres conversion

jgoday:
See #1264.

This PR adds both Get/Put implicits (doobie.postgres.PgisInstances) to allow conversions between Postgis Geography(Point) (coordinates as spherical) and Point types.

Surely additional conversions (POLYLINE, POLYGON...) are needed, It can be implemented in a similar way if this implementation is correct.

Here is an example that would be valid with this change:

```scala
  /// CREATE TABLE countries(name TEXT PRIMARY KEY, location GEOGRAPHY(POINT, 4326));
  case class Country(code: String, location: Option[Point])

  def findCountry(n: String): ConnectionIO[Option[Country]] =
    sql"select name, location from countries where name = $n".query[Country].option
```

### DIFF
--- a/modules/postgres/src/main/scala/doobie/postgres/pgisinstances.scala
+++ b/modules/postgres/src/main/scala/doobie/postgres/pgisinstances.scala
@@ -6,8 +6,8 @@ package doobie.postgres
 
 import doobie._
 import doobie.util.invariant._
-
 import org.postgis._
+import org.postgresql.util.PGobject
 
 import scala.reflect.ClassTag
 import org.tpolecat.typename._
@@ -15,9 +15,10 @@ import org.tpolecat.typename._
 trait PgisInstances {
 
   // PostGIS outer types
-  implicit val PGgeometryType: Meta[PGgeometry] = Meta.Advanced.other[PGgeometry]("geometry")
-  implicit val PGbox3dType: Meta[PGbox3d]       = Meta.Advanced.other[PGbox3d]("box3d")
-  implicit val PGbox2dType: Meta[PGbox2d]       = Meta.Advanced.other[PGbox2d]("box2d")
+  implicit val PGGeographyType: Meta[PGgeography] = Meta.Advanced.other[PGgeography]("geography")
+  implicit val PGgeometryType: Meta[PGgeometry]   = Meta.Advanced.other[PGgeometry]("geometry")
+  implicit val PGbox3dType: Meta[PGbox3d]         = Meta.Advanced.other[PGbox3d]("box3d")
+  implicit val PGbox2dType: Meta[PGbox2d]         = Meta.Advanced.other[PGbox2d]("box2d")
 
   // Constructor for geometry types via the `Geometry` member of PGgeometry
   @SuppressWarnings(Array("org.wartremover.warts.AsInstanceOf", "org.wartremover.warts.Throw"))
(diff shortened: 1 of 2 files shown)


jatcwang:
Thanks!

jatcwang:
Hi @jgoday, I just noticed that there's already an instance for Point in this file (`implicit val PointType: Meta[Point]                           = geometryType[Point]`). (Note that Meta = Put + Get).

I'm not familiar with using PostGIS in general, is there much difference between getting the point via geography vs geometry?

jgoday:
Hi @jatcwang 
> Hi @jgoday, I just noticed that there's already an instance for Point in this file (`implicit val PointType: Meta[Point] = geometryType[Point]`). (Note that Meta = Put + Get).

You're right, I have unintentionally tried to avoid ambiguity between Geometry / Geography Meta conversions,
and an error has been created, not detected by the current tests. Sadly I was only focusing on the conversion from the Geography type (really sorry for this).

So, **this PR needs to be reverted**, because it breaks the geometry -> point implicit conversions via previous ```PointType: Meta[Point]                           = geometryType[Point]```


> 
> I'm not familiar with using PostGIS in general, is there much difference between getting the point via geography vs geometry? If not, I think it is ok to revert this PR? 

Geography points are not cartesian, are spherical (angular coordinates).
Currently, there's only possible to read points if they are defined as geometry in the database.
(PGgeometry and PGgeography are not related, both extends PGgeo separatedly)
http://postgis.net/workshops/postgis-intro/geography.html

> What issues where you trying to solve with this PR?

See #1264, basically if you have a column in Postgis of type Geography(POINT) and you want to retrieve it with doobie as Point (angular coordinates).
```
  /// CREATE TABLE countries(name TEXT PRIMARY KEY, location GEOGRAPHY(POINT, 4326));
  case class Country(code: String, location: Option[Point])

  def findCountry(n: String): ConnectionIO[Option[Country]] =
    sql"select name, location from countries where name = $n".query[Country].option
```

Would be a valid solutions to define another different implicit conversions object, such as PgisGeographyInstance,  to avoid geometry/geography conflicts?

Or would it be better to try changing the current PointType conversion to work with both Geometry / Geography?

Sorry again for the inconveniences.

jatcwang:
No worries I missed it too :)

I agree with what you're proposing - a different implicit object would be fine.

If you got time, can you also add some note in https://github.com/tpolecat/doobie/blob/main/modules/docs/src/main/mdoc/docs/15-Extensions-PostgreSQL.md to note that Geography Points are supported but require a different import, and because the `Point` type used in the same in both cases, care must be taken such that the instances aren't in the same 

An alternative solution I can think of is to require the user to explicitly import either the geography Point instance or geometry Point instance, something like:

```
// other imports
import doobie.postgres.pgisimplicits._

case class HasGeographyPoint(geoPoint: Point)
object HasGeographyPoint {
  // Derive the Read instance once to make sure the correct Point instance is used all the time when Read[HasGeographyPoint] is required.
  implicit read: Read[HasGeographyPoint] = {
    implicit val p: Meta[Point] = doobie.postgres.pgisimplicits.geometryPoint // note geometryPoint isn't implicit by default so isn't in the implicit scope by the import at the top of the file.
    Read.recordRead
  }
}
```
The docs will need to point this out, ofc.

Up to you which solution you go for :slightly_smiling_face:

jgoday:
Hi @jatcwang, I just created this new PR #1633 defining a new object with the geography implicit conversions.

PR #1632, opened 2022-01-11
remove jdk9+ stuff

tpolecat:
Somehow [this commit](https://github.com/tpolecat/doobie/commit/3231e7928e7eee5a4a91b32d7bdc2836381d1f0e) got dropped and we ended up with JDK9+ JDBC stuff in here. This removes it and pushes the CI version back to 8.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         scala: [2.12.15, 2.13.7, 3.1.0]
-        java: [temurin@11]
+        java: [temurin@8]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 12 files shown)


tpolecat:
ffs

tpolecat:
now tests are failing, what on earth

jatcwang:
This may help https://stackoverflow.com/questions/61490701/offsetdatetime-parse-different-behavior-in-java-8-and-java-11

Seems like OffsetDateTime parsing become more lenient in JDK 9+ (optional minutes)

tpolecat:
It turns out I was installing JDK8 but the build was actually running in JDK11 all this time, per @armanbilge. Computers are the worst.

SethTisue:
scala/community-build@e8d41c1683142a559fc054ad624f1f258aeb80b1

PR #1633, opened 2022-01-12
define pgisgeographyinstances for geography postgis types

jgoday:
See #1264.

Previous reverted PR #1625 overrides default postgis PointType implicit conversion by mistake.
 
This PR adds both pgisgeographyimplicits conversions to use geomtries backed by geodetic coordinataes (postgis geography).

Here is an example that would be valid with this change:

  /// CREATE TABLE countries(name TEXT PRIMARY KEY, location GEOGRAPHY(POINT, 4326));
  case class Country(code: String, location: Option[Point])

  def findCountry(n: String): ConnectionIO[Option[Country]] =
    sql"select name, location from countries where name = $n".query[Country].option


jatcwang (approved):
Thanks :)

PR #1654, opened 2022-02-12
Add Monoid and Semigroup instances for ConnectionIO

Fristi:
Using this quite a lot to combine affected rows, but also upon retrieval of stuff to combine sets together.

I've found there is a generic way to express this:

```scala
implicit final def semigroupF[A](implicit A: Semigroup[A], F: Apply[F]): Semigroup[F[A]] =
    Apply.semigroup[F, A]
```

It's however not standard in cats, and I think it would fit here well?

Any suggestions welcome!

### DIFF
--- a/modules/core/src/main/scala/doobie/free/Aliases.scala
+++ b/modules/core/src/main/scala/doobie/free/Aliases.scala
@@ -4,6 +4,7 @@
 
 package doobie.free
 
+import cats.{Monoid, Semigroup}
 import doobie.WeakAsync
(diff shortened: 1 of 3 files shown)


jatcwang:
Thanks @Fristi I think this makes sense. Would you be able to add some tests for me? :)

Fristi:
Done

jatcwang (approved):

jatcwang:
Thanks!

PR #1668, opened 2022-03-22
Added weaver module

danielkarch:
This PR adds the module `doobie-weaver` which allows testing queries
etc. in the weaver testing framework.

If you think this would be worthwhile to add to doobie, I would add documentation to this PR.


danielkarch:
I just added documentation.

jatcwang:
Thanks!

PR #1690, opened 2022-05-02
Add functions for creating HikariTransactor that automatically create connectEC

sideeffffect:
These new methods provide good default behavior. Using these makes it difficult to shoot oneself to own foot by passing in a wrong EC.

### DIFF
--- a/modules/hikari/src/main/scala/doobie/hikari/Config.scala
+++ b/modules/hikari/src/main/scala/doobie/hikari/Config.scala
@@ -19,7 +19,7 @@ import scala.concurrent.duration.Duration
 /** Configuration case class, susceptible to PureConfig.
   * Helps with creating `com.zaxxer.hikari.HikariConfig`,
   * which in turn is used to create `doobie.hikari.HikariTransactor`.
-  * See the method `HikariTransactor.fromConfig` */
+  * See the method `HikariTransactor.fromConfigAutoEc` */
 final case class Config(
   catalog: Option[String] = None,
   connectionTimeout: Duration = Duration(30, TimeUnit.SECONDS),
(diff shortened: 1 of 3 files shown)


modules/hikari/src/main/scala/doobie/hikari/Config.scala
```diff
@@ -19,15 +19,15 @@ import scala.concurrent.duration.Duration
...
 final case class Config(
   catalog: Option[String] = None,
   connectionTimeout: Duration = Duration(30, TimeUnit.SECONDS),
   idleTimeout: Duration = Duration(10, TimeUnit.MINUTES),
   leakDetectionThreshold: Duration = Duration.Zero,
-  maximumPoolSize: Option[Int] = None,
+  maximumPoolSize: Option[Int] = Some(10),
   maxLifetime: Duration = Duration(30, TimeUnit.MINUTES),
-  minimumIdle: Option[Int] = None,
+  minimumIdle: Option[Int] = Some(10),
```
jatcwang, line 30: If these values are already defaulted to 10 in hikari, do we need to set them explicitly here?

sideeffffect: Nope, but it makes it more user friendly. Users will immediately see what are the default values that they will get. Not doing so would be more obscure and users would have to do more investigation to see what will actually happen.

sideeffffect:
@jatcwang gentle ping :smile_cat:

modules/hikari/src/main/scala/doobie/hikari/Config.scala
jatcwang: Gotcha thanks!

sideeffffect:
🎉

PR #1693, opened 2022-05-17
Add On Air Entertainment to list of adopters.

OnAirEntertainment-Scala:
Added On Air Entertainment to the list of doobie adopters.


jatcwang:
Thanks!

PR #1701, opened 2022-06-25
build.sbt updates

guymers:
Remove adding scala-2/scala-3 directories as sbt does it now.

Remove Scala 3 exclusions, specs2 is available for Scala 3 and docs can
be complied with Scala 3 with the removal of `scala-collection-compat`.


jatcwang:
Thanks!

PR #1735, opened 2022-08-26
Update Postgres Java time instances

guymers:
The Postgres Java time instances have been updated to more closely match
what the driver allows. To verify this, the check tests have been
updated to also fetch the value.


jatcwang:
Thanks

Issue #1481, opened 2021-06-22
Analysis should catch mismatch for types with or without timezone in PostgreSQL instances


nigredo-tori:
- Writing or reading `OffsetDateTime` or `Instant` as `TIMESTAMP` is generally an error.
- Writing or reading `LocalDateTime` as `TIMESTAMPTZ` is generally an error.

This was highlighted for me by a recent-ish change, where the implementation for corresponding instances for PostgreSQL has changed. For example, where previously `Instant` was passed to the database as `java.util.Date`, with the system timezone, it is now passed as `OffsetDateTime` with zero offset. This means that any code that was writing `Instant` to `TIMESTAMP` columns was silently broken by this. Reading this data has its own set of problems.

The most annoying part for me is that the analysis checks in place don't catch that, because (AFAICT) pgjdbc doesn't correctly label output **columns types** as `TIMESTAMP_WITH_TIMEZONE` to preserve compatibility with Hibernate. :facepalm: Maybe there is some other piece of metadata we can use here? If, say, `getColumnName` produces different values for `TIMESTAMP` and `TIMESTAMPTZ`, we can use that to disambiguate the two. Although I imagine that would require an overhaul of the `Get` and `Meta` machinery.

nigredo-tori:
I've done some testing, and `getColumnTypeName` seems to produce "timestamp" and "timestamptz" depending on the type. So we can definitely use it to disambiguate the two.

jatcwang:
Yeah I did try doing this when adding first-class PostgreSQL java.time support but observed the same thing you did. I think we could switch to use getColumnTypeName, but I think that'll require a rewrite of the analyzer to be DB-specific as well as the instances (e.g. Put[Instant]) now needs to specify their database type

nigredo-tori:
> a rewrite of the analyzer to be DB-specific

Or we can move some of the analysis logic to the instances, since they are already, effectively, DB-specific. I.e. `Get` can check a column type and column type name from the metadata, and tell whether those match the expected type. I imagine this can even be done without breaking binary compatibility, if we preserve all the case class machinery for `Get` and `Put` - although that will certainly be noisy and ugly.

PR #1736, opened 2022-08-29
Better Java time instances for PostgreSQL

guymers:
Given that the Postgres JDBC driver does not return *WithTimezone types,
change the type when creating meta column and parameter objects. This
allows the Java time instance for `OffsetDateTime` to be defined as
requiring a column to be a `TimestampWithTimezone`.

### DIFF
--- a/modules/core/src/main/scala/doobie/util/analysis.scala
+++ b/modules/core/src/main/scala/doobie/util/analysis.scala
@@ -20,9 +20,28 @@ object analysis {
 
   /** Metadata for the JDBC end of a column/parameter mapping. */
   final case class ColumnMeta(jdbcType: JdbcType, vendorTypeName: String, nullability: Nullability, name: String)
+  object ColumnMeta {
+    def apply(jdbcType: JdbcType, vendorTypeName: String, nullability: Nullability, name: String): ColumnMeta = {
+      new ColumnMeta(tweakJdbcType(jdbcType, vendorTypeName), vendorTypeName, nullability, name)
+    }
+  }
 
   /** Metadata for the JDBC end of a column/parameter mapping. */
   final case class ParameterMeta(jdbcType: JdbcType, vendorTypeName: String, nullability: Nullability, mode: ParameterMode)
+  object ParameterMeta {
+    def apply(jdbcType: JdbcType, vendorTypeName: String, nullability: Nullability, mode: ParameterMode): ParameterMeta = {
+      new ParameterMeta(tweakJdbcType(jdbcType, vendorTypeName), vendorTypeName, nullability, mode)
+    }
+  }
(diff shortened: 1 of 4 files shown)


guymers:
Would resolve #1481

modules/core/src/main/scala/doobie/util/analysis.scala
```diff
@@ -20,9 +20,28 @@ object analysis {
...
+      new ParameterMeta(tweakJdbcType(jdbcType, vendorTypeName), vendorTypeName, nullability, mode)
+    }
+  }
+
+  private def tweakJdbcType(jdbcType: JdbcType, vendorTypeName: String) = jdbcType match {
+    // the Postgres driver does not return *WithTimezone types but they are pretty much required for proper analysis
+    // https://github.com/pgjdbc/pgjdbc/issues/2485
+    // https://github.com/pgjdbc/pgjdbc/issues/1766
+    case JdbcType.Time if vendorTypeName.compareToIgnoreCase("timetz") == 0 => JdbcType.TimeWithTimezone
+    case JdbcType.Timestamp if vendorTypeName.compareToIgnoreCase("timestamptz") == 0 => JdbcType.TimestampWithTimezone
```
sideeffffect, line 42: or `"TIMESTAMP WITH TIMEZONE"`?

sideeffffect (approved):
Looks great! Thank you :+1:

modules/core/src/main/scala/doobie/util/analysis.scala
guymers: Are there JDBC drivers that return a type of `Timestamp` but a vendor type name of `TIMESTAMP WITH TIMEZONE`?

sideeffffect: What I meant was like this
```scala
    case JdbcType.Timestamp if vendorTypeName.compareToIgnoreCase("timestamptz") == 0 || vendorTypeName.compareToIgnoreCase("TIMESTAMP WITH TIMEZONE") == 0 => ...
```

guymers: Is `case JdbcType.Timestamp if vendorTypeName.compareToIgnoreCase("TIMESTAMP WITH TIMEZONE") == 0` ever true though?

Postgres returns `timestamptz`, h2 returns `JdbcType.TimestampWithTimezone`, mySQL doesn't have a timestamp with timezone column type

sideeffffect: Fair enough :+1:

jatcwang (approved):

jatcwang:
Thanks!

PR #1739, opened 2022-09-07
Add Fragments.in with 3 parameters

vincentdehaan:
This PR adds a `Fragments.in` variant for tuples of three elements. Hence, we can now do something like this:

```
WHERE (col1, col2, col3) IN ((1,2,3), (2,3,4))
```


jatcwang:
I think this is superceded by #2128! Thanks

PR #1742, opened 2022-09-17
Increase check test time resolution

guymers:
When checking if time related columns are written and read, use the highest resolution supported by the underlying database.


jatcwang:
Thanks!

PR #1743, opened 2022-09-27
Bump HikariCP version to 5

Fristi:


build.sbt
```diff
@@ -326,8 +331,7 @@ lazy val hikari = project
     name := "doobie-hikari",
     description := "Hikari support for doobie.",
     libraryDependencies ++= Seq(
-      //needs to be excluded, otherwise coursier may resolve slf4j-api 2 if > Java 11
-      "com.zaxxer"     % "HikariCP"   % hikariVersion exclude("org.slf4j", "slf4j-api"),
+      "com.zaxxer"     % "HikariCP"   % hikariVersion,
```
jatcwang, line 334: I'll test whether adding back `exclude("org.slf4j", "slf4j-api")` is fine.

Fristi: Let me know, I can add it back if you want

PR #1745, opened 2022-10-07
Refactor Hikari config

sideeffffect:
Technically a breaking change, but very small and it's worth the simplification and we're still before 1.0.0 proper :crossed_fingers:


jatcwang (approved):

jatcwang:
Thanks I'll make a note of it in the release notes

Issue #1763, opened 2022-11-09
Update query check test fails


takapi327:
Hi there 👋

When checking a query statement created with parameters, the error "Parameter metadata not available for given statement (analysisspec.scala:61)" occurs.

**Version in use**
- Scala: 2.13.7
- doobie: 1.0.0-RC2
- mysql-connector-java: 8.0.26

```scala
"SQL Check" should {
  "Update Test Failure" in {
    val test1 = "Test"
    val test2 = 1
    check(sql"INSERT INTO todo_task (id, title, description, state) VALUES (${Some(1)}, ${test1}, ${Some("Test")}, ${test2})".update)
  }

  "Update Test Success" in {
    check(sql"INSERT INTO todo_task (id, title, description, state) VALUES (1, Test, Test, 1)".update)
  }
}
```

```console
[info]   Update Test Failure
[info]     + Update0 defined at SQLCheckTest.scala:27
[info]       
[info]         INSERT INTO todo_task (id, title, description, state) VALUES (?, ?, ?, ?)
[info]       
[error]       x SQL Compiles and TypeChecks
[error]        Parameter metadata not available for the given statement (analysisspec.scala:61)
[info]   Update Test Success
[info]     + Update0 defined at SQLCheckTest.scala:33
[info]       
[info]         INSERT INTO todo_task (id, title, description, state) VALUES (?, ?, ?, ?)
[info]       
[info]       + SQL Compiles and TypeChecks
```

I believe this is the same error as when a Select statement query with parameters is executed using the check method.

```scala
"Select Test Failure" in {
  val test = 1L
  check(
    sql"SELECT id, title, description, state FROM todo_task WHERE id = $test"
    .query[(Option[Long], String, Option[String], Short)]
  )
}
```

```console
[info]   Select Test Failure
[info]     + Query0[(Option[Long], String, Option[String], Short)] defined at SQLCheckTest.scala:39
[info]       
[info]         SELECT id, title, description, state FROM todo_task WHERE id = ?
[info]       
[error]       x SQL Compiles and TypeChecks
[error]        Parameter metadata not available for the given statement (analysisspec.scala:61)
```

If you are testing a query for a Select statement with a parameter, there is a checkOutput method that can be used and the test will succeed.

```scala
"Select Test Success" in {
  val test = 1L
  checkOutput(
    sql"SELECT id, title, description, state FROM todo_task WHERE id = $test"
      .query[(Option[Long], String, Option[String], Short)]
  )
}
```

[...]

PR #1764, opened 2022-11-10
Correction of check test for update queries

takapi327:
From [1763](https://github.com/tpolecat/doobie/issues/1763)


modules/example/src/main/scala/example/AnalysisTest.scala
```diff
@@ -49,7 +49,7 @@ object AnalysisTest {
     """.query[PGcircle]
   }
 
-  def update(code: String, name: Int): Update0 =
+  def update(name: String, code: String): Update0 =
```
takapi327, line 52: This was not consistent with the table definition itself, so it was corrected and used as a test query.

modules/example/src/test/scala-2/example/AnalysisTestSpecs2.scala
```diff
@@ -22,6 +22,6 @@ class AnalysisTestSpecs2 extends Specification with IOChecker {
   // check(AnalysisTest.arrayTest2)
   check(AnalysisTest.pointTest)
   // check(AnalysisTest.pointTest2)
-  // check(AnalysisTest.update("foo", 42))
+  checkOutput(AnalysisTest.update("foo", "bkah"))
```
takapi327, line 25: I wanted to test on [IOCheckerCheck](https://github.com/tpolecat/doobie/blob/main/modules/specs2/src/test/scala/doobie/specs2/CheckerTests.scala#L30), but this one seemed to be set up to test only simple queries, so I am testing on the example project.

jatcwang: It's ok :) I think at some point we need to realign all the tests (scalatest/munit/weaver etc etc) so they test all the scenarios

jatcwang:
Thank you! :100:

PR #1788, opened 2022-12-21
upgrade to Scala 3.2.1/2.13.10/2.12.17 + fs2 3.4.0

jatcwang:


build.sbt
```diff
@@ -3,32 +3,32 @@ import sbt.dsl.LinterLevel.Ignore
...
+lazy val scala3Version       = "3.2.1"
 lazy val slf4jVersion         = "1.7.36"
 lazy val weaverVersion        = "0.7.15"
 
 // Basic versioning and publishing stuff
 ThisBuild / tlBaseVersion := "1.0"
 ThisBuild / tlCiReleaseBranches := Seq("main") // publish snapshits on `main`
-ThisBuild / scalaVersion := scala213Version
-ThisBuild / crossScalaVersions := Seq(scala212Version, scala213Version, scala30Version)
+ThisBuild / scalaVersion := scala212Version
```
sideeffffect, line 30: Why change to 2.12 as the default?

sideeffffect (commented):
Need to run `githubWorkflowGenerate`

build.sbt
jatcwang: Ah was checking some errors in 2.12. Will revert thanks!

PR #1789, opened 2022-12-21
Improve Java time meta experience

guymers:
Provide Java time meta instances without requiring any imports. These instances follow the JDBC specification and map a single column type to a java.time class.

Any instances for `ZonedDateTime` have been removed. Storing one requires the `LocalDateTime` and `ZoneId` to be stored, it cannot safely be put into an `OffsetDateTime`.


sideeffffect (approved):
Looks great! Nice work :bow:

jatcwang:
Thanks for the PR. I'll need to double check that this works for MySQL too (though we don't official support it I'd like to avoid breaking any of their typecheck tests)

Specifically from my reading, in MySQL their `TIMESTAMP` type works equivalently to PostgreSQL's `TIMESTAMPTZ` when it comes to JDBC behaviour (Semantically a "instant in time" and supports mapping to OffsetDateTime in the JDBC driver)

guymers:
Here's some MySQL time instances https://github.com/tpolecat/doobie/pull/1791.

Note that I have never used MySQL.

PR #1791, opened 2022-12-25
MySQL Java time instances

guymers:


build.sbt
```diff
@@ -266,7 +280,7 @@ lazy val postgres = project
       import doobie._, doobie.implicits._
       import doobie.postgres._, doobie.postgres.implicits._
       implicit val cs = IO.contextShift(scala.concurrent.ExecutionContext.global)
-      val xa = Transactor.fromDriverManager[IO]("org.postgresql.Driver", "jdbc:postgresql:world", "postgres", "")
+      val xa = Transactor.fromDriverManager[IO]("org.postgresql.Driver", "jdbc:postgresql:world", "postgres", "password")
```
jatcwang, line 283: Not opposed to this change but can you split this out to a separate PR for me? :slightly_smiling_face:

guymers: https://github.com/tpolecat/doobie/pull/1792

modules/mysql/src/main/scala/doobie/mysql/JavaTimeInstances.scala
```diff
@@ -0,0 +1,53 @@
...
+      List(JT.VarChar, JT.Date, JT.Time),
+      classOf[java.time.OffsetDateTime]
+    )
+
+  implicit val JavaTimeInstantMeta: Meta[java.time.Instant] =
+    JavaTimeOffsetDateTimeMeta.timap(_.toInstant)(OffsetDateTime.ofInstant(_, ZoneOffset.UTC))
+
+  implicit val JavaTimeLocalDateTimeMeta: Meta[java.time.LocalDateTime] =
+    Basic.oneObject(
+      JT.Timestamp,
```
sideeffffect, line 28: MySQL/MariaDB has only `TIMESTAMP` which behaves as if `WITH TIME ZONE`, right? It is the equivalent of Java's `Instant`.

So that would mean that `LocalDateTime` is not suitable for `TIMESTAMP`, right?

modules/mysql/src/main/scala/doobie/mysql/JavaTimeInstances.scala
```diff
@@ -0,0 +1,53 @@
...
+trait JavaTimeInstances extends MetaConstructors {
+
+  implicit val JavaTimeOffsetDateTimeMeta: Meta[java.time.OffsetDateTime] =
+    Basic.oneObject(
+      JT.Timestamp,
+      List(JT.VarChar, JT.Date, JT.Time),
+      classOf[java.time.OffsetDateTime]
+    )
+
+  implicit val JavaTimeInstantMeta: Meta[java.time.Instant] =
```
sideeffffect, line 23: Shouldn't this be the "primary" Meta instance and the one for `OffsetDateTime` be derived from it?

modules/mysql/src/main/scala/doobie/mysql/JavaTimeInstances.scala
guymers: Yeah those are pretty wrong, should be fixed now.

modules/mysql/src/main/scala/doobie/mysql/JavaTimeInstances.scala
guymers: `Instant` isn't mentioned in the JDBC spec so drivers don't have to provide a way to convert a column to/from it. The MySQL driver errors if you ask it for an `Instant` so it is implemented in terms of `OffsetDateTime`.

sideeffffect (approved):
:heart:

sideeffffect:
@guymers why did you close this? Is it superseded by https://github.com/tpolecat/doobie/pull/1789 ?

guymers:
Its been 5 months and I don't use mysql.

jatcwang:
I'll get this through the finish line. Thanks

sideeffffect:
@guymers should this be merged before or after https://github.com/tpolecat/doobie/pull/1789 ?

guymers:
Doesn't really matter, but may as well do this one.

jatcwang:
Superceded by #2024 after #2023 allowing more precise typechecking based on DB vendor's type name.
@guymers I've added you as a coauthor in the commit. Thanks!

PR #1792, opened 2022-12-26
Use offical postgis docker image

guymers:
A password is now required when connecting to the container.


sideeffffect (approved):
:heart:

jatcwang:
Cheerios!

PR #1842, opened 2023-04-15
Ensure backward compatibility of Hikari Config

sideeffffect:
/cc @jatcwang


jatcwang:
Thanks @sideeffffect I'm not too sure what motivated this change given that we're breaking back compat in 1.0?
When you're talking about backwards compatibility do you mean binary back compat?

sideeffffect:
The point is to enable backward compatible evolution of `Config`. When we would potentially add a new field to `Config`, we don't want to break binary compatibility.

Binary compatibility with case classes is a tricky topic in Scala https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html#evolving-code-without-breaking-binary-compatibility

jatcwang:
If you want to keep backwards binary compatibility you'll need to use package private I think. Using private makes it unavailable at the bytecode level.

sideeffffect:
Are you referring to the access modifier of the constructor of `Config`?
Making it `private` doesn't make it private in JVM bytecode, it is actually still publicly accessible in JVM bytecode. But it makes it inaccessible from Scala code. And that is good enough for keeping backward compatibility with our Scala users.

jatcwang:
I've added a demo of making method private breaking bin compat https://github.com/jatcwang/binary-compatibility-guide
```
[error] Test suite IncompatibilityShowcases failed with java.lang.IllegalAccessError: class com.example.LibA$ tried to access private method 'int all_incompat.MakeMethodPrivate$.willBePrivate()'
```
You can see that projects like cats use package private to maintain bin compat.

sideeffffect:
> making method private breaking bin compat

Of course it breaks :smiley: There's probably a misunderstanding. I know the changes that I'm proposing in this PR are breaking bin compatibility. I'm introducing them so that once we get to `1.0.0` we can evolve `Config` in a bin compatible way.

modules/hikari/src/main/scala/doobie/hikari/Config.scala
```diff
@@ -103,4 +138,92 @@ object Config {
...
+    initializationFailTimeout,
+    isolateInternalQueries,
+    readOnly,
+    registerMbeans,
+    schema,
+    transactionIsolation
+  )
+
+  @nowarn
+  def fromProduct(p: Product): Config = p.productArity match {
```
jatcwang, line 196: Sorry not sure why we need this?

sideeffffect: Scala 3 defines `fromProduct` in companion object of case classes. If you add a new field to the case class, you need to add a new case to `fromProduct` to ensure compatibility with old code which counts with the old number of fields.
I've improved the comment on top of the case class.

sideeffffect:
@jatcwang so what do you think? Can we merge this?

jatcwang:
Will need some time @sideeffffect. I ran javap on a Scala 3 compiled case class but still couldn't find `fromProduct` so I'm a bit confused. Will get this in for RC3

sideeffffect:
Did you look into the companion object?

sideeffffect:
@jatcwang do you think we could merge this PR?

sideeffffect:
@jatcwang ping :smile_cat:

Issue #1821, opened 2023-02-06
Remove LogHandler on Query/Update and allow setting a String label instead


jatcwang:
In #1566 we now have LogHandlerM that is setup on an Transactor.

To avoid confusion we should remove LogHandler current set on individual Query/Update. In replacement, we want to allow adding a `label: String` to each Query/Update, which the Transactor's LogHandlerM can then use that to decide how it logs the Query/Update when it's executed.

PR #1849, opened 2023-05-11
Replace logHandler with label in Query and Update

george-wilson-rea:
I really want `RC3` to be released, so I've had a go at #1821 . What do you think?

* I've used "unlabaled" in a case where the user does't provide a label. Do you like that, or would you like empty string or something else? Maybe `Option[String]`?
* I've added the label, but I don't understand what to do with it in the transactor.


jatcwang (approved):
Thanks @george-wilson-rea that's exactly what I have in mind :)
I don't think we're actually logging it atm right? That can be added in a follow up PR if you want to take a stab at it, or leave it and I'll get to it 👍

george-wilson-rea:
@jatcwang I'm not logging it anywhere. I don't really understand where to log it. The issue says somewhere in the `Transactor` but where? Admittedly I don't really understand how doobie is implemented; I just followed the types to make this change.

If you could tell me where and how to do the logging, I'm happy to do it as part of this pull request. I also want to move `"unlabeled"` out to a constant somewhere, rather than duplicating the string literal in several places.

It looks like the build failed; should I be worried about that?

Also I notice that a bunch of the scala steward PRs are still open after quite some time. Can you merge them?

george-wilson-rea:
Looks like the build breakage is docs-related. I'll see what I can do...

jatcwang:
Thanks! If you want to take a stab at it, you can modify `doobie.util.log.LogEvent` to include `label` and follow the type errors.
(You should eventually modify e.g. `doobie.util.query.Query#executeQuery`)
I recommend adding it behind the  `args` parameter :)

george-wilson-rea:
Ok, I've tried logging the label. See what you think. I still need to fix the docs.

george-wilson-rea:
Hopefully the tests will pass now. I don't know what to update the docs with because I don't understand where the LogHandler gets set now.

jatcwang:
I'll push some commits to update the docs. Thanks for your PR!

hamnis:
maybe the label should be a newtype surrounding string, as having raw strings as implicit parameters feels bad.

jatcwang:
@hamnis I agree but the label doesn't seem to be an implicit parameter atm?

hamnis:
@jatcwang it is an implicit here: https://github.com/tpolecat/doobie/pull/1849/files#diff-97913cd15deed5b279aee8307f32609d81658d2af5abfccbd53005e58b960bb4R206

that seems to be the only instance though

jatcwang:
@hamnis Good catch thank you. Fixed :)

Issue #1850, opened 2023-05-11
Revisit fragment helper (and/or) behavior when provided collection is empty


jatcwang:
Thinking of revisiting the behavior of fragment helper modified in #1553.

For `and`/`or`, I still think that defaulting to true / false when provided collection is empty will be a footgun. 

I think a combination of andOpt (and perhaps a version of andOpt that takes a empty-able collection) will handle the usecase sufficiently?

cc @sideeffffect

sideeffffect:
Let me explain here why I implemented it that way and why it makes sense:

For `AND`, the _neutral element_ is `true`, because `x AND true` = `x`.
For `OR`, the _neutral element_ is `false`, because `x OR false` = `x`.

And it is natural to substitute the _neutral element_ (or sometimes also called _identity element_) for `List.empty`.

So for example it's perfectly fine for `or(List.empty)` to evaluate to `false`. If there isn't even a single true thing in `x OR y OR z OR ...`, then it should be `false`. That's how most people think of using OR.

jatcwang:
I guess I can think of multiple scenarios that the silent behavior is not desirable.

Example: Imagine some sort of list view which
* Always filters the items based on the user's role 
* Has zero or more filters that you can apply, and will show items that matches any of the filters. A real world example of this is shopping website filters

If we have one filter active then it works:

```
val filters = List(..)
and(fr"user_role = ${role}", or(filters)) -- user_role = ? AND (cond1 OR cond2 OR ...)
```

However in the case where there are no filters, defaulting to FALSE leads to the wrong behavior `user_role = ? AND FALSE` -- Now we're returning no rows.
Similar issues occur for AND conditions defaulting to TRUE.

I am generally wary of silent behaviors like these that makes the code "clean" but lead to bugs in edge case scenarios (or when the surrounding code changes).

--- 

What I propose:

- Change `andOpt`/`orOpt` return `Option[Fragment]` so they "disappear" if no parameters are provided or all are None.
- Add a version of `andOpt`/`orOpt` with signature `andOpt[F[_]: Foldable]: Option[Fragment]`
- Change `and[F[_]: Foldable](fs: F[Fragment])`  to `and[F[_]: Reducible](fs: F[Fragment])`
[...]

PR #1875, opened 2023-06-17
Add andFallbackFalse and orFallbackTrue fragment helpers

sideeffffect:
closes https://github.com/tpolecat/doobie/issues/1850

From the example of @jatcwang  https://github.com/tpolecat/doobie/issues/1850#issuecomment-1546765160
it would be used like this:
```scala
  val role = 1
  val filters = List(fr"") // but could be empty
  and(fr"user_role = ${role}", orFallbackTrue(filters))
```

### DIFF
--- a/modules/core/src/main/scala/doobie/util/fragments.scala
+++ b/modules/core/src/main/scala/doobie/util/fragments.scala
@@ -53,6 +53,12 @@ object fragments {
     case None => fr"${true}"
   }
 
+  /** Returns `(f1 AND f2 AND ... fn)`, or `false` for empty `fs`. */
+  def andFallbackFalse[F[_]: Foldable](fs: F[Fragment]): Fragment = fs.toList.toNel match {
+    case Some(fs) => parentheses(fs.intercalate(fr"AND"))
+    case None => fr"${false}"
+  }
+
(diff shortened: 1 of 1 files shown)


modules/core/src/main/scala/doobie/util/fragments.scala
```diff
@@ -53,6 +53,12 @@ object fragments {
     case None => fr"${true}"
   }
 
+  /** Returns `(f1 AND f2 AND ... fn)`, or `false` for empty `fs`. */
+  def andFallbackFalse[F[_]: Foldable](fs: F[Fragment]): Fragment = fs.toList.toNel match {
+    case Some(fs) => parentheses(fs.intercalate(fr"AND"))
+    case None => fr"${false}"
```
jatcwang, line 59: I think I'd prefer this to be a constant i.e. fr"FALSE"

jatcwang:
Thanks @sideeffffect I've recently merged in #1876 and this will work well with that.

jatcwang:
I'll work on adding this in

jatcwang:
Superceded by #1875. Let me know if you want to change fr"FALSE" to fr"${false}"

PR #1881, opened 2023-06-18
Simplify fragments

sideeffffect:
I've rethought the recent changes we've made in recent time to `fragments`. And I think I have good news: I think we can simplify `fragments` quite a lot.
 * We don't need to artificially constrain the use only to `Reducible`s. Most of the combinators work sensibly even for the empty case, i.e. `Foldable`s
 * The problem with the default value on empty case we can solve with a parameter with default value
 * I've used custom `Fragment` combinators that returned `Option[Fragment]`, and they have been nothing but pain to work with. It's so so much better design to _always_ return just `Fragment` from `Fragment` combinators -- it's lower complexity and easier to work with for users.

### DIFF
--- a/modules/core/src/test/scala/doobie/util/FragmentsSuite.scala
+++ b/modules/core/src/test/scala/doobie/util/FragmentsSuite.scala
@@ -15,9 +15,11 @@ class FragmentsSuite extends munit.FunSuite {
   import cats.effect.unsafe.implicits.global
 
   val xa = Transactor.fromDriverManager[IO](
-    "org.h2.Driver",
-    "jdbc:h2:mem:queryspec;DB_CLOSE_DELAY=-1",
-    "sa", ""
+    driver = "org.h2.Driver",
+    url = "jdbc:h2:mem:queryspec;DB_CLOSE_DELAY=-1",
+    user = "sa",
+    password = "",
+    logHandler = None
   )
 
   val nel  = List(1,2,3).toNel.getOrElse(sys.error("unpossible"))
--- a/modules/core/src/test/scala/doobie/util/FragmentsSuite.scala
+++ b/modules/core/src/test/scala/doobie/util/FragmentsSuite.scala
@@ -20,253 +20,124 @@ class FragmentsSuite extends munit.FunSuite {
     "sa", ""
   )
 
-  val nelInt  = NonEmptyList.of(1,2,3)
-  val listInt = nelInt.toList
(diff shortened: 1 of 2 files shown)


modules/core/src/test/scala/doobie/util/FragmentsSuite.scala
```diff
@@ -17,276 +17,131 @@ class FragmentsSuite extends munit.FunSuite {
...
-  test("orderByOpt (Foldable empty) ") {
-    assertEquals(orderByOpt(List.empty[Fragment]).query[Unit].sql, "")
-  }
-
-  test("orderByOpt (Foldable many) ") {
-    assertEquals(orderByOpt(List(fr0"a", fr0"b")).query[Unit].sql, "ORDER BY a, b")
-  }
-  
-  test("Usage test: whereAndOpt") {
-    assertEquals(whereAndOpt(Some(fr"hi"), orOpt(List.empty[Fragment]), orOpt(List(fr"a", fr"b"))).query[Unit].sql, "WHERE hi AND (a OR b ) ")
```
jatcwang, line 287: Do you think given this example is painful to work with? I feel like this is alright?

guymers:
Can we just revert back to what was in 1.0.0-RC2? It was simple, one method for each and no boolean parameters with default values...

modules/core/src/test/scala/doobie/util/FragmentsSuite.scala
sideeffffect: Or course, this small example is fine as such. But in more complex queries, the options are more annoying to work with.

With this PR taken into account, you can express the same idea like this:
```scala
whereAnd(Some(fr"hi"), orOpt(List.empty[Option[Fragment]], onEmpty = true), or(List(fr"a", fr"b")))
```

sideeffffect:
One specific example that I've just come across where returning `Option[Fragment]` makes live unnecessarily difficult:
```scala
sql"""
  ...
  ${Fragments.andOpt(emailIsNull, substring)}
  ...
"""
```
You can't interpolate this fragment. It fails, because the result is of `andOpt` is `Option[Fragment]`. It would work only if it were a mere `Fragment`.

guymers:
`${Fragments.andOpt(emailIsNull, substring).getOrElse(sql"TRUE")}`
or 
 `${Fragments.andOpt(emailIsNull, substring).orEmpty}`
depending on the context?

PR #1890, opened 2023-07-06
Fix NPE in Scala 3 derivation for Write

jacektomaszek:
#1816 fixed NPE for Read instances, but bug persisted for Write instances.

@jatcwang mentioned that recursive definitions for Read do not exist, if that's also the case for Write we can make the same change as in #1816 but for Write.

5d52836 commit introduces test that succeeds for Read, but fails for Write. Next commit introduces the mentioned change that makes the Write test also pass.

### DIFF
--- a/modules/core/src/main/scala/doobie/util/write.scala
+++ b/modules/core/src/main/scala/doobie/util/write.scala
@@ -10,7 +10,36 @@ import doobie.free.{ FPS, FRS, PreparedStatementIO, ResultSetIO }
 import java.sql.{ PreparedStatement, ResultSet }
 import doobie.util.fragment.Fragment
 import doobie.util.fragment.Elem
+import scala.annotation.implicitNotFound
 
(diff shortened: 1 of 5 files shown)


modules/core/src/main/scala/doobie/util/write.scala
```diff
@@ -10,7 +10,36 @@ import doobie.free.{ FPS, FRS, PreparedStatementIO, ResultSetIO }
...
+
+  scala> Write[Foo]
+
+and similarly with Put:
+
+  scala> Put[Foo]
+
+And find the missing instance and construct it as needed. Refer to Chapter 12
+of the book of doobie for more information.
+""")
```
jatcwang, line 42: This is amazing. Thank you! 💯

jatcwang:
Oopsie my bad I should've fixed Write too. Thanks for this!

PR #1896, opened 2023-07-18
Add module for log4cats support for debugging purposes

sideeffffect:
Creates the `Log4CatsDebuggingLogHandler` class

### DIFF
--- a/modules/log4cats/src/main/scala/doobie/log4cats/Log4CatsLogHandler.scala
+++ b/modules/log4cats/src/main/scala/doobie/log4cats/Log4CatsLogHandler.scala
@@ -1,0 +1,48 @@
+// Copyright (c) 2013-2020 Rob Norris and Contributors
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package doobie.log4cats
+
+import doobie.LogHandler
+import doobie.util.log._
+import org.typelevel.log4cats._
+
+class Log4CatsLogHandler[F[_]](logger: MessageLogger[F]) extends LogHandler[F]{
+  override def run(logEvent: LogEvent): F[Unit] = logEvent match {
+    case Success(s, a, l, e1, e2) =>
+      logger.info(
+        s"""Successful Statement Execution:
+           |
+           |  ${s.linesIterator.dropWhile(_.trim.isEmpty).mkString("\n  ")}
+           |
+           | arguments = [${a.mkString(", ")}]
+           | label     = $l
+           |   elapsed = ${e1.toMillis.toString} ms exec + ${e2.toMillis.toString} ms processing (${(e1 + e2).toMillis.toString} ms total)
+        """.stripMargin)
+
+    case ProcessingFailure(s, a, l, e1, e2, t) =>
+      logger.warn(
+        s"""Failed Resultset Processing:
+           |
+           |  ${s.linesIterator.dropWhile(_.trim.isEmpty).mkString("\n  ")}
+           |
+           | arguments = [${a.mkString(", ")}]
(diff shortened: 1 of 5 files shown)


modules/log4cats/src/main/scala/doobie/log4cats/Log4CatsLogHandler.scala
```diff
@@ -0,0 +1,48 @@
...
+
+class Log4CatsLogHandler[F[_]](logger: MessageLogger[F]) extends LogHandler[F]{
+  override def run(logEvent: LogEvent): F[Unit] = logEvent match {
+    case Success(s, a, l, e1, e2) =>
+      logger.info(
+        s"""Successful Statement Execution:
+           |
+           |  ${s.linesIterator.dropWhile(_.trim.isEmpty).mkString("\n  ")}
+           |
+           | arguments = [${a.mkString(", ")}]
```
jatcwang, line 19: I think logging arguments by default would be a security violation in many workplaces. Perhaps we can make this an option (default false)

modules/log4cats/src/main/scala/doobie/log4cats/Log4CatsLogHandler.scala
```diff
@@ -0,0 +1,48 @@
...
+           | label     = $l
+           |   elapsed = ${e1.toMillis.toString} ms exec + ${e2.toMillis.toString} ms processing (${(e1 + e2).toMillis.toString} ms total)
+        """.stripMargin)
+
+    case ProcessingFailure(s, a, l, e1, e2, t) =>
+      logger.warn(
```
jatcwang, line 25: I think this should be an error too? (e.g. error in Put/Get definition)

modules/log4cats/src/main/scala/doobie/log4cats/Log4CatsLogHandler.scala
sideeffffect: I just copied whatever `jdkLogHandler` does.
https://github.com/tpolecat/doobie/blob/main/modules/free/src/main/scala/doobie/util/log.scala#L66

jatcwang: jdkLogHandler is not for production use (as written in the scaladoc). Although I think we can update that scaladoc too to provide additional info

modules/log4cats/src/main/scala/doobie/log4cats/Log4CatsLogHandler.scala
sideeffffect: It seemed to me not as severe like `ExecFailure`, and giving it a different priority also give more power to tune what (and how) is logged.

But if you'd still prefer to have both as errors, no problem, just let me know.

modules/log4cats/src/main/scala/doobie/log4cats/Log4CatsLogHandler.scala
sideeffffect: This class would still be very valuable for debugging purposes in non-production environments.
Indeed it is true that it could log sensitive data.

modules/log4cats/src/main/scala/doobie/log4cats/Log4CatsLogHandler.scala
jatcwang: Thinking about it more, perhaps both these failure cases should be `warn` level because 

- If the failure isn't intentional, their code is going to fail anyway and their execution will fail with an Exception anwyay. 
- If the (Execution) failure is intentional, then we don't really want to log at error level to avoid triggering alerts.

WDYT? I don't really mind

PR #1906, opened 2023-08-17
Use `syncStep` in `WeakAsync.liftK`

armanbilge:
Really interesting brainstorm with @TimWSpence. We can avoid the overhead of submitting to `Dispatcher`, starting a fiber, creating an async boundary, etc. by first attempting to `syncStep` i.e. interpret `F[_]` directly into `G[_]`. If the given effect is only using `Sync` ops then it can be interpreted to entirety without needing the `Dispatcher`. If not, the `Dispatcher` serves as a fallback.


jatcwang (approved):
Cool stuff TIL!

Issue #1902, opened 2023-08-10
Transactor constructors in 2 effects


lgmyrek:
With the introduction of LogHandler[M] initializing Transactor[M] and then transforming it into Transactor[M0] in some contextual M0 is no longer as feasible as we want LogHandler to be in M0.

so it would be nice to have sth like:

```scala
  def fromHikariConfig2K[
    M[_]: Sync,
    M0[_]: Async
  ](
    hikariConfig: HikariConfig,
    logHandler: Option[LogHandler[M0]] = None
  ): Resource[M, HikariTransactor[M0]] =
  for {
    _ <- Sync[M].delay(hikariConfig.validate()).toResource
    
    ...
```

jatcwang:
I mostly use concrete effect types so it didn't cross my mind but happy to review a PR with this change :)

satorg:
Hi @lgmyrek , I wonder if the `mapK` method in `Transactor` could help you with that:
```scala
HikariTransactor.fromHikariConfig[M](...).map(_.mapK(liftMtoM0))
```
..if I'm not missing something from your question..

lgmyrek:
@jatcwang sure, any naming preferences? 😄 

@satorg it did help me when there was no LogHandler, let's say M0 has request-id, spans, and other contextual things, the only way I can make use of those in LogHandler is when LogHandler is impelemnted being aware of those.

I could initialize the transactor in M0 and mapK the resource to M, but that would mean I need to provide some `fake context` and that is quite hacky

satorg:
@lgmyrek , thank you for the explanation. I think I got your case. I feel you could still make use of `mapK`, even though in a slightly different way:
1. construct you `Transactor` in `M` as usual without any `LogHandler` attached:
    ```transM <- fromHikariConfig[M](..., logHandler = None)```
2. at the point where you have your `M0` ready (along with your `LogHandler[M0]`),
   - lift your `Transactor[M]` into `M0`:
     ```
      val transM0 = transM.mapK(liftMtoM0)
      ```
   - then inject your `LogHandler[M0]` into `transM0`:
       ```
       val logHandler: LogHandler[M0] = ???
       val transM0final: Transactor[M0] = transM0.copy(interpret0 = KleisliInterpreter[M0](logHandler).ConnectionInterpreter)
       ```
Yeah, I know – looks a bit hairy :)

But perhaps it would be way easier if we would have a method in `Transactor` like
```
[...]

PR #1939, opened 2023-11-04
Allow for different effects of HikariTransactor and its creation

sideeffffect:
/cc @piton4k
closes https://github.com/tpolecat/doobie/issues/1902

### DIFF
--- a/modules/docs/src/main/mdoc/docs/14-Managing-Connections.md
+++ b/modules/docs/src/main/mdoc/docs/14-Managing-Connections.md
@@ -74,7 +74,7 @@ object HikariApp extends IOApp {
         config.setPassword("")
         config
       }
-      xa <- HikariTransactor.fromHikariConfig[IO](hikariConfig)
+      xa <- HikariTransactor.fromHikariConfig[IO, IO](hikariConfig)
     } yield xa
 
 
--- a/modules/hikari/src/main/scala/doobie/hikari/HikariTransactor.scala
+++ b/modules/hikari/src/main/scala/doobie/hikari/HikariTransactor.scala
@@ -75,6 +75,9 @@ object HikariTransactor {
 
   /** Resource yielding a new `HikariTransactor` configured with the given Config.
    * The `connectEC` is created automatically, with the same size as the Hikari pool.
+   *
+   * @tparam M0 the effect to create a [[HikariTransactor]]
+   * @tparam M the effect under which the [[HikariTransactor]] runs
    */
   def fromConfig[M0[_]: Sync, M[_]: Async](
       config: Config,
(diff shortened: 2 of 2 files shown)


jatcwang:
This is related to #1902 is that right?

sideeffffect:
Oh yes, it is 👍
/cc @lgmyrek

sideeffffect:
@jatcwang :pray:

jatcwang:
I want to test the use case to get a better understanding. 
A nitpick for now is that `Resource[M, HikariTransactor[N]]` is ideally `Resource[N, HikariTransactor[M]]` as most people are used to seeing `Transactor[M]`.

satorg:
May I add my two cents on this please. I wonder if it was better to create a separate set of functions with two effects instead of introducing breaking changes to the existing ones. Then the existing function could be kept compatible but their implementation could be changed to call the new more generic functions. The reasoning is that I guess likely 90-95% users (or so) do not need the effect separation. I believe it is a good practice in general – to provide simplified functions/methods for most common usages.

jatcwang:
Yes I agree that'd be great @satorg :)

sideeffffect:
> most people are used to seeing `Transactor[M]`

Good idea! I've renamed the other type variable to `M0`.

> provide simplified functions/methods for most common usages

It's not worth it. You can always just write
```scala
HikariTransactor.fromHikariConfig[IO, IO](hikariConfig)
```
instead of `fromHikariConfig[IO]`. It's a trivial change (that you typically need to make just once in your app). Duplicating all the functions just isn't worth the small benefit this would bring.

satorg:
> It's a trivial change (that you typically need to make just once in your app). Duplicating all the functions just isn't worth the small benefit this would bring.

I totally agree that it is not that big deal. However, on the flip side changing existing functions
a. introduces a breaking change urging everyone to take an action upon the next upgrade;
b. introduces the new syntax which the majority of users do not care about.

Needless to say that adding a couple of new functions to the library is not a big deal either. No code duplication necessary there – more common used functions can simply call to more generic but relatively rare used ones. In other words: either do it just once in the library and that's it _versus_ everyone does just once but in each project that uses the library.

That said, I realize that it is absolutely matter of perception which of the two sides outweighs.

sideeffffect:
> No code duplication necessary there

Unfortunately it is a lot of code duplication. Majority of the code would be the long function signatures (which you can't reuse in contrast to function bodies). We're in an RC phase, we can afford changes like these.

Of course, if we were stable already, I'd go with the worse, but compatible, way of duplicating the code.

jatcwang:
I'm in general agreement with @satorg here. It's best to uphold backwards compatibility and user-friendliness unless our hands are forced. 

But I can do the suggested change in a follow-up PR, no problem.

PR #1951, opened 2023-11-17
Fix future leak

TimWSpence:
Fix cancelation leak


TimWSpence:
@armanbilge another for you (sorry 😆). That `uncancelable` is necessary right?

armanbilge (approved):
another good catch, this looks right!

jatcwang:
Thanks :)

PR #1961, opened 2023-12-18
Split methods for HikariTransactor constructed with two effects

jatcwang:
Partially revert the change in #1939 so that existing code don't fail to compile. Methods which construct HikariTransactor in a different effect now has the suffix "...SeparateEffect". The original single-effect methods are now implemented in terms of the separate-effect methods.

Related: #1939

### DIFF
--- a/modules/docs/src/main/mdoc/docs/14-Managing-Connections.md
+++ b/modules/docs/src/main/mdoc/docs/14-Managing-Connections.md
@@ -74,7 +74,7 @@ object HikariApp extends IOApp {
         config.setPassword("")
         config
       }
-      xa <- HikariTransactor.fromHikariConfig[IO, IO](hikariConfig)
+      xa <- HikariTransactor.fromHikariConfig[IO](hikariConfig)
     } yield xa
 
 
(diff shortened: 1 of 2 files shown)


jatcwang:
@satorg @sideeffffect Feedback welcome especially the name..it's a bit long :sweat_smile:

sideeffffect (commented):
Maybe `...Generic` instead of `...SeparateEffect`, because that's veeery long?

jatcwang:
Or maybe `...WithResEffect` (With Resource Effect)? :thinking:

satorg:
@jatcwang thank you for taking on it!

I personally don't have a strict preference on the naming.
I could suggest though a conciser suffix `...M2` (e.g. `fromConfigCustomEcM2`). The idea behind such a suffix is the following:
...sometimes effectful methods are marked with the `F` suffix (because of `F[_]`)
...doobie prefers `M[_]` and in this case introduces two M-s – `M` and `M0`, so the overall suffix could be `M2`.

But I totally realize that such a name could look cryptic and for me personally any of the suggested names above will work just as well.

jatcwang:
For this case I don't mind it being a bit longer since it's often only written once in an application. I think `...WithResEffect` is probably the clearest to me. `..Generic` is short but doesn't feel too descriptive of what's happening

sideeffffect (approved):
Let's get this merged

jatcwang:
Thanks all.

PR #1977, opened 2024-01-10
Add withLogHandler method to Transactor

sideeffffect:


satorg (approved):
Although I am not a maintainer of this repository, I would really appreciate if this change could be accepted. This PR introduces a new capability of injecting log-handlers locally, which was not that easy previously.

I believe it is important because without such a capability it is quite tricky to make log messages having their own unique logger names. So that by default all the log messages usually share the same logger name, which is not always that convenient especially in big projects.

That said, I feel it would not hurt if there was a test added for this feature. E.g. `modules/core/src/test/scala/doobie/util/TransactorSuite.scala` seems to be a good place for that.

Issue #533, opened 2017-06-16
withGeneratedKeys & transaction statements missing from logs


Arcovion:
On 0.4.2-SNAPSHOT inserts using `withUniqueGeneratedKeys` aren't being logged:
```scala
sql"INSERT INTO...".update.run
```
works, but this doesn't:
```scala
sql"INSERT INTO...".update.withUniqueGeneratedKeys[Long]("id")
```
I'm using an implicit `LogHandler`

Also, is there any way of logging the BEGIN/COMMIT/ROLLBACK statements? I think it's useful info, makes it easier to parse visually when examining multiple transactions. 
You have no way of knowing which statements have been rolled back otherwise, nor can you see the time it takes to perform a ROLLBACK/COMMIT like you can with other statements.

tpolecat:
Current logging is limited, obviously. There will be good structured logging in an upcoming release.

Kim-SeongCheol:
Is it still not solved?

halfhp:
this would be great to have.  any update?

dpevunov-cp:
Looks like this issue is still relevant and no logs are produced for this case

jatcwang:
I'll close this since it's related to #1888 and can be fixed together

PR #1989, opened 2024-02-04
Add 'scalafmt' to the project

satorg:
This PR:
- Adds **sbt-scalafmt** plugin.
- Creates **.scalafmt.conf**<sup>1</sup>.
- Applies **scalafmtSbt** and **+scalafmtAll** commands<sup>2</sup>.
- Creates **.git-blame-ignore-revs** to exclude commits with auto-reformatting and code auto-generation from **git blame**<sup>3</sup>.
<hr/>

1. The **.scalafmt.conf** mostly contains default settings with a few exceptions. The exceptions are to try minimizing further reformatting. In a case if some specific settings are more preferable for the project, let me know please – I am open to accommodate any configuration changes in here.
2. There's no manual formatting in the PR _except_ the **FreeGen2.scala** file – the latter contains a lot of strings that turn into code and there are quite a lot of pretty long strings among them. Turns out **scalafmt** is not really great in such cases and can end up with weird formatting. Therefore, I added some manual reformatting so that after **scalafmtSbt** the reformatted code still looks pretty decent and readable. All the manually reformatted code is gathered in a single commit: [fix formatting in FreeGen2.scala](https://github.com/tpolecat/doobie/pull/1989/commits/177c2445bcfb4fb2abc24e6bdf1c4ea95767797e).
3. If this branch gets rebased then the content of the **.git-blame-ignore-revs** file has to be updated accordingly.


jatcwang:
Thank you! I'll need to find an opportune time to run this and merge this in (to avoid conflicts in all other PRs)

satorg:
@jatcwang , I've actualized the PR: rebased onto **main**,  updated scalafmt version, re-applied formattings and **.git-blame-ignore-revs**

Let me know if you have any suggestions/concerns please.
Thank you!

jatcwang:
Thanks @satorg. Originally I wanted to merge this after RC6 because I didn't know about .git-blame-ignore-revs. But now I think we can get this in before RC6. I have a change with quite some lines of changes I'm working on so I'll sort this one out after that :pray:

satorg:
@jatcwang sure, np. Let me know please when I should update this PR to synchronize it with the upstream.

Also I think that once this PR gets settled in main, it would make sense to enforce scalafmt checks in CI/CD with a follow-up PR. I think I could take on that too – just let me know when to step in please. Thank you!

jatcwang:
@satorg Agree about the enforcement. Feel free to implement it as part of this PR :)

satorg:
The PR is updated:
- Bump scalafmt up to v3.8.2
- Reformatt with the new version
- add scalafmt checks to the CI workflow

satorg:
@jatcwang , turns out the new Scalafmt version (3.8.2) started reformatting in multiline strings a little more aggressively.
It  resulted in these changes: https://github.com/tpolecat/doobie/pull/1989/commits/d8c3d1ceb74d10899fca7ea7ecdb6b8156a1d9ca

The changes don't break the build but look a bit awkward to me.
If you'd like to, I can either
- leave it as is and therefore add the commit to `.git-blame-ignore-revs` (not there yet), OR
- try to manually improve it (and keep outside of `.git-blame-ignore-revs`).

Wdyt?

jatcwang:
Thanks @satorg. I agree it looks worse but happy to leave it as is, as that code will probably change as part of my work to fix #533

satorg:
Looks like the CI failed because `freeGen2` command updates auto-generated sources...

The problem here is that if I apply `scalafmt` to auto-generated sources and commit the changes (as it is done now in the PR), then `freeGen2` reverts all the formatting and these files appear as uncommitted.

However, if I don't apply formatting to those files, then scalafmt checks will begin failing.

Perhaps the simplest way to resolve this conundrum could be excluding auto-generated files from scalafmt's scope. I'll try to accomplish it shortly.

satorg:
The PR is updated:
1. Disable formatting in auto-generated files via adding `// format: off` to the beginning of each file.
2. Reformat again to apply the changes.

satorg:
@jatcwang, hopefully, the CI issue has been resolved now.

satorg:
@jatcwang , just in case – in regards to the auto-generated sources:

Perhaps, in makes sense to bring in an approach that is used in Cats for that:
https://github.com/typelevel/cats/blob/dd891334505b35e010a12da254924811de838c85/build.sbt#L114

I.e. all the auto-generated sources are written into directories like `.../target/scala-2.13/src_managed/main` and therefore never auto-formatted _nor_ even committed to Git.

I feel, this this PR is ready to go, because making changes to the source generation process would require even more seemingly unrelated changes to this PR.

However, if you'd like such an approach, I could take on it in a follow-up PR.

jatcwang:
Big thanks for sorting this out @satorg!

Issue #936, opened 2019-07-04
support for complex types such as TSRANGE


ShahOdin:
Currently, For whatever reason, Doobie doesn't have out of the box support for types such as `TSRANGE`. To use those with doobie, you have to work with PGObjects directly. I propose adding a type class like:

    trait PGSchema[A] {
      val schemaName: NonEmptyList[String]
      def fromString(s: String): Either[String, A]
      def toString(a: A): String
    }

and a helper function like this:

    def metaForPGSchema[A: TypeTag](implicit pgSchema: PGSchema[A]): Meta[A] = Meta.Advanced
    .other[PGobject](pgSchema.schemaName.head)
      .teimap[A](
      pgO => pgSchema.fromString(pgO.getValue)
    ){
      a =>
        val pgO = new PGobject
        pgO.setType(pgSchema.schemaName.head)
        pgO.setValue(pgSchema.toString(a))
        pgO
    }

Makes the library easier to use.

J0kerPanda:
@ShahOdin , I wonder if you really need a distinct `PGSchema` class. Though it provides a convenient way to define both `Meta[A]` and `Meta[List[A]]`, it doesn't quite fit with `Get` and `Put` doobie already has. With `PGSchema` one can only define a whole `Meta`, meaning having to provide both `fromString` and `toString` functions, while one of them can be unnecessary.

Perhaps we can (similarly to `pgEnumString` in a way) define methods in the `doobie.postgres.Instances` trait to achieve granularity:

- `pgObjectGet[A: TypeTag](schemaTypes: NonEmptyList[String], fromString: String => A): Get[A]`  
- `pgObjectPut[A: TypeTag](schemaTypes: NonEmptyList[String], toString: A => String): Put[A]`  
- `pgObjectArrayGet[A: TypeTag](schemaTypes: NonEmptyList[String], fromString: String => A): Get[List[A]]`  
- `pgObjectArrayPut[A: TypeTag](schemaTypes: NonEmptyList[String], toString: A => String): Put[List[A]]`

(convenience methods):
- `pgObjectMeta[A: TypeTag](schemaTypes: NonEmptyList[String], fromString: String => A, toString: A => String): Meta[A]`
- `pgObjectListMeta[A: TypeTag](schemaTypes: NonEmptyList[String], fromString: String => A, toString: A => String): Meta[List[A]]`

[...]

PR #2008, opened 2024-03-19
PostgreSQL Range Types

tomohavvk:
Hollo community, I would like to introduce my vision for [PostgreSQL](https://www.postgresql.org/docs/current/rangetypes.html) range types implementation in **doobie** since it was requested in this issue but not implemented yet https://github.com/tpolecat/doobie/issues/936

I hope together we can improve this implementation and release the Range Types support into life

---
### Range types

The following range types are supported, and map to **doobie** generic class:

- the `int4range` schema type maps to `Range[Int]`
- the `int8range` schema type maps to `Range[Long]`
- the `numrange`  schema type maps to `Range[Float]` | `Range[Double]` | `Range[BigDecimal]`
- the `daterange` schema type maps to `Range[java.time.LocalDate]`
- the `tsrange`   schema type maps to `Range[java.time.LocalDateTime]`
- the `tstzrange` schema type maps to `Range[java.time.OffsetDateTime]`

```scala mdoc:silent
case class Range[T](lowerBound: Option[T], upperBound: Option[T], edge: Edge)
```

To control the inclusive and exclusive bounds according to the [PostgreSQL](https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-INCLUSIVITY) specification you need to use a special `Edge` enumeration when creating a `Range`:

```scala mdoc:silent
object Edge {
  case object `(_,_)` extends Edge
  case object `(_,_]` extends Edge
  case object `[_,_)` extends Edge
  case object `[_,_]` extends Edge
  case object `empty` extends Edge
}
```

> In the text form of a range, an inclusive lower bound is represented by '[' while an exclusive lower bound is represented by '('. Likewise, an inclusive upper bound is represented by ']', while an exclusive upper bound is represented by ')'.
---
Note: the [range types](https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES) mappings are defined in a different object (`rangeimplicits`). To enable it you must import them explicitly:

```scala mdoc:silent
import doobie.postgres.rangeimplicits._
```

To create for example custom implementation of `Range[Byte]` you can use the public method which declared in the following package `doobie.postgres.rangeimplicits`:

```scala mdoc:silent
def rangeMeta[T](sqlRangeType: String)(implicit D: RangeBoundDecoder[T], E: RangeBoundEncoder[T]): Meta[Range[T]]
```

The main requirements is to need to implement the `RangeBoundDecoder[Byte]` end `RangeBoundEncoder[Byte]` which are essentially:

```scala mdoc:silent
type RangeBoundDecoder[T] = String => T
type RangeBoundEncoder[T] = T => String
```

For a `Range[Byte]`, the meta and bounds encoder and decoder would appear as follows:
```scala mdoc:silent
import doobie.postgres.rangeimplicits._
import doobie.postgres.types.Range.{RangeBoundDecoder, RangeBoundEncoder}

implicit val byteBoundEncoder: RangeBoundEncoder[Byte] = _.toString
implicit val byteBoundDecoder: RangeBoundDecoder[Byte] = _.toByte

implicit val byteRangeMeta: Meta[Range[Byte]] = rangeMeta[Byte]("int4range")

...
val int4rangeWithByteBoundsQuery = sql"select '[-128, 127)'::int4range".query[Range[Byte]]
```


satorg:
Hello @tomohavvk, thank you for your thorough work!
I believe the support for range types will be a fantastic addition to Doobie!

However, I would like to discuss some potential pitfalls I've found out in the PR (if I may).

### 1. Unintentional implicit conversions

Typedefs `RangeBoundEncoder` and `RangeBoundDecoder` followed by implicit function values of those types are nothing but disguised implicit conversion. Therefore if a user imports those implicits via `import doobie.postgres.rangeimplicits._`, the following behavior gets enabled automatically in the scope (scala 2.13.13 repl):

```
scala> type RangeBoundEncoder[T] = T => String
type RangeBoundEncoder

scala> implicit val intBoundEncoder: RangeBoundEncoder[Int] = _.toString
val intBoundEncoder: RangeBoundEncoder[Int] = $Lambda/0x000000012853cc00@3caa4d85

scala> def foo(str: String) = println(s"supposed to be a string: $str")
def foo(str: String): Unit

scala> foo(123) // <-- passing an `Int` into a function that takes `String` !
supposed to be a string: 123
```
I.e. all the provided conversions to and from `String` become implicit, which is definitely not what most of users mean.

Moreover, those implicits are not really necessary out there, compare please:
- version with implicits:
    ```scala
    type RangeBoundEncoder[T] = T => String
    type RangeBoundEncoder[T] = T => String
    implicit val intBoundEncoder: RangeBoundEncoder[Int]     = _.toString
    implicit val longBoundEncoder: RangeBoundEncoder[Long]   = _.toString
    implicit val floatBoundEncoder: RangeBoundEncoder[Float] = _.toString
    implicit val intBoundDecoder: RangeBoundDecoder[Int]     = java.lang.Integer.valueOf(_)
    implicit val longBoundDecoder: RangeBoundDecoder[Long]   = java.lang.Long.valueOf(_)
    implicit val floatBoundDecoder: RangeBoundDecoder[Float] = java.lang.Float.valueOf(_)

    implicit val intRangeMeta: Meta[Range[Int]]     = rangeMeta("int4range")
    implicit val longRangeMeta: Meta[Range[Long]]   = rangeMeta("int8range")
    implicit val floatRangeMeta: Meta[Range[Float]] = rangeMeta("numrange")
    ```
- versus a more straightforward version without conversion implicits:
    ```scala
    implicit val intRangeMeta: Meta[Range[Int]]     = rangeMeta("int4range")(_.toString, java.lang.Integer.valueOf)
    implicit val longRangeMeta: Meta[Range[Long]]   = rangeMeta("int8range")(_.toString, java.lang.Long.valueOf)
    implicit val floatRangeMeta: Meta[Range[Float]] = rangeMeta("numrange")(_.toString, java.lang.Float.valueOf)
    ```
In other words, the implicit conversions there do not seem simplifying anything (nor improve clarity IMO) and in fact require even more lines of code to put everything together eventually.

### 2. Inexact type correspondence

Speaking of the `numrange` type:
```scala
implicit val floatRangeMeta: Meta[Range[Float]]                   = rangeMeta("numrange")
implicit val doubleRangeMeta: Meta[Range[Double]]                 = rangeMeta("numrange")
implicit val bigDecimalRangeMeta: Meta[Range[BigDecimal]]         = rangeMeta("numrange")
```
AFAIK, `numrange` is a built-in range type for the `numeric` scalar. So yes, the best corresponding JVM type is `BigDecimal` indeed. However, providing `Meta` for ranges of `Double` and `Float` by default might not be very good idea – a user may want to create a custom range type for floating-point numbers and then they can run into ambiguous implicit problem in that case (from [here](https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-DEFINING)):
```sql
CREATE TYPE floatrange AS RANGE (subtype = float8, ...);
```
Then somewhere in user code:
```
implicit val doubleRangeMeta: Meta[Range[Double]] = rangeMeta("floatrange")
```
So I would probably suggest to hold on creating too many type shortcuts from the beginning. Ultimately, if there is demand for that from users, it will be easier to add them later (comparing to deprecating and removing stuff in a case of complaints).

### 3. Over-backticking

It is solely my personal perceptions, but does using a lot of back-ticked identifiers in the code really improves it? I feel it may make code more obscure instead. Especially because not all users like to see ASCII-art in their code. For example, `(_,_)` – this one might look a bit obscene, don't you think?

I feel that more straightforward and less fancy naming could be more appreciated, e.g.:
```scala
object Edge {
  case object InclIncl extends Edge
  case object InclExcl extends Edge
  case object ExclIncl extends Edge
  case object ExclExcl extends Edge
  case object Empty extends Edge
}
```
Also, the suggested approach allows to construct a Range like this:
```scala
Range(Some(1), Some(2), Edge.empty)
```
Which I am not sure about if it makes a lot of sense.
In other words, modeling of the `Range` type may deserve additional thought.

tomohavvk:
Hello @satorg, thank you very much for the constructive comments, which have truly improved the implementation. I supported each of your suggestions, and here's what I've at the moment:

### 1. Unintentional implicit conversions
I completely removed any mentions of `RangeBoundEncoder` and `RangeBoundDecoder`. Now, there are no implicits; everything needs to be passed explicitly:
```scala mdoc:silent
def rangeMeta[T](sqlRangeType: String)(encode: T => String, decode: String => T): Meta[Range[T]]
```
### 2. Inexact type correspondence
Done. `numrange` only maps to `Range[BigDecimal]`. We can easily extend it with more numeric types in the future(if needed)

### 3. Over-backticking
> It is solely my personal perceptions, but does using a lot of back-ticked identifiers in the code really improves it? I feel it may make code more obscure instead. Especially because not all users like to see ASCII-art in their code. For example, (_,_) – this one might look a bit obscene, don't you think?

I thought about it :) but it didn't stop me because in my opinion it was the easiest way to understand, when you actually look at those brackets that are in `PostgreSQL` and you don't need to thinking about one more mapping, however, I agree that such code is rare and can be annoying. Therefore, I gladly rewrote it and now no one will see it `(_,_)` except in the mind. Now it looks:
```scala mdoc:silent
object Edge {
  case object ExclExcl extends Edge
  case object ExclIncl extends Edge
  case object InclExcl extends Edge
  case object InclIncl extends Edge
}
```

### 4. Empty range
> Also, the suggested approach allows to construct a Range like this:
`Range(Some(1), Some(2), Edge.empty)`
Which I am not sure about if it makes a lot of sense.
In other words, modeling of the Range type may deserve additional thought.


I redesigned the range type with this in mind, now the range hierarchy looks like this:
```scala mdoc:silent
sealed trait Range[+T]
case object EmptyRange extends Range[Nothing]
case class NonEmptyRange[T](lowerBound: Option[T], upperBound: Option[T], edge: Edge) extends Range[T]
```
We can still create a range like: `NonEmptyRange[Int](None, None, ExclExcl)` which can surprise at first glance, but this is a valid case from PostgreSQL's point of view and we need to support it:
```
    nr     | isempty | lower | upper 
-----------+---------+-------+-------
 (,)       | f       |       |      
 [3,)      | f       |     3 |      
 (,5)      | f       |       |     5
 [1.1,2.2) | f       |   1.1 |   2.2
 empty     | t       |       |      
 [1.7,1.7] | f       |   1.7 |   1.7
```

If there are any suggestions with improvements, I will be glad to consider them.

Issue #1888, opened 2023-06-30
LogHandler does not capture streaming queries or bulk updates


soujiro32167:
Scala: 2.13.11
Doobie: 1.0.0-RC4
Postgres: 14.3

Repro: 

```scala
import cats.effect.{ExitCode, IO, IOApp}
import doobie.implicits.*

object Repro extends IOApp{
  val xa = Transactor.fromDriverManager[IO](
    driver = "org.postgresql.Driver",
    url = "jdbc:postgresql:postgres",
    user = "postgres",
    password = "postgres",
    logHandler = Some(LogHandler.jdkLogHandler)
  )

  def run(args: List[String]): IO[ExitCode] = for {
    _ <- sql"SELECT 1".query[Int].unique.transact(xa) // this gets logged
    _ <- sql"SELECT 1".query[Int].stream.compile.lastOrError.transact(xa) // this doesn't
    _ <- Update[Int]("SELECT ?").updateMany(List(1)).transact(xa) // this doesn't
  } yield ExitCode.Success
}
```

satorg:
I think it is related to (or maybe duplicates) #772.

soujiro32167:
😱

soujiro32167:
from https://github.com/tpolecat/doobie/issues/772#issuecomment-420791244
> It may be worthwhile to go ahead and add a Log constructor to each Free algebra so we can plumb this in from the bottom

It looks like RC3 did exactly that - there is now a `PerformLogging` in every algebra, so it seems. Anything I can help with?

jatcwang:
For `stream` it shouldn't be too hard to plumb the SQL, args and label through. For updateMany it's a bit weirder but if we log only the first set of `args` then it'll work too (though imperfect)

jatcwang:
From #533, seems like `withGeneratedKeys` has the same problems

jatcwang:
Please give 1.0-RC6 a try when it's released which should fix this issue

PR #2064, opened 2024-07-17
Fix logging for streaming and updateMany

### LINKED ISSUE #533: withGeneratedKeys & transaction statements missing from logs

jatcwang:
Introduces three new functions in the high level API for Connection (`doobie.hi.HC`).

Fixes #533, #1888

These helpers offer a few improvements:

- Logging support
- Handle and log exceptions throw during the PreparedStatement creation (Some JDBC drivers do this if the table doesn't exist)
- A slighly simpler "flat" API for anyone that wants to customize their query

## `executeWithResultSet`
This is a generalization for any query that executes a PreparedStatement and processes a ResultSet. This covers two cases:

- A normal select query (JDBC's `executeQuery`)
- An update which generates returned rows (i.e. `RETURNING` / `.executeUpdate + .getGeneratedKeys`)

All query methods like Query#to and Update#withUniqueGeneratedKeys have switched to use this internally.

##  `executeWithoutResultSet`

Similar to `executeWithResultSet` but for queries that doesn't need to process a ResultSet (i.e. simple updates).

Methods like `Update`'s `updateMany`, `update` have switched to use this internally

## `stream`
This will be replacing the existing `stream` method (which is currently deprecated because it doesn't support logging)

`Query#stream` and all the `Update#withGeneratedKeys*` have switched to this internally.

Further improvements can probably be made on top of these methods so users don't need to construct `LoggingInfo` themselves.


henricook:
Thanks for this improvement @jatcwang . I'm trying to convert some existing code when migrating from RC6 -> RC7 due to the deprecation and wonder if you have a minute to provide some guidance? 🙏🏻 

Old code:

```scala
HC.prepareStatement(sql)(HPS.executeQuery {
      for {
        md <- FRS.getMetaData
        rs <- row(md).whileM[List](HRS.next)
      } yield RawResultSet(header(md), rs)
    })
```

It's unclear to me how to convert this to the FC.prepareStatement / executeWithResultSet style, i've got something like this so far:

```scala
  private def exec(sql: String): ConnectionIO[RawResultSet] = {

    val preparedStatement = FC.prepareStatement(sql)

    doobie.hi.connection.executeWithResultSet(
      create = preparedStatement,
      prep = ???,
      exec = ???,
      process = ???,
      LoggingInfo(sql, NonBatch(List.empty), "fooLabel")
    )
```

It's unclear to me how to/what to put in prep, exec and process - might you have any pointers? Or a method I'm missing which just takes care of it for me?

My use case here is running (admin) user provided SQL without any bound parameters - just a straight string -> query mapping.

jatcwang:
Hi @henricook sorry the documentation for custom JDBC operations isn't really there and I haven't gotten around to #2134 yet.

In your case,
- prep: You don't need a prepare step, so you can use `FPS.unit`
- exec: `FPS.executeQuery`
- process: 
```
for {
  md <- FRS.getMetaData
  rs <- row(md).whileM[List](HRS.next)
} yield RawResultSet(header(md), rs)
```

Hope that works!

henricook:
Thanks for a super speedy response @jatcwang. I was reading the MR but Doobie novice that I am it took me a while to realise that IFPS FPS etc are all remappings of things like `doobie.free.preparedstatement`.

Here's my compiling and working final item based on your comments, currently without the 'doobie idiomatic' remapping of `doobie.free.preparedstatement' to `FPS` and `doobie.hi.connection` to `IHC`  - i'll do that next

Notes: 
  - I have no params to bind, so this isn't suitable if you're reading and hoping to bind parameters. 
  - `mkLabel` is a custom method I use just to prefix a repository name onto the label

```scala
private def exec(sql: String): ConnectionIO[RawResultSet] = {

    val preparedStatement = FC.prepareStatement(sql)

    doobie.hi.connection.executeWithResultSet(
      create = preparedStatement,
      prep = doobie.free.preparedstatement.unit,
      exec = doobie.free.preparedstatement.executeQuery,
      process = (for {
        md <- FRS.getMetaData
        rs <- row(md).whileM[List](HRS.next)
      } yield RawResultSet(header(md), rs)),
      LoggingInfo(sql, NonBatch(List.empty), mkLabel("executeQuery"))
    )
```

jatcwang:
Thanks for posting an update @henricook glad it worked :partying_face:

strokyl:
Thanks @henricook for the shared code, it was really useful to me today

PR #2071, opened 2024-07-29
docs: related projects

arturaz:
Added a list of related projects that people otherwise might not find.


satorg:
Yeah, I didn't know about those project previously.

jatcwang:
Thanks!

Issue #1922, opened 2023-09-13
Cancel the DB side query when doobie execution is cancelled 


jatcwang:
If we run a long running query but then cancel it (e.g. application-side timeout or `IO.race`), we might want to cancel the DB side query by calling PreparedStatement#cancel(). (Simply calling `close()` which we currently do does not seem to cancel the query on the DB-side)

e.g. https://discord.com/channels/632277896739946517/632727524434247691/1151507294664786021

Will need some more thought on whether this is universally desirable, and if not we'll need to make this configurable somehow (e.g. opt-in or opt-out)

armanbilge:
> if not we'll need to make this configurable somehow (e.g. opt-in or opt-out)

Repeating my comment from Discord: a user can always opt-out of cancelation by adding `.uncancelable`.

AlekseyKovalevsky:
Hello! Can I as a user customise transactor somehow from the outside to get the behaviour mentioned? Just not to wait for this feature to be implemented.

jatcwang:
@AlekseyKovalevsky In plain Java/JDBC, what needs to be called is PreparedStatement#cancel, but this has to be done from another thread (since JDBC operations are blocking). Therefore in order to do the same in Doobie I think we'll need to fork a fiber to cancel the preparedstatement when a cancellation is signaled from the caller of `.transact`.

armanbilge:
Actually this is a perfect usecase for the `cancelable` combinator in Cats Effect.

https://github.com/typelevel/cats-effect/blob/85915dd87382fe5e50099f724d64ecc7e35d8051/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala#L219-L269

https://typelevel.org/cats-effect/docs/faq#how-do-i-cancel-or-timeout-delay-or-blocking

https://github.com/typelevel/fs2/blob/41521428769dda218863e2d26c322a48c64ba1c9/io/jvm/src/main/scala/fs2/io/net/unixsocket/UnixSocketsPlatform.scala#L103-L104

gontard:
Hello. 
I am trying to achieve that with Doobie, and it seems impossible for now. Has anyone found a workaround?

TalkingFoxMid:
Conversation and possible fix are there https://github.com/tpolecat/doobie/pull/2077

jatcwang:
#2079 and #2088 should address cancellation for the vast majority of the cases.

TalkingFoxMid:
[...]

PR #2077, opened 2024-08-05
fix query cancellation

TalkingFoxMid:
After migrating Doobie to Cats Effect 3 (CE3), query cancellation no longer works as expected. When a query is canceled (using IO.cancel), the current fiber blocks, and PreparedStatement.close is not called immediately. This issue arises because in CE3, the cancellation of a fiber on a blocked thread does not trigger cancellation handlers right away, while PreparedStatement.close needs to be called in a cancellation handler.

Here is a code example that demonstrates this behavior. In CE2, the cancellation handler runs immediately, but in CE3, it only runs after the thread becomes unblocked:

```
import cats.effect.{Async, IO, IOApp}

object OnCancelTest extends IOApp.Simple {
  import scala.concurrent.duration._

  def goDB: IO[Unit] = IO(println("START goDB")).map(
    _ => Thread.sleep(10000)
  )
    .guarantee(
      IO(println("SEND CANCELLATION VIA JDBC"))
    )


  override def run: IO[Unit] =
    for {
      f <- goDB.start
      _ <- IO.sleep(1.second)
      _ <- f.cancel
      _ <- IO(println("CANCELLED goBD"))
    } yield ()
}
``` 
I also added unit test which reproduce that behaviour. 

Alternative more appropriate solution: https://github.com/tpolecat/doobie/pull/2079#issuecomment-2285146784

### DIFF
--- a/modules/free/src/main/scala/doobie/free/Primitive.scala
+++ b/modules/free/src/main/scala/doobie/free/Primitive.scala
@@ -1,0 +1,44 @@
+// Copyright (c) 2013-2020 Rob Norris and Contributors
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package doobie.free
+
+import cats.data.Kleisli
+import cats.effect.Async
+import cats.effect.kernel.Outcome
+import doobie.WeakAsync
+
+sealed trait Primitive[M[_]] {
+  def primitive[J, A](f: J => A): Kleisli[M, J, A]
+}
+
+object Primitive {
+  case class Default[M[_]]()(implicit asyncM: WeakAsync[M]) extends Primitive[M] {
+
+    override def primitive[J, A](f: J => A): Kleisli[M, J, A] = Kleisli { a =>
+      // primitive JDBC methods throw exceptions and so do we when reading values
+      // so catch any non-fatal exceptions and lift them into the effect
+      try {
+        asyncM.blocking(f(a))
+      } catch {
+        case scala.util.control.NonFatal(e) => asyncM.raiseError(e)
+      }
+    }
+  }
+
+  case class Cancellable[M[_]]()(implicit asyncM: Async[M]) extends Primitive[M] {
+
+    override def primitive[J, A](f: J => A): Kleisli[M, J, A] = Kleisli { a =>
+      import cats.syntax.all._
+      import cats.effect.syntax.all._
+
+      asyncM.uncancelable { unmask =>
+        for {
+          jdbcBlockedFiber <- asyncM.blocking(f(a)).attempt.start
+          a <- unmask(jdbcBlockedFiber.join).flatMap(_.embedNever.rethrow)
+        } yield a
+      }
+    }
+  }
+}
--- a/modules/free/src/main/scala/doobie/free/Primitive.scala
+++ b/modules/free/src/main/scala/doobie/free/Primitive.scala
@@ -6,7 +6,6 @@ package doobie.free
 
 import cats.data.Kleisli
 import cats.effect.Async
-import cats.effect.kernel.Outcome
 import doobie.WeakAsync
 
 sealed trait Primitive[M[_]] {
--- a/modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
+++ b/modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
@@ -11,7 +11,6 @@ import cats.~>
 import cats.data.Kleisli
 import cats.effect.kernel.{Async, Outcome, Poll, Sync}
 import cats.free.Free
-import doobie.util.cancellation.CancellationForker
 import doobie.util.log.{LogEvent, LogHandler}
 
 import scala.concurrent.Future
@@ -102,6 +101,11 @@ class KleisliInterpreter[M[_]](logHandler: LogHandler[M])(implicit val asyncM: A
     // primitive JDBC methods throw exceptions and so do we when reading values
     // so catch any non-fatal exceptions and lift them into the effect
     import cats.syntax.all._
+//    try {
+//      asyncM.blocking(f(a))
+//    } catch {
+//      case scala.util.control.NonFatal(e) => asyncM.raiseError(e)
+//    }
     for {
       jdbcBlockedFiber <- asyncM.start(asyncM.blocking(f(a)).attempt)
       a <- jdbcBlockedFiber.join.flatMap {
--- a/modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
+++ b/modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
@@ -9,11 +9,11 @@ package doobie.free
 // Library imports
 import cats.~>
 import cats.data.Kleisli
-import cats.effect.kernel.{Async, Poll, Sync}
+import cats.effect.kernel.{ Poll, Sync }
 import cats.free.Free
 import doobie.WeakAsync
+import doobie.free.Primitive
 import doobie.util.log.{LogEvent, LogHandler}
-
 import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
 
@@ -81,7 +81,8 @@ object KleisliInterpreter {
 // Family of interpreters into Kleisli arrows for some monad M.
 class KleisliInterpreter[M[_]](logHandler: LogHandler[M], customPrimitive: Option[Primitive[M]] = None)(implicit val asyncM: WeakAsync[M]) { outer =>
 
-  private val _primitive = customPrimitive.getOrElse(Primitive.Default[M])
+  private val _primitive = customPrimitive.getOrElse(Primitive.Default[M]())
+
 
   // The 14 interpreters, with definitions below. These can be overridden to customize behavior.
   lazy val NClobInterpreter: NClobOp ~> Kleisli[M, NClob, *] = new NClobInterpreter { }
@@ -101,6 +102,7 @@ class KleisliInterpreter[M[_]](logHandler: LogHandler[M], customPrimitive: Optio
 
   // Some methods are common to all interpreters and can be overridden to change behavior globally.
   def primitive[J, A](f: J => A): Kleisli[M, J, A] = _primitive.primitive(f)
+
   def raw[J, A](f: J => A): Kleisli[M, J, A] = primitive(f)
   def raiseError[J, A](e: Throwable): Kleisli[M, J, A] = Kleisli(_ => asyncM.raiseError(e))
   def monotonic[J]: Kleisli[M, J, FiniteDuration] = Kleisli(_ => asyncM.monotonic)
--- a/modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
+++ b/modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
@@ -9,10 +9,11 @@ package doobie.free
 // Library imports
 import cats.~>
 import cats.data.Kleisli
-import cats.effect.kernel.{ Poll, Sync }
+import cats.effect.kernel.{Async, Outcome, Poll, Sync}
 import cats.free.Free
-import doobie.WeakAsync
+import doobie.util.cancellation.CancellationForker
 import doobie.util.log.{LogEvent, LogHandler}
+
 import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
 
@@ -73,12 +74,12 @@ import doobie.free.callablestatement.{ CallableStatementIO, CallableStatementOp 
 import doobie.free.resultset.{ ResultSetIO, ResultSetOp }
 
 object KleisliInterpreter {
-  def apply[M[_]: WeakAsync](logHandler: LogHandler[M]): KleisliInterpreter[M] =
+  def apply[M[_]: Async](logHandler: LogHandler[M]): KleisliInterpreter[M] =
     new KleisliInterpreter[M](logHandler)
 }
 
 // Family of interpreters into Kleisli arrows for some monad M.
-class KleisliInterpreter[M[_]](logHandler: LogHandler[M])(implicit val asyncM: WeakAsync[M]) { outer =>
+class KleisliInterpreter[M[_]](logHandler: LogHandler[M])(implicit val asyncM: Async[M]) { outer =>
 
   // The 14 interpreters, with definitions below. These can be overridden to customize behavior.
   lazy val NClobInterpreter: NClobOp ~> Kleisli[M, NClob, *] = new NClobInterpreter { }
@@ -100,11 +101,18 @@ class KleisliInterpreter[M[_]](logHandler: LogHandler[M])(implicit val asyncM: W
   def primitive[J, A](f: J => A): Kleisli[M, J, A] = Kleisli { a =>
     // primitive JDBC methods throw exceptions and so do we when reading values
     // so catch any non-fatal exceptions and lift them into the effect
-    try {
-      asyncM.blocking(f(a))
-    } catch {
-      case scala.util.control.NonFatal(e) => asyncM.raiseError(e)
-    }
+    import cats.syntax.all._
+    for {
+      jdbcBlockedFiber <- asyncM.start(asyncM.blocking(f(a)).attempt)
+      a <- jdbcBlockedFiber.join.flatMap {
+        case Outcome.Succeeded(fa) => fa.flatMap {
+          case Left(e) => e.raiseError[M, A]
+          case Right(value) => value.pure[M]
+        }
+        case Outcome.Errored(e) => e.raiseError[M, A]
+        case Outcome.Canceled() => asyncM.never[A] // Cannot be cancelled out of scope
+      }
+    } yield a
   }
   def raw[J, A](f: J => A): Kleisli[M, J, A] = primitive(f)
   def raiseError[J, A](e: Throwable): Kleisli[M, J, A] = Kleisli(_ => asyncM.raiseError(e))
--- a/modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
+++ b/modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
@@ -9,8 +9,9 @@ package doobie.free
 // Library imports
 import cats.~>
 import cats.data.Kleisli
-import cats.effect.kernel.{Async, Outcome, Poll, Sync}
+import cats.effect.kernel.{Async, Poll, Sync}
 import cats.free.Free
+import doobie.WeakAsync
 import doobie.util.log.{LogEvent, LogHandler}
 
 import scala.concurrent.Future
@@ -73,12 +74,14 @@ import doobie.free.callablestatement.{ CallableStatementIO, CallableStatementOp 
 import doobie.free.resultset.{ ResultSetIO, ResultSetOp }
 
 object KleisliInterpreter {
-  def apply[M[_]: Async](logHandler: LogHandler[M]): KleisliInterpreter[M] =
-    new KleisliInterpreter[M](logHandler)
+  def apply[M[_]: WeakAsync](logHandler: LogHandler[M], customPrimitive: Option[Primitive[M]] = None): KleisliInterpreter[M] =
+    new KleisliInterpreter[M](logHandler, customPrimitive)
 }
 
 // Family of interpreters into Kleisli arrows for some monad M.
-class KleisliInterpreter[M[_]](logHandler: LogHandler[M])(implicit val asyncM: Async[M]) { outer =>
+class KleisliInterpreter[M[_]](logHandler: LogHandler[M], customPrimitive: Option[Primitive[M]] = None)(implicit val asyncM: WeakAsync[M]) { outer =>
+
+  private val _primitive = customPrimitive.getOrElse(Primitive.Default[M])
 
   // The 14 interpreters, with definitions below. These can be overridden to customize behavior.
   lazy val NClobInterpreter: NClobOp ~> Kleisli[M, NClob, *] = new NClobInterpreter { }
@@ -97,27 +100,7 @@ class KleisliInterpreter[M[_]](logHandler: LogHandler[M])(implicit val asyncM: A
   lazy val ResultSetInterpreter: ResultSetOp ~> Kleisli[M, ResultSet, *] = new ResultSetInterpreter { }
 
   // Some methods are common to all interpreters and can be overridden to change behavior globally.
-  def primitive[J, A](f: J => A): Kleisli[M, J, A] = Kleisli { a =>
-    // primitive JDBC methods throw exceptions and so do we when reading values
-    // so catch any non-fatal exceptions and lift them into the effect
-    import cats.syntax.all._
-//    try {
-//      asyncM.blocking(f(a))
-//    } catch {
-//      case scala.util.control.NonFatal(e) => asyncM.raiseError(e)
-//    }
-    for {
-      jdbcBlockedFiber <- asyncM.start(asyncM.blocking(f(a)).attempt)
-      a <- jdbcBlockedFiber.join.flatMap {
-        case Outcome.Succeeded(fa) => fa.flatMap {
-          case Left(e) => e.raiseError[M, A]
-          case Right(value) => value.pure[M]
-        }
-        case Outcome.Errored(e) => e.raiseError[M, A]
-        case Outcome.Canceled() => asyncM.never[A] // Cannot be cancelled out of scope
-      }
-    } yield a
-  }
+  def primitive[J, A](f: J => A): Kleisli[M, J, A] = _primitive.primitive(f)
   def raw[J, A](f: J => A): Kleisli[M, J, A] = primitive(f)
   def raiseError[J, A](e: Throwable): Kleisli[M, J, A] = Kleisli(_ => asyncM.raiseError(e))
   def monotonic[J]: Kleisli[M, J, FiniteDuration] = Kleisli(_ => asyncM.monotonic)
--- a/modules/hikari/src/test/scala/doobie/HikariQueryCancellationSuite.scala
+++ b/modules/hikari/src/test/scala/doobie/HikariQueryCancellationSuite.scala
@@ -1,0 +1,56 @@
+// Copyright (c) 2013-2020 Rob Norris and Contributors
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package doobie
+
+import cats.effect.*
+import cats.effect.unsafe.implicits.global
+import com.zaxxer.hikari.HikariConfig
+import doobie.hikari.HikariTransactor
+import doobie.implicits.*
+import doobie.util.transactor
+
+import scala.concurrent.duration.DurationInt
+
+class HikariQueryCancellationSuite extends munit.FunSuite {
+
+  // Typically you construct a transactor this way, using lifetime-managed thread pools.
+  val transactorRes: Resource[IO, Transactor[IO]] =
+    (for {
+      hikariConfig <- Resource.pure {
+        val config = new HikariConfig()
+        config.setDriverClassName("org.postgresql.Driver")
+        config.setJdbcUrl("jdbc:postgresql:world")
+        config.setUsername("postgres")
+        config.setPassword("password")
+        config.setMaximumPoolSize(2)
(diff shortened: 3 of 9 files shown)


TalkingFoxMid:
> Thanks for trying this out @TalkingFoxMid. Have you seen the comments in #1922?
> 
> One of the thing we want to do is to signal to the database that it should cancel the running query on the database side too. I believe that we might need `PreparedStatement#cancel` for that. If you want to simulate a long running query in postgres you can try using `pg_sleep` - I think that way you can test whether a query was really cancelled on the database side.

Thanks for the reply. Cancellation on the database side is already working in this solution. I verified it with the following test:
1) Fiber 1 locks the table.
2) Fiber 2 attempts to insert data.
3) Cancel Fiber 2.
4) Fiber 1 unlocks the table.
5) Select from the table → The table is empty.

This can also be demonstrated in the QueryCancellationSuite by adding a join on the lock fiber before the final table read. I believe that in this case, the insert query (2) was canceled on the DB side.

TalkingFoxMid:
> Thanks for trying this out @TalkingFoxMid. Have you seen the comments in #1922?
> 
> One of the thing we want to do is to signal to the database that it should cancel the running query on the database side too. I believe that we might need `PreparedStatement#cancel` for that. If you want to simulate a long running query in postgres you can try using `pg_sleep` - I think that way you can test whether a query was really cancelled on the database side.

 But I'll also try testing this with pg_sleep to confirm. Thanks!

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
```diff
@@ -100,11 +101,18 @@ class KleisliInterpreter[M[_]](logHandler: LogHandler[M])(implicit val asyncM: W
...
     // so catch any non-fatal exceptions and lift them into the effect
-    try {
-      asyncM.blocking(f(a))
-    } catch {
-      case scala.util.control.NonFatal(e) => asyncM.raiseError(e)
-    }
+    import cats.syntax.all._
+    for {
+      jdbcBlockedFiber <- asyncM.start(asyncM.blocking(f(a)).attempt)
+      a <- jdbcBlockedFiber.join.flatMap {
```
satorg, line 107: I've got some concerns regarding this code. They may or may not be valid though, but let me pronounce them:
- `Sync#blocking` is non-cancellable by design. Therefore the execution pipeline can be cancelled either before entering `blocking` or right after it completes, but not inside. Therefore it can only cancel a fiber that runs the query, but not the operation itself.
- There can be a potential fiber leak with this code: if a cancel signal comes when the fiber is created, but before it gets joined, then here it goes (<sub>btw, I believe this is one of the reasons why `background` is almost always preferable over low-level fiber start/join manipulations</sub>).
- Firing a fiber and then joining it immediately after looks quite suspicious. What is it attempted to accomplish there exactly?

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
```diff
@@ -73,12 +74,12 @@ import doobie.free.callablestatement.{ CallableStatementIO, CallableStatementOp
 import doobie.free.resultset.{ ResultSetIO, ResultSetOp }
 
 object KleisliInterpreter {
-  def apply[M[_]: WeakAsync](logHandler: LogHandler[M]): KleisliInterpreter[M] =
+  def apply[M[_]: Async](logHandler: LogHandler[M]): KleisliInterpreter[M] =
```
satorg, line 77: I feel that upgrading `WeakAsync` to more powerful `Async` here can break a lot of client code. So it may cause quite a turmoil. There are a lot of dependencies and assumptions across the library towards `WeakAsync` typeclass already. I would rather try to reconsider that fiber start/join approach below, especially because it does not seem absolutely necessary to me.

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
TalkingFoxMid: > Therefore it can only cancel the fiber that runs the query, but not the operation itself.

Yes, it is. But cancellation handler that sends the cancel query to the DB is on the fiber that can be cancelled now. So, the blocked fiber will also be canceled from the JDBC side (the database stops the JDBC thread).

> There can be a potential fiber leak with this code

I think there is no fiber leak, the asyncM.blocking fiber will continue running until the JDBC thread finishes dealing with the query. When the JDBC query is completed and the thread is unblocked, that fiber will be cleaned up. Currently, Doobie works in exactly the same way. However, there is a problem with this approach when it comes to background tasks. The cancellation of a fiber also tries to cancel the jdbcBlockedFiber, but the issue is that the cancellation of jdbcBlockedFiber is blocking and doesn't work effectively. So, using background tasks doesn't solve the problem

> What is it attempted to accomplish there exactly?

You cannot cancel jdbcBlockedFiber because it is running on a blocked JDBC thread. However, you can cancel the jdbcBlockedFiber.join.start fiber, because it is blocked asynchronously. This allows the cancellation handler, which sends a cancel query to the database, to be executed. Please refer to the example code in the description, as it illustrates the problem.

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
TalkingFoxMid: I don't think it's possible to solve this problem in another way. However, I have some ideas on how to minimize compatibility issues. I'll try to implement a fix soon.

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
satorg: > I think there is no fiber leak, the asyncM.blocking fiber will continue running until the JDBC thread finishes dealing with the query. When the JDBC query is completed and the thread is unblocked, that fiber will be cleaned up.

Actually, this is what I would call a "fiber leak". Because the fiber goes out of control at this point and lives solely on the mercy of JDBC.

> This allows the cancellation handler, which sends a cancel query to the database, to be executed.

Could you elaborate a bit – why do you think a cancel query is sent to the database in this case? Where is it happening exactly?

I see that you are trying to reproduce it with the test. However, I'm not sure it is working because the query gets cancelled gracefully. Perhaps it is happening just because the connection to the DB gets closed after the cancellation, which in turn cancels the query. However, it may not always work with connection pooling, for example.

> There is a problem with this approach when it comes to background tasks. The cancellation of a fiber also tries to cancel the jdbcBlockedFiber, but the issue is that the cancellation of jdbcBlockedFiber is blocking and doesn't work.

Yes, exactly. Because it guarantees that there's no leak in that case. I feel it was designed that way for a good reason.

TalkingFoxMid: > Actually, this is what I would call a "fiber leak". Because the fiber goes out of control at this point and lives solely on the mercy of JDBC.

There is no way in CE3 to cancel a fiber running on a blocked thread, I suppose. So the fiber lives solely on the mercy of JDBC anyway.

TalkingFoxMid: > Could you elaborate a bit – why do you think a cancel query is sent to the database in this case? Where is it happening exactly?

I added a debug log to PreparedStatementInterpreter.close to check if the raw JDBC method (PreparedStatement.close) was called after the fiber was canceled.

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
```diff
@@ -73,12 +74,12 @@ import doobie.free.callablestatement.{ CallableStatementIO, CallableStatementOp
...
     new KleisliInterpreter[M](logHandler)
 }
 
 // Family of interpreters into Kleisli arrows for some monad M.
-class KleisliInterpreter[M[_]](logHandler: LogHandler[M])(implicit val asyncM: WeakAsync[M]) { outer =>
+class KleisliInterpreter[M[_]](logHandler: LogHandler[M])(implicit val asyncM: Async[M]) { outer =>
```
jatcwang, line 82: Note that this whole file is generated so you may need to port the changes to FreeGen2.scala at some point and run `freeGen2` to generate this file.

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
satorg: > There is no way in CE3 to cancel a fiber running on a blocked thread, I suppose. So the fiber lives solely on the mercy of JDBC anyway.

Actually, there are two ways in general: [interruptible](https://github.com/typelevel/cats-effect/blob/be1b65687215b1332a7be80cd6759d147680c04f/kernel/shared/src/main/scala/cats/effect/kernel/Sync.scala#L93) and [cancellable](https://github.com/typelevel/cats-effect/blob/be1b65687215b1332a7be80cd6759d147680c04f/kernel/shared/src/main/scala/cats/effect/kernel/GenSpawn.scala#L264). The former is unlikely to help in this case, but the latter could come in handy, perhaps.

satorg: See also https://github.com/tpolecat/doobie/issues/1922#issuecomment-1751461106

TalkingFoxMid: There is one disadvantage with using *cancellable*. While it allows a cancellation handler to be executed even if a fiber is running on a blocking thread, it doesn't work properly when we wrap the call in a timeout. In this scenario, the client’s query to the service will remain stuck until the JDBC thread is unblocked. Below is an illustrative example of this issue in code:

```scala
import cats.effect.{Async, IO, IOApp}

object OnCancelTest extends IOApp.Simple {
  import scala.concurrent.duration._

  def goDB: IO[Unit] = IO(println("START goDB")).map(
    _ => Thread.sleep(10000)
  )
  .guarantee(
    IO(println("SEND CANCELLATION VIA JDBC"))
  )

  override def run: IO[Unit] =
    for {
      f <- goDB.timeout(
        1.seconds
      ).start
      _ <- f.join
      _ <- IO(println("Timeout")) // This prints only after 10 seconds
    } yield ()
}
```

TalkingFoxMid: So now we have two possible solutions:

1) Don't control the JDBC fiber lifecycle at all: This approach works well with timeouts.

2) Control the JDBC fiber lifecycle using Cancellable: However, this approach leads to issues where timeouts do not work as expected.

TalkingFoxMid: There is also another workaround: you can run the JDBC query directly on a blocking ExecutionContext using Future, for example. Then, convert it to IO using IO.fromFuture. With this approach, there won't be a JDBC fiber, so there's no need to worry about controlling the fiber's lifecycle. (IO.fromFuture will cancel the fiber instantly, as it is implemented through IO.async).

TalkingFoxMid: @jatcwang I’d like to hear your opinion as well.  I'd like to understand which approach is more correct.

armanbilge: > it doesn't work properly when we wrap the call in a timeout.

@TalkingFoxMid Thanks for raising this, would you be able to give a small demonstration of this problem, ideally using only Cats Effect? I worked with Daniel on `cancelable` in https://github.com/typelevel/cats-effect/pull/3460 and I would like to understand this limitation.

satorg: @TalkingFoxMid , I'm not sure how you came to the conclusion that `cancelable` does not respect timeouts. Besides, the code that you suggest as an example of the issue doesn't seem having any business with `cancelable` per se (or am I missing some point here?)

On the other hand, I tried it myself and it seems working pretty well to me:
```scala
import cats.effect.*
import scala.concurrent.duration.*

object TimeoutCancellableApp extends IOApp.Simple:
  @volatile
  private var taskCancelled: Boolean = false

  private def blockingButCancellableTask: IO[Unit] = IO.blocking {
    // Runs forever unless canceled.
    while (!taskCancelled) Thread.sleep(100)
  }

  private def cancelTask: IO[Unit] = IO { taskCancelled = true }

  override def run: IO[Unit] =
    blockingButCancellableTask
      .cancelable(cancelTask)
      .timeout(3.seconds)
```
When I run this example, I get the following output:
```
[error] java.util.concurrent.TimeoutException: 3 seconds
```
i.e. it breaks the execution after 3 seconds as requested.

jatcwang: @TalkingFoxMid I have pushed a test case using pg_sleep and connection pool.
From what I can tell the cancellation is happening but that's from PreparedStatment#close() being called (due to various `bracket`). Whether the DB-side query is cancelled when PreparedStatment#close() is called is implementation dependent but since pgjdbc does cancel the query I am happy with the general approach.

In terms of the implementation details of the cancellation I'll need to study it more and will be following this discussion closely. Thanks!

satorg: @TalkingFoxMid @jatcwang ,
Although I'm still not pretty sure it is the best approach with firing a fiber and then relying on some external cancelation magic, but it looks working indeed (I checked) and also it looks simpler to accomplish comparing to a possible version with `cancelable` and explicit cancelation effect.

However, I would probably suggest to add some protection against fiber leaks and also simplify results processing a little, i.e. the `primitive` function might look like the following:
```scala
  def primitive[J, A](f: J => A): Kleisli[M, J, A] = Kleisli { a =>

    asyncM.uncancelable { unmask =>
      for {
        jdbcBlockedFiber <- asyncM.blocking(f(a)).attempt.start
        a <- unmask(jdbcBlockedFiber.join).flatMap(_.embedNever.rethrow)
      } yield a
    }
  }
```
Here, cancelation is masked for the entire execution and fiber creation workout, but then gets unmasked just for the fiber's `join` operation. It should guarantee that there will be no leaked fibers possible. I checked it with both `QueryCancellationSuite` and `HikariQueryCancellationSuite` – both seem working correctly to me.

TalkingFoxMid: > @TalkingFoxMid , I'm not sure how you came to the conclusion that `cancelable` does not respect timeouts. Besides, the code that you suggest as an example of the issue doesn't seem having any business with `cancelable` per se (or am I missing some point here?)
> 
> On the other hand, I tried it myself and it seems working pretty well to me:
> 
> ```scala
> import cats.effect.*
> import scala.concurrent.duration.*
> 
> object TimeoutCancellableApp extends IOApp.Simple:
>   @volatile
>   private var taskCancelled: Boolean = false
> 
>   private def blockingButCancellableTask: IO[Unit] = IO.blocking {
>     // Runs forever unless canceled.
>     while (!taskCancelled) Thread.sleep(100)
>   }
> 
>   private def cancelTask: IO[Unit] = IO { taskCancelled = true }
> 
>   override def run: IO[Unit] =
>     blockingButCancellableTask
>       .cancelable(cancelTask)
>       .timeout(3.seconds)
> ```
> 
> When I run this example, I get the following output:
> 
> ```
> [error] java.util.concurrent.TimeoutException: 3 seconds
> ```
> 
> i.e. it breaks the execution after 3 seconds as requested.

Consider the following scenario: 

A query is canceled due to a timeout -> 
Triggers a call to PreparedStatement.close() -> 
Cancellation request is sent to the DB ->
However, the DB does not immediately respond to the PreparedStatement.close() request (possibly due to a failover or another issue). In this situation, the application is left waiting for the JDBC operation to complete (because we are trying to guarantee JDBC-fiber closing), causing the user query to hang (and there timeout does not work as expected)

It is not possible to instantly cancel the JDBC operation **via an external state** using cancellable (like it is in your example) in this case. I suppose that calling PreparedStatement.cancel() is the only way to attempt this, but it appears to be unreliable because it depends on the DB availability

TalkingFoxMid: > > it doesn't work properly when we wrap the call in a timeout.
> 
> @TalkingFoxMid Thanks for raising this, would you be able to give a small demonstration of this problem, ideally using only Cats Effect? I worked with Daniel on `cancelable` in [typelevel/cats-effect#3460](https://github.com/typelevel/cats-effect/pull/3460) and I would like to understand this limitation.


**cancellable** provides a way to cancel a fiber running on a blocked thread by changing an external state, allowing the blocked thread to detect the change and stop execution. However, in our case, there is a potential issue: changing the external state via PreparedStatement.close() may not always be available, as it depends on the database's availability. If the database is slow to respond, the close operation might not execute instantly. This delay occurs because the cancellation request has to be sent to the database over the network, which can introduce additional latency. Consequently, the timeout might not cancel the query after the intended 300ms, but instead after 600ms or more, due to this added latency

This is the only limitation of the Cancelable interface that I see in this context

TalkingFoxMid: > @TalkingFoxMid @jatcwang , Although I'm still not pretty sure it is the best approach with firing a fiber and then relying on some external cancelation magic, but it looks working indeed (I checked) and also it looks simpler to accomplish comparing to a possible version with `cancelable` and explicit cancelation effect.
> 
> However, I would probably suggest to add some protection against fiber leaks and also simplify results processing a little, i.e. the `primitive` function might look like the following:
> 
> ```scala
>   def primitive[J, A](f: J => A): Kleisli[M, J, A] = Kleisli { a =>
> 
>     asyncM.uncancelable { unmask =>
>       for {
>         jdbcBlockedFiber <- asyncM.blocking(f(a)).attempt.start
>         a <- unmask(jdbcBlockedFiber.join).flatMap(_.embedNever.rethrow)
>       } yield a
>     }
>   }
> ```
> 
> Here, cancelation is masked for the entire execution and fiber creation workout, but then gets unmasked just for the fiber's `join` operation. It should guarantee that there will be no leaked fibers possible. I checked it with both `QueryCancellationSuite` and `HikariQueryCancellationSuite` – both seem working correctly to me.

Could you elaborate on where jdbcBlockedFiber cancellation occurs? I wrote a simple code example, and it doesn't seem that this fiber is being canceled:

```
object OnCancelTest extends IOApp.Simple {
  import scala.concurrent.duration._

  def goDB: IO[Unit] =
    IO.uncancelable(unmask =>
      for {
        jdbcBlockedFiber <- (IO.sleep(10.seconds).onCancel(IO(println("cancel blocked fiber")))).start
        _ <-  unmask(jdbcBlockedFiber.join)
      } yield ()
    )


  override def run: IO[Unit] =
    for {
      f <- goDB.onCancel(IO(println("OnCancel"))).start
      _ <- IO.sleep(1.second)
      _ <- f.cancel
      _ <- IO(println("Done"))
      _ <- IO.never[Unit]
    } yield ()
}
```

"cancel blocked fiber" prints only after 10 seconds, but not after f.cancel is called

TalkingFoxMid: @satorg  @jatcwang I had prepared an alternative solution there: https://github.com/tpolecat/doobie/pull/2079 
It is based on adding **cancelable** operation to WeakASync

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
TalkingFoxMid: Done

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
TalkingFoxMid: Done with optional parameter primitive

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
TalkingFoxMid: @satorg I thought about fiber leak and have an insight. Doesn't fiber leak in that context equals thread leak ? Code is:

```
asyncM.blocking(f(a)).start
```

I suppose there is only one case when fiber can hang forever: if blocking thread hangs too with that task **f(a)** forever.  I don't really think it is relevant to worry about fiber leak here because: fiber leak => thread leak, and we now we allow ourselves to believe that there is no thread leak

armanbilge: > If the database is slow to respond, the close operation might not execute instantly. The delay occurs also because the cancellation request has to be sent to the database over the network, which can introduce additional latency. Consequently, the timeout might not cancel the query after the intended 300ms, but instead after 600ms or more, due to this added latency (and it leads to user request stucks)

@TalkingFoxMid Thanks for explaining. If I understood correctly, what you are describing is not a limitation of `cancelable` but actually a general feature of Cats Effect, which is that it backpressures on cancelation i.e. if you request cancelation of some fiber, it will wait until that cancelation fully completes.

Consider this example:
```scala
IO.sleep(2.seconds).onCancel(IO.sleep(1.second)).timeOut(1.second)
```

Even though the timeout is after 1 second, it will take 2 seconds to complete because of the `onCancel` finalizer. Is this what you are considering to be a limitation?

jatcwang (approved):

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
TalkingFoxMid: > > If the database is slow to respond, the close operation might not execute instantly. The delay occurs also because the cancellation request has to be sent to the database over the network, which can introduce additional latency. Consequently, the timeout might not cancel the query after the intended 300ms, but instead after 600ms or more, due to this added latency (and it leads to user request stucks)
> 
> @TalkingFoxMid Thanks for explaining. If I understood correctly, what you are describing is not a limitation of `cancelable` but actually a general feature of Cats Effect, which is that it backpressures on cancelation i.e. if you request cancelation of some fiber, it will wait until that cancelation fully completes.
> 
> Consider this example:
> 
> ```scala
> IO.sleep(2.seconds).onCancel(IO.sleep(1.second)).timeout(1.second)
> ```
> 
> Even though the timeout is after 1 second, it will take 2 seconds to complete because of the `onCancel` finalizer. Is this what you are considering to be a limitation?


Consider also this example:

```
import cats.effect.{Async, IO, IOApp}
import scala.concurrent.duration._

object OnCancelTest extends IOApp.Simple {
  @volatile var blocked = true

  def goDB: IO[Unit] =
    IO {
      while (blocked) {Thread.sleep(100)}
    }.cancelable(
      {IO.sleep(5.seconds) >> IO {blocked = false}}.start.void
    )

  override def run: IO[Unit] =
    for {
      fiber <- goDB.start
      _ <- IO.sleep(100.millis)
      _ <- fiber.cancel
      _ <- IO(println("Cancelled"))
    } yield ()
}
```

In that snippet, fiber.cancel was blocked not only because it waits for the cancelable handler to complete (the handler completes instantly because it only starts another fiber). The problem also lies in the fact that the external state wasn't modified instantly, even if the cancellation handler completed successfully. This is similar to what we see in Doobie: even after a cancel query is sent to the database, there is still some latency before the database unblocks the blocked JDBC thread running query.

I don't think this is a limitation of CE3, but it does seem like a slight disadvantage of this approach in the context of the current problem in Doobie. 

This approach can cause the user query to hang for a long time because we are trying to wait for the query to be canceled in the database. However, there are situations when you just need to **timeout** immediately, return the default value and continue executing the request further. This is exactly our case, so this disadvantage is relevant for us

TalkingFoxMid: @satorg Could you give an approve here?

a-khakimov (approved):

modules/hikari/src/test/scala/doobie/HikariQueryCancellationSuite.scala
```diff
@@ -21,9 +21,9 @@ class HikariQueryCancellationSuite extends munit.FunSuite {
       hikariConfig <- Resource.pure {
         val config = new HikariConfig()
         config.setDriverClassName("org.postgresql.Driver")
-        config.setJdbcUrl("jdbc:postgresql:world")
+        config.setJdbcUrl("jdbc:postgresql://localhost:5432/postgres")
         config.setUsername("postgres")
-        config.setPassword("password")
+        config.setPassword("mysecretpassword")
```
jatcwang, line 26: These should be reverted back to `jdbc:postgresql:world` right?

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
armanbilge: @TalkingFoxMid thanks for including that second example, that was very helpful.

> However, there are situations when you just need to **timeout** immediately, return the default value and continue executing the request further. This is exactly our case, so this disadvantage is relevant for us

Cats Effect specifically has a [`timeoutAndForget`](https://github.com/typelevel/cats-effect/blob/caee0e86f815b641af61f19becd4eac27a769f11/core/shared/src/main/scala/cats/effect/IO.scala#L832-L851) API for this.

I agree that sometimes behavior like this is necessary, but that should be the (careful) choice of the user. Once backpressure is removed, it's not possible to restore it. I am dubious about removing backpressure signals in a library, since then users no longer have agency to make the appropriate decision for their specific usecase.

modules/free/src/main/scala/doobie/free/Primitive.scala
```diff
@@ -0,0 +1,43 @@
...
+import cats.data.Kleisli
+import cats.effect.Async
+import doobie.WeakAsync
+
+sealed trait Primitive[M[_]] {
+  def primitive[J, A](f: J => A): Kleisli[M, J, A]
+}
+
+object Primitive {
+  case class Default[M[_]]()(implicit asyncM: WeakAsync[M]) extends Primitive[M] {
```
satorg, line 16: In general, it is recommended to avoid using case classes in publicly available libraries unless absolutely necessary. I guess, the primary reason for that here was to get a static `apply` method. Totally understandable it is, but case classes generate a lot of other stuff which makes them difficult to deal with in terms of maintaining binary compatibility. For example: a `copy` method, inheriting `Product`, and the last but not least a notorious `unapply` static method.

A regular class should work just as well here, and if the static `apply` is desirable, it's better to add it manually.

satorg: The same is applicable to the `case class Cancelable` below.

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
```diff
@@ -73,12 +74,15 @@ import doobie.free.callablestatement.{ CallableStatementIO, CallableStatementOp
...
+  def apply[M[_]: WeakAsync](logHandler: LogHandler[M], customPrimitive: Option[Primitive[M]] = None): KleisliInterpreter[M] =
+    new KleisliInterpreter[M](logHandler, customPrimitive)
 }
 
 // Family of interpreters into Kleisli arrows for some monad M.
-class KleisliInterpreter[M[_]](logHandler: LogHandler[M])(implicit val asyncM: WeakAsync[M]) { outer =>
+class KleisliInterpreter[M[_]](logHandler: LogHandler[M], customPrimitive: Option[Primitive[M]] = None)(implicit val asyncM: WeakAsync[M]) { outer =>
+
+  private val _primitive = customPrimitive.getOrElse(Primitive.Default[M]())
+
```
satorg, line 85: Just a suggestion here to consider:
```suggestion
class KleisliInterpreter[M[_]](logHandler: LogHandler[M], customPrimitive: Primitive[M])(implicit val asyncM: WeakAsync[M]) { outer =>

  def this(logHandler: LogHandler[M])(implicit val asyncM: WeakAsync[M]) =
    this(logHandler, Primitive.Default[M])
```
Now you can use `customPrimitive` directly.
I.e. in this case the overloaded constructor helps to avoid dealing with `Option`.

modules/free/src/main/scala/doobie/free/kleisliinterpreter.scala
satorg: @TalkingFoxMid , 
> However, there are situations when you just need to timeout immediately, return the default value and continue executing the request further.

The catch here is that most likely there will be no immediate relief whatsoever. As @jatcwang found out recently, even if we give up here immediately, the query will be canceled as a part of bracket's release effect at the end of the transaction. Therefore in a case of long-running cancel query, the entire transaction will have to await until the cancel query completes anyway.

jatcwang:
Superceded by #2079

PR #2079, opened 2024-08-07
fix query cancellation with cancelable

TalkingFoxMid:
Alternative fix of problem described here https://github.com/tpolecat/doobie/pull/2077

### DIFF
--- a/modules/core/src/main/scala/doobie/hi/connection.scala
+++ b/modules/core/src/main/scala/doobie/hi/connection.scala
@@ -175,7 +175,11 @@ object connection {
     }
 
     createLogged
-      .bracket(ps => IFC.embed(ps, prepLogged *> execAndProcessLogged))(ps => IFC.embed(ps, IFPS.close))
+      .bracket(ps =>
+        WeakAsyncConnectionIO.cancelable(
+          IFC.embed(ps, prepLogged *> execAndProcessLogged),
+          IFC.embed(ps, IFPS.close)
+        ))(IFC.embed(_, IFPS.close))
   }
 
   /** Execute a PreparedStatement query and provide rows from the ResultSet in chunks
--- a/modules/core/src/test/scala/doobie/util/UpdateLogSuite.scala
+++ b/modules/core/src/test/scala/doobie/util/UpdateLogSuite.scala
@@ -4,7 +4,7 @@
 
 package doobie.util
 
-import cats.effect.{IO, IOLocal}
+import cats.effect.{IO, Ref}
 import cats.syntax.all.*
 import doobie.*
 import doobie.implicits.*
@@ -18,28 +18,30 @@ class UpdateLogSuite extends munit.FunSuite {
 
   import cats.effect.unsafe.implicits.global
 
-  val ioLocal: IOLocal[LogEvent] =
-    IOLocal[LogEvent](null).unsafeRunSync()
+  val logEventRef: Ref[IO, LogEvent] =
+    Ref.of[IO, LogEvent](null).unsafeRunSync()
 
   val xa = Transactor.fromDriverManager[IO](
     "org.h2.Driver",
     "jdbc:h2:mem:queryspec;DB_CLOSE_DELAY=-1",
     "sa",
     "",
-    logHandler = Some(ev => ioLocal.set(ev))
+    logHandler = Some(ev => logEventRef.set(ev))
   )
 
-  def eventForCIO[A](cio: ConnectionIO[A]): LogEvent =
-    (
-      sql"create table if not exists foo (c1 integer, c2 varchar)".update.run *> cio
-    )
-      .transact(xa)
-      .attempt
-      .flatMap { res =>
-        val _ = res
-        ioLocal.get
-      }
-      .unsafeRunSync()
+  def eventForCIO[A](cio: ConnectionIO[A]): LogEvent = {
+    logEventRef.set(null) *>
+      (
+        sql"create table if not exists foo (c1 integer, c2 varchar)".update.run *> cio
+      )
+        .transact(xa)
+        .attempt
+        .flatMap { res =>
+          val _ = res
+          logEventRef.get
+        }
+  }
+    .unsafeRunSync()
 
   def successEventForCIO[A](cio: ConnectionIO[A]): Success =
     eventForCIO(cio) match {
--- a/modules/hikari/src/test/scala/doobie/HikariQueryCancellationSuite.scala
+++ b/modules/hikari/src/test/scala/doobie/HikariQueryCancellationSuite.scala
@@ -1,0 +1,56 @@
+// Copyright (c) 2013-2020 Rob Norris and Contributors
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package doobie
+
+import cats.effect.*
+import cats.effect.unsafe.implicits.global
+import com.zaxxer.hikari.HikariConfig
+import doobie.hikari.HikariTransactor
+import doobie.implicits.*
+import doobie.util.transactor
+
+import scala.concurrent.duration.DurationInt
+
+class HikariQueryCancellationSuite extends munit.FunSuite {
+
+  // Typically you construct a transactor this way, using lifetime-managed thread pools.
+  val transactorRes: Resource[IO, Transactor[IO]] =
+    (for {
+      hikariConfig <- Resource.pure {
+        val config = new HikariConfig()
+        config.setDriverClassName("org.postgresql.Driver")
+        config.setJdbcUrl("jdbc:postgresql://localhost:5432/postgres")
+        config.setUsername("postgres")
+        config.setPassword("password")
+        config.setMaximumPoolSize(2)
+        config
+      }
+      transactor <- HikariTransactor.fromHikariConfig[IO](hikariConfig)
+    } yield transactor)
+      .map(_.copy(strategy0 = transactor.Strategy.void))
+
+  test("Query cancel with Hikari") {
+    val insert = for {
+      _ <- sql"CREATE TABLE if not exists blah (i text)".update.run
+      _ <- sql"truncate table blah".update.run
+      _ <- sql"INSERT INTO blah values ('1')".update.run
+      _ <- sql"INSERT INTO blah select concat(2, pg_sleep(1))".update.run
+    } yield ()
+    val scenario = transactorRes.use { xa =>
+      for {
+        fiber <- insert.transact(xa).start
+        _ <- IO.sleep(200.millis) *> fiber.cancel
+        _ <- IO.sleep(3.second)
+        _ <- fiber.join.attempt
+        result <- sql"select * from blah order by i".query[String].to[List].transact(xa)
+      } yield {
+        assertEquals(result, List("1"))
+      }
+    }
+
+    scenario.unsafeRunSync()
+  }
+
+}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
         run: sbt +update
 
       - name: Start up Postgres
-        run: docker-compose up -d
+        run: docker compose up -d
 
       - name: Check Headers
         run: sbt '++ ${{ matrix.scala }}' headerCheckAll
--- a/build.sbt
+++ b/build.sbt
@@ -43,7 +43,7 @@ ThisBuild / tlSonatypeUseLegacyHost := false
 ThisBuild / githubWorkflowJavaVersions := Seq(JavaSpec.temurin("11"))
 ThisBuild / githubWorkflowBuildPreamble ++= Seq(
   WorkflowStep.Run(
-    commands = List("docker-compose up -d"),
+    commands = List("docker compose up -d"),
     name = Some("Start up Postgres")
   ),
   WorkflowStep.Sbt(
@@ -426,6 +426,7 @@ lazy val hikari = project
     libraryDependencies ++= Seq(
       // needs to be excluded, otherwise coursier may resolve slf4j-api 2 if > Java 11
       "com.zaxxer" % "HikariCP" % hikariVersion exclude ("org.slf4j", "slf4j-api"),
+      "org.postgresql" % "postgresql" % postgresVersion % "test",
       "com.h2database" % "h2" % h2Version % "test",
       "org.slf4j" % "slf4j-api" % slf4jVersion,
       "org.slf4j" % "slf4j-nop" % slf4jVersion % "test"
--- a/modules/core/src/test/scala/doobie/util/QueryLogSuite.scala
+++ b/modules/core/src/test/scala/doobie/util/QueryLogSuite.scala
@@ -5,7 +5,7 @@
 package doobie.util
 
 import cats.syntax.all.*
-import cats.effect.{IO, IOLocal}
+import cats.effect.{IO, Ref}
 import doobie.free.connection.ConnectionIO
 import doobie.implicits.*
 import doobie.util.log.Parameters.NonBatch
@@ -18,19 +18,24 @@ class QueryLogSuite extends munit.FunSuite with QueryLogSuitePlatform {
 
   import cats.effect.unsafe.implicits.global
 
-  val ioLocal: IOLocal[LogEvent] =
-    IOLocal[LogEvent](null).unsafeRunSync()
+  val logEventRef: Ref[IO, LogEvent] =
+    Ref.of[IO, LogEvent](null).unsafeRunSync()
 
   val xa = Transactor.fromDriverManager[IO](
     "org.h2.Driver",
     "jdbc:h2:mem:queryspec;DB_CLOSE_DELAY=-1",
     "sa",
     "",
-    logHandler = Some(ev => ioLocal.set(ev))
+    logHandler = Some(ev => logEventRef.set(ev))
   )
 
-  def eventForCIO[A](cio: ConnectionIO[A]): LogEvent =
-    cio.transact(xa).attempt.flatMap(_ => ioLocal.get).unsafeRunSync()
+  def eventForCIO[A](cio: ConnectionIO[A]): LogEvent = {
+    for {
+      _ <- logEventRef.set(null)
+      _ <- cio.transact(xa).attempt
+      log <- logEventRef.get
+    } yield log
+  }.unsafeRunSync()
 
   def successEventForCIO[A](cio: ConnectionIO[A]): Success =
     eventForCIO(cio) match {
--- a/modules/free/src/main/scala/doobie/WeakAsync.scala
+++ b/modules/free/src/main/scala/doobie/WeakAsync.scala
@@ -15,6 +15,7 @@ import scala.concurrent.duration.FiniteDuration
 trait WeakAsync[F[_]] extends Sync[F] {
   def fromFuture[A](fut: F[Future[A]]): F[A]
   def fromFutureCancelable[A](fut: F[(Future[A], F[Unit])]): F[A]
+  def cancelable[A](fa: F[A], fin: F[Unit]): F[A]
 }
 
 object WeakAsync {
@@ -39,6 +40,8 @@ object WeakAsync {
       override def onCancel[A](fa: F[A], fin: F[Unit]): F[A] = F.onCancel(fa, fin)
       override def fromFuture[A](fut: F[Future[A]]): F[A] = F.fromFuture(fut)
       override def fromFutureCancelable[A](fut: F[(Future[A], F[Unit])]): F[A] = F.fromFutureCancelable(fut)
+
+      override def cancelable[A](fa: F[A], fin: F[Unit]): F[A] = F.cancelable(fa, fin)
     }
 
   /** Create a natural transformation for lifting an `Async` effect `F` into a `WeakAsync` effect `G`
--- a/modules/free/src/main/scala/doobie/free/blob.scala
+++ b/modules/free/src/main/scala/doobie/free/blob.scala
@@ -59,6 +59,7 @@ object blob { module =>
       def onCancel[A](fa: BlobIO[A], fin: BlobIO[Unit]): F[A]
       def fromFuture[A](fut: BlobIO[Future[A]]): F[A]
       def fromFutureCancelable[A](fut: BlobIO[(Future[A], BlobIO[Unit])]): F[A]
+      def cancelable[A](fa: BlobIO[A], fin: BlobIO[Unit]): F[A]
       def performLogging(event: LogEvent): F[Unit]
 
       // Blob
@@ -118,6 +119,9 @@ object blob { module =>
     }
     case class FromFutureCancelable[A](fut: BlobIO[(Future[A], BlobIO[Unit])]) extends BlobOp[A] {
       def visit[F[_]](v: Visitor[F]) = v.fromFutureCancelable(fut)
+    }
+    case class Cancelable[A](fa: BlobIO[A], fin: BlobIO[Unit]) extends BlobOp[A] {
+      def visit[F[_]](v: Visitor[F]) = v.cancelable(fa, fin)
     }
     case class PerformLogging(event: LogEvent) extends BlobOp[Unit] {
       def visit[F[_]](v: Visitor[F]) = v.performLogging(event)
@@ -181,6 +185,7 @@ object blob { module =>
   def onCancel[A](fa: BlobIO[A], fin: BlobIO[Unit]) = FF.liftF[BlobOp, A](OnCancel(fa, fin))
   def fromFuture[A](fut: BlobIO[Future[A]]) = FF.liftF[BlobOp, A](FromFuture(fut))
   def fromFutureCancelable[A](fut: BlobIO[(Future[A], BlobIO[Unit])]) = FF.liftF[BlobOp, A](FromFutureCancelable(fut))
+  def cancelable[A](fa: BlobIO[A], fin: BlobIO[Unit]) = FF.liftF[BlobOp, A](Cancelable(fa, fin))
   def performLogging(event: LogEvent) = FF.liftF[BlobOp, Unit](PerformLogging(event))
 
   // Smart constructors for Blob-specific operations.
@@ -216,6 +221,7 @@ object blob { module =>
       override def onCancel[A](fa: BlobIO[A], fin: BlobIO[Unit]): BlobIO[A] = module.onCancel(fa, fin)
       override def fromFuture[A](fut: BlobIO[Future[A]]): BlobIO[A] = module.fromFuture(fut)
       override def fromFutureCancelable[A](fut: BlobIO[(Future[A], BlobIO[Unit])]): BlobIO[A] = module.fromFutureCancelable(fut)
+      override def cancelable[A](fa: BlobIO[A], fin: BlobIO[Unit]): BlobIO[A] = module.cancelable(fa, fin)
     }
     
   implicit def MonoidBlobIO[A : Monoid]: Monoid[BlobIO[A]] = new Monoid[BlobIO[A]] {
--- a/modules/free/src/main/scala/doobie/free/callablestatement.scala
+++ b/modules/free/src/main/scala/doobie/free/callablestatement.scala
@@ -80,6 +80,7 @@ object callablestatement { module =>
       def onCancel[A](fa: CallableStatementIO[A], fin: CallableStatementIO[Unit]): F[A]
       def fromFuture[A](fut: CallableStatementIO[Future[A]]): F[A]
       def fromFutureCancelable[A](fut: CallableStatementIO[(Future[A], CallableStatementIO[Unit])]): F[A]
+      def cancelable[A](fa: CallableStatementIO[A], fin: CallableStatementIO[Unit]): F[A]
       def performLogging(event: LogEvent): F[Unit]
 
       // CallableStatement
@@ -361,6 +362,9 @@ object callablestatement { module =>
     }
     case class FromFutureCancelable[A](fut: CallableStatementIO[(Future[A], CallableStatementIO[Unit])]) extends CallableStatementOp[A] {
       def visit[F[_]](v: Visitor[F]) = v.fromFutureCancelable(fut)
+    }
+    case class Cancelable[A](fa: CallableStatementIO[A], fin: CallableStatementIO[Unit]) extends CallableStatementOp[A] {
+      def visit[F[_]](v: Visitor[F]) = v.cancelable(fa, fin)
     }
     case class PerformLogging(event: LogEvent) extends CallableStatementOp[Unit] {
       def visit[F[_]](v: Visitor[F]) = v.performLogging(event)
@@ -1090,6 +1094,7 @@ object callablestatement { module =>
   def onCancel[A](fa: CallableStatementIO[A], fin: CallableStatementIO[Unit]) = FF.liftF[CallableStatementOp, A](OnCancel(fa, fin))
   def fromFuture[A](fut: CallableStatementIO[Future[A]]) = FF.liftF[CallableStatementOp, A](FromFuture(fut))
   def fromFutureCancelable[A](fut: CallableStatementIO[(Future[A], CallableStatementIO[Unit])]) = FF.liftF[CallableStatementOp, A](FromFutureCancelable(fut))
+  def cancelable[A](fa: CallableStatementIO[A], fin: CallableStatementIO[Unit]) = FF.liftF[CallableStatementOp, A](Cancelable(fa, fin))
   def performLogging(event: LogEvent) = FF.liftF[CallableStatementOp, Unit](PerformLogging(event))
 
   // Smart constructors for CallableStatement-specific operations.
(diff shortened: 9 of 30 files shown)


satorg:
@TalkingFoxMid thank you for giving it a shot!

@jatcwang, to be honest, I personally like this approach better.
It might look more complicated due to more changes made, but in fact it is not that much because most of the changes happened in the auto-generated code.

I wonder, is there a chance that we could weigh this PR in?
Also I would like to summon @armanbilge as a "cancelation guru" :)

armanbilge (commented):
> but in fact it is not that much because most of the changes happened in the auto-generated code.

yes, would you mind pointing me at the non-auto-generated changes? 😅 still at a glance I think this approach is better too, since its using the existing `cancelable` IIUC.

jatcwang:
@armanbilge Non-autogenerated changes are in [modules/core/src/main/scala/doobie/hi/connection.scala](https://github.com/tpolecat/doobie/pull/2079/files#diff-12f38be9e7f2b792794d779b9cb20dc06011065316f0ac06568a660831137de3)

modules/hikari/src/test/scala/doobie/HikariQueryCancellationSuite.scala
```diff
@@ -0,0 +1,56 @@
...
+        _ <- IO.sleep(200.millis) *> fiber.cancel
+        _ <- IO.sleep(3.second)
+        _ <- fiber.join.attempt
+        result <- sql"select * from blah order by i".query[String].to[List].transact(xa)
+      } yield {
+        assertEquals(result, List("1"))
+      }
+    }
+
+    scenario.unsafeRunSync()
```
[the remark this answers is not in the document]
jatcwang: We currently don't use it at the moment - I can handle this in another PR to move all tests to use CatsEffectSuite

jatcwang:
`execImpl` pretty much handles all Query/Update methods so I made the following updates:

```
     createLogged
-      .bracket(ps => IFC.embed(ps, prepLogged *> execAndProcessLogged))(ps => IFC.embed(ps, IFPS.close))
+      .flatMap(ps =>
+        WeakAsyncConnectionIO.cancelable(
+          IFC.embed(ps, prepLogged *> execAndProcessLogged)
+            .attempt.flatTap(_ => IFC.embed(ps, IFPS.close)).rethrow,
+          IFC.embed(ps, IFPS.close)))
```
which made `HikariQueryCancellationSuite` pass. (Posting the diff as I cannot push to the PR)

We should also need to make Streaming queries cancellable (`doobie.hi.connection#stream`) but I can tackle this in a separate PR.

jatcwang:
In general I'm quite happy with this approach too as it should have less overhead.

modules/core/src/main/scala/doobie/hi/connection.scala
```diff
@@ -436,20 +454,27 @@ object connection {
...
-      IFC.embed(s, IFPS.close))
+    IFC.prepareStatement(sql, rst.toInt, rsc.toInt, rsh.toInt).flatMap(
+      s => WeakAsyncConnectionIO.cancelable(IFC.embed(s, k).attempt.flatTap(_ => IFC.embed(s, IFPS.close)).rethrow, IFC.embed(s, IFPS.close))
+    )
 
   /** @group Prepared Statements */
   def prepareStatement[A](sql: String, agk: AutoGeneratedKeys)(k: PreparedStatementIO[A]): ConnectionIO[A] =
-    IFC.prepareStatement(sql, agk.toInt).bracket(s => IFC.embed(s, k))(s => IFC.embed(s, IFPS.close))
+    IFC.prepareStatement(sql, agk.toInt).flatMap(
+      s => WeakAsyncConnectionIO.cancelable(IFC.embed(s, k).attempt.flatTap(_ => IFC.embed(s, IFPS.close)).rethrow, IFC.embed(s, IFPS.close))
```
armanbilge, line 464: `.attempt.flatTap(_ => ...).rethrow` is `.guarantee(...)

jatcwang: I'm guessing the intention here is to avoid calling ps.close another time when there's a cancellation?

TalkingFoxMid: Thanks! I did it this way to prevent double calling of IFPS.close in case of cancelation

```
guaranteeCase {
            case Outcome.Canceled() => WeakAsyncConnectionIO.unit
            case _ => IFC.embed(ps, IFPS.close)
          }
```

TalkingFoxMid:
@armanbilge I think I've found an issue with cancelable that I don't know how to work around. Consider the following code example:

```scala
object TestCancelable extends IOApp.Simple {
  val acquire: IO[String] = IO.println("START AQ") >> IO.sleep(2.seconds) >> IO.println("END AQ") >> IO("RESOURCE")

  def evaluation: IO[Unit] = IO.uncancelable {
    unmask =>
      acquire.flatMap(
        r =>
          unmask {
            IO.sleep(5.seconds) >> IO(println(s"USE $r"))
          }.onCancel(IO(println(s"release $r")))
      )
  }

  override def run: IO[Unit] =
    for {
      f <- evaluation.start
      _ <- IO.sleep(1.second)
      _ <- f.cancel
      _ <- IO.never[Unit]
    } yield ()
}
```

Notice that I use IO.uncancelable to guarantee that resource acquisition can't be canceled. This code is very similar to `IO.bracket`. In this case, after f.cancel is called, the only effect we can cancel is the resource-using effect: `IO.sleep(5.seconds) >> IO(println(s"USE $r"))`. That's why resource usage is unmasked. In this case, the cancellation handler is called, so there is a message in the console saying "release RESOURCE."

However, if we change **onCancel** to **cancelable**, the guarantee of resource release is no longer present. Furthermore, there is also a "USE RESOURCE" print, so it seems that cancelable makes the unmasked effect uncancelable altogether. Here is edited code:

```scala
import cats.effect.{IO, IOApp}

import scala.concurrent.duration.DurationInt

object TestCancelable extends IOApp.Simple {
  val acquire: IO[String] = IO.println("START AQ") >> IO.sleep(2.seconds) >> IO.println("END AQ") >> IO("RESOURCE")

  def evaluation: IO[Unit] = IO.uncancelable {
    unmask =>
      acquire.flatMap(
        r =>
          unmask {
            IO.sleep(5.seconds) >> IO(println(s"USE ${r}"))
          }.cancelable(IO(println(s"release ${r}")))
      )
  }

  override def run: IO[Unit] =
    for {
      f <- evaluation.start
      _ <- IO.sleep(1.seconds)
      _ <- f.cancel
      _ <- IO.never[Unit]
    } yield ()
}
```

TalkingFoxMid:
I believe we should use uncancelable in Doobie when changing:

```scala
createLogged
  .bracket(ps => IFC.embed(ps, prepLogged *> execAndProcessLogged))(ps => IFC.embed(ps, IFPS.close))
```

to:

```scala
WeakAsyncConnectionIO.uncancelable { unmask =>
  createLogged.flatMap { ps =>
    WeakAsyncConnectionIO.cancelable(
      IFC.embed(ps, prepLogged *> execAndProcessLogged).guaranteeCase {
        case Outcome.Canceled() => WeakAsyncConnectionIO.unit
        case _ => IFC.embed(ps, IFPS.close)
      },
      IFC.embed(ps, IFPS.close)
    )
  }
}
```

The bracket function guarantees that resource acquisition cannot be canceled, but as the previous example shows, cancelable behaves unexpectedly when used with an uncancelable mask on resource acquisition. This makes it unclear how to work around this situation.

TalkingFoxMid:
@armanbilge

I assume the issue arises because cancelable is being used inside another uncancelable block, making the poll(fiber.join) uncancelable due to the outer uncancelable. Consider this example where I've slightly reworked the cancelable method by adding an outer Poll argument. This solution works as expected:

```
import cats.effect.kernel.Poll
import cats.effect.{Async, IO, IOApp}
import scala.concurrent.duration.DurationInt

object TestCancelable extends IOApp.Simple {
  def cancelable[A](outer: Poll[IO])(fa: IO[A], fin: IO[Unit]): IO[A] = {
    IO.uncancelable { poll =>
      fa.start.flatMap { fiber =>
        poll(outer(fiber.join))
          .onCancel(fin *> fiber.cancel)
          .flatMap(_.embed(poll(outer(IO.canceled *> IO.never[A]))))
      }
    }
  }

  val acquire: IO[String] = IO.println("START AQ") >> IO.sleep(2.seconds) >> IO.println("END AQ") >> IO("RESOURCE")

  def evaluation: IO[Unit] = IO.uncancelable { unmask =>
    acquire.flatMap { r =>
      cancelable(unmask)(
        IO.sleep(5.seconds) >> IO(println(s"USE $r")),
        IO(println(s"release $r"))
      )
    }
  }

  override def run: IO[Unit] =
    for {
      f <- evaluation.start
      _ <- IO.sleep(1.second)
      _ <- f.cancel
      _ <- IO.never[Unit]
    } yield ()
}
```

This code prints:

```
START AQ
END AQ
release RESOURCE
```

Is there any way to overcome this using the current CE3 capabilities?

armanbilge:
@TalkingFoxMid

> However, if we change **onCancel** to **cancelable**

`onCancel` and `cancelable` are not equivalent. `task.onCancel(fin)` runs `fin` _after_ `task` is canceled. `task.cancelable(fin)` runs `fin` _concurrently_ while `task` is running.

---

> the guarantee of resource release is no longer present. Furthermore, there is also a "USE RESOURCE" print, so it seems that cancelable makes the unmasked effect uncancelable altogether. Here is edited code:

To be cancelable, the `cancelable` needs to be _inside_ the `unmask`.

```diff
@@ -10,8 +10,8 @@ object TestCancelable extends IOApp.Simple {
       acquire.flatMap(
         r =>
           unmask {
-            IO.sleep(5.seconds) >> IO(println(s"USE $r"))
-          }.onCancel(IO(println(s"release $r")))
+            (IO.sleep(5.seconds) >> IO(println(s"USE $r"))).onCancel(IO(println(s"release $r")))
+          }
       )
   }
```

After making this change the result is:
```
START AQ
END AQ
```

TalkingFoxMid:
> ```diff
> .onCancel(IO(println(s"release $r")))
> ```

Consider this example also:

```

object TestCancelable extends IOApp.Simple {
  val acquire: IO[String] = IO.println("START AQ") >> IO.sleep(2.seconds) >> IO.println("END AQ") >> IO("RESOURCE")

  def evaluation: IO[Unit] = IO.uncancelable {
    unmask =>
      acquire.flatMap(
        r =>
          unmask {
            IO(Thread.sleep(5)) >> IO(println(s"USE $r")).onCancel(IO(println(s"release $r")))
          }
      )
  }

  override def run: IO[Unit] =
    for {
      f <- evaluation.start
      _ <- IO.sleep(1.second)
      _ <- f.cancel
      _ <- IO.never[Unit]
    } yield ()
}
```

Here i wrote Thread.sleep, but there is no "release RESOURCE" in console also. But in case of doobie we need cancelation handler to be executed instantly after *f.cancel* is called - to cancel query in database

TalkingFoxMid:
> @TalkingFoxMid
> 
> > However, if we change **onCancel** to **cancelable**
> 
> `onCancel` and `cancelable` are not equivalent. `task.onCancel(fin)` runs `fin` _after_ `task` is canceled. `task.cancelable(fin)` runs `fin` _concurrently_ while `task` is running.
> 
> > the guarantee of resource release is no longer present. Furthermore, there is also a "USE RESOURCE" print, so it seems that cancelable makes the unmasked effect uncancelable altogether. Here is edited code:
> 
> To be cancelable, the `cancelable` needs to be _inside_ the `unmask`.
> 
> ```diff
> @@ -10,8 +10,8 @@ object TestCancelable extends IOApp.Simple {
>        acquire.flatMap(
>          r =>
>            unmask {
> -            IO.sleep(5.seconds) >> IO(println(s"USE $r"))
> -          }.onCancel(IO(println(s"release $r")))
> +            (IO.sleep(5.seconds) >> IO(println(s"USE $r"))).onCancel(IO(println(s"release $r")))
> +          }
>        )
>    }
> ```
> 
> After making this change the result is:
> 
> ```
> START AQ
> END AQ
> ```

Consider this example also:

```
object TestCancelable extends IOApp.Simple {
  val acquire: IO[String] = IO.println("START AQ") >> IO.sleep(2.seconds) >> IO.println("END AQ") >> IO("RESOURCE")

  def evaluation: IO[Unit] = IO.uncancelable {
    unmask =>
      acquire.flatMap(
        r =>
          unmask {
            IO(Thread.sleep(5)) >> IO(println(s"USE $r")).onCancel(IO(println(s"release $r")))
          }
      )
  }

  override def run: IO[Unit] =
    for {
      f <- evaluation.start
      _ <- IO.sleep(1.second)
      _ <- f.cancel
      _ <- IO.never[Unit]
    } yield ()
}
```

Here i wrote Thread.sleep, but there is no "release RESOURCE" in console also. But in case of doobie we need cancelation handler to be executed instantly after f.cancel is called - to cancel query in database. As I see in you example also - "release RESOURCE" isn't printing. 

The goal is to wrap resource acquiring in uncancelable (to prevent resource leak here) and write `cancelable` in a way which leads to instant cancel handler executing  (even if thread is blocked).

armanbilge:
@TalkingFoxMid thanks, I think there is still some confusion. `cancelable` is _not_ a drop-in equivalent for `onCancel`. They are different combinators which serve different purposes. You may even use them together:

```scala
object TestCancelable extends IOApp.Simple {
  val acquire: IO[String] = IO.println("START AQ") >> IO.sleep(2.seconds) >> IO.println("END AQ") >> IO("RESOURCE")

  def evaluation: IO[Unit] = IO.uncancelable {
    unmask =>
      acquire.flatMap(
        r =>
          unmask {
            (IO.sleep(5.seconds) >> IO(println(s"USE ${r}"))).cancelable(IO(println(s"canceling USE ${r}")))
          }.onCancel(IO(println(s"release ${r}")))
      )
  }

  override def run: IO[Unit] =
    for {
      f <- evaluation.start
      _ <- IO.sleep(1.seconds)
      _ <- f.cancel
      _ <- IO.never[Unit]
    } yield ()
}
```

```
START AQ
END AQ
release RESOURCE
```

TalkingFoxMid:
> @TalkingFoxMid thanks, I think there is still some confusion. `cancelable` is _not_ a drop-in equivalent for `onCancel`. They are different combinators which serve different purposes. You may even use them together:
> 
> ```scala
> object TestCancelable extends IOApp.Simple {
>   val acquire: IO[String] = IO.println("START AQ") >> IO.sleep(2.seconds) >> IO.println("END AQ") >> IO("RESOURCE")
> 
>   def evaluation: IO[Unit] = IO.uncancelable {
>     unmask =>
>       acquire.flatMap(
>         r =>
>           unmask {
>             (IO.sleep(5.seconds) >> IO(println(s"USE ${r}"))).cancelable(IO(println(s"canceling USE ${r}")))
>           }.onCancel(IO(println(s"release ${r}")))
>       )
>   }
> 
>   override def run: IO[Unit] =
>     for {
>       f <- evaluation.start
>       _ <- IO.sleep(1.seconds)
>       _ <- f.cancel
>       _ <- IO.never[Unit]
>     } yield ()
> }
> ```
> 
> ```
> START AQ
> END AQ
> release RESOURCE
> ```

Thanks for an answer. In your example there is "release RESOURCE" print due to onCancel but there is no "canceling USE" print - so `cancelable` isn't working i suppose. But we write not IO.sleep but Thread.sleep to simulate a fiber on a blocked thread  (this is exactly similar to case in doobie, we are trying to cancel a fiber which is running on a blocked thread) - onCancel is not longer execute effect instantly but `canceled` isn't working also! That is an issue i say about.

The complete example is

```
object TestCancelable extends IOApp.Simple {
  val acquire: IO[String] = IO.println("START AQ") >> IO.sleep(2.seconds) >> IO.println("END AQ") >> IO("RESOURCE")

  def evaluation: IO[Unit] = IO.uncancelable {
    unmask =>
      acquire.flatMap(
        r =>
          unmask {
            (IO.(Thread.sleep(5)) >> IO(println(s"USE ${r}"))).cancelable(IO(println(s"canceling USE ${r}")))
          }.onCancel(IO(println(s"release ${r}")))
      )
  }

  override def run: IO[Unit] =
    for {
      f <- evaluation.start
      _ <- IO.sleep(1.seconds)
      _ <- f.cancel
      _ <- IO.never[Unit]
    } yield ()
}
```

Prints

```
START AQ
END AQ
```

So cancelable is not working here, isn't it?

armanbilge:
> So cancelable is not working here, isn't it?

Strange, I was not able to replicate your result. It works for me. Here is the exact code I am running, which is directly copied from your example:

```scala
//> using dep org.typelevel::cats-effect::3.5.4

import cats.effect.{IO, IOApp}

import scala.concurrent.duration.DurationInt

object TestCancelable extends IOApp.Simple {
  val acquire: IO[String] = IO.println("START AQ") >> IO.sleep(2.seconds) >> IO.println("END AQ") >> IO("RESOURCE")

  def evaluation: IO[Unit] = IO.uncancelable {
    unmask =>
      acquire.flatMap(
        r =>
          unmask {
            (IO(Thread.sleep(5)) >> IO(println(s"USE ${r}"))).cancelable(IO(println(s"canceling USE ${r}")))
          }.onCancel(IO(println(s"release ${r}")))
      )
  }

  override def run: IO[Unit] =
    for {
      f <- evaluation.start
      _ <- IO.sleep(1.seconds)
      _ <- f.cancel
      _ <- IO.never[Unit]
    } yield ()
}
```

```
% scala-cli --server=false sandbox.scala
START AQ
END AQ
release RESOURCE
```

---

> But if we write not IO.sleep but Thread.sleep to simulate a fiber on a blocked thread (this is exactly similar to case in doobie, we are trying to cancel a fiber which is running on a blocked thread)

Note, the correct way to write a fiber on a blocked thread is `IO.blocking(Thread.sleep(...))` (or `IO.interruptible` because `sleep` support cancelation via `Thread#interrupt`).

satorg:
@TalkingFoxMid ,
> The complete example is
> ...
> Prints
> ...

Actually, your example (after I fix `IO.(Thread.sleep(5))` to `IO(Thread.sleep(5))`) prints the following:
```
START AQ
END AQ
release RESOURCE
```
I.e. the callback effect `fin` registered in `unmask { ... }.onCancel( fin )` gets triggered.

The reason why the `fin` effect provided to `cancelable` is not executed in your example is that due to the specific timings the cancelation occurs at the boundary of the unmasked region. I.e. the task gets canceled _before_ is gets into `cancelable`.

However, if you change your timings to the following values:
<details>

```scala
  val acquire: IO[String] =
    IO.println("START AQ") >> IO.sleep(2.seconds) >> IO.println("END AQ") >> IO("RESOURCE")

  def evaluation: IO[Unit] = IO.uncancelable { unmask =>
    acquire.flatMap(r =>
      unmask {
        // Sleep time increased to 2 seconds (was 5 ms)
        (IO(Thread.sleep(2000)) >> IO(println(s"USE $r")))
          .cancelable(IO(println(s"canceling USE $r")))
      }.onCancel(IO(println(s"release ${r}")))
    )
  }

  override def run: IO[Unit] =
    for {
      f <- evaluation.start
      _ <- IO.sleep(3.seconds) // increased to 3 seconds
      _ <- f.cancel
      _ <- IO.never[Unit]
    } yield ()

```

</details>
then the app prints the following:

```
START AQ
END AQ
canceling USE RESOURCE
release RESOURCE
```
Here the cancel signal comes _inside_ the unmasked region.
Therefore `cancelable` is applied and the cancel signal takes an effect right after `Thread.sleep` completes (which cannot be interrupted).

TalkingFoxMid:
> def cancelable[A](outer: Poll[IO])(fa: IO[A], fin: IO[Unit]): IO[A] = {
>     IO.uncancelable { poll =>
>       fa.start.flatMap { fiber =>
>         poll(outer(fiber.join))
>           .onCancel(fin *> fiber.cancel)
>           .flatMap(_.embed(poll(outer(IO.canceled *> IO.never[A]))))
>       }
>     }
>   }

Sorry for confusion and thank you all for trying to explain, it was very helpful. Now I see that there is cases when `onCancel` works and `cancelable` is not working, and vice versa (even if our effect is running on a blocked thread). So the most appropriate way to guarantee PreparedStatement cancelation I suppose is to use `onCancel` and `cancelable` together, to cover all possible cases of cancelation after resource was created.

However there is also small issue: call duplication. There are some cases when `cancelable` and `onCancel` can be executed both and this causes the effect to execute twice. It might worth if effect isn't idempotent (if we want guarantee exactly-once effect completing) but in case of PreparedStatement I suppose there is no such problem

TalkingFoxMid:
@satorg @armanbilge @jatcwang 

I've finally done a solution that seems to be complete now and looking forward for your comments/approves

 As for streams, I suggest doing this in another PR later

modules/core/src/test/scala/doobie/util/UpdateLogSuite.scala
```diff
@@ -18,27 +18,29 @@ class UpdateLogSuite extends munit.FunSuite {
 
   import cats.effect.unsafe.implicits.global
 
-  val ioLocal: IOLocal[LogEvent] =
-    IOLocal[LogEvent](null).unsafeRunSync()
+  val logEventRef: Ref[IO, LogEvent] =
+    Ref.of[IO, LogEvent](null).unsafeRunSync()
```
TalkingFoxMid, line 22: IOLocal is no longer works in logHandler because there is `cancelable` which forks a fiber in a middle of query processing

TalkingFoxMid: Shall we guarantee that IOLocal works with logHandler ?

modules/hikari/src/test/scala/doobie/HikariQueryCancellationSuite.scala
```diff
@@ -0,0 +1,56 @@
...
+import cats.effect.*
+import cats.effect.unsafe.implicits.global
+import com.zaxxer.hikari.HikariConfig
+import doobie.hikari.HikariTransactor
+import doobie.implicits.*
+import doobie.util.transactor
+
+import scala.concurrent.duration.DurationInt
+
+class HikariQueryCancellationSuite extends munit.FunSuite {
```
TalkingFoxMid, line 16: I've remove QueryCancellationSuite in favor of more appropriate HikariQueryCancellationSuite because current solution doesn't work well with QueryCancellationSuite. Here is code for history:

```
// Copyright (c) 2013-2020 Rob Norris and Contributors
// This software is licensed under the MIT License (MIT).
// For more information see LICENSE or https://opensource.org/licenses/MIT

package doobie.util

import cats.effect.IO
import doobie.Transactor
import doobie.*
import doobie.implicits.*
import cats.syntax.all.*

import scala.concurrent.duration.DurationInt

class QueryCancellationSuite extends munit.FunSuite {
  import cats.effect.unsafe.implicits.global

  val xa = Transactor.fromDriverManager[IO](
    driver = "org.h2.Driver",
    url = "jdbc:h2:mem:queryspec;DB_CLOSE_DELAY=-1;DEFAULT_LOCK_TIMEOUT=60000;LOCK_TIMEOUT=60000",
    user = "sa",
    password = "",
    logHandler = None
  )


  test("Query cancel") {
    val scenario = WeakAsync.liftIO[ConnectionIO].use { elevator =>
      for {
        _ <- sql"CREATE TABLE IF NOT EXISTS example_table ( id INT)".update.run.transact(xa)
        _ <- sql"TRUNCATE TABLE example_table".update.run.transact(xa)
        _ <- sql"INSERT INTO example_table (id) VALUES (1)".update.run.transact(xa)
        _ <- {
          sql"select * from example_table for update".query[Int].unique >> elevator.liftIO(IO.never)
        }.transact(xa).start

        insertWithLockFiber <- {
          for {
            _ <- IO.sleep(100.milli)
            insertFiber <- sql"UPDATE example_table SET id = 2".update.run.transact(xa).start
            _ <- IO.sleep(100.milli)
            _ <- insertFiber.cancel

          } yield ()
        }.start

        _ <- IO.race(insertWithLockFiber.join, IO.sleep(9.seconds) >> IO(fail("Cancellation is blocked")))
        result <- sql"SELECT * FROM example_table".query[Int].to[List].transact(xa)
      } yield assertEquals(result, List(1))
    }
    scenario.unsafeRunSync()
  }
}

```

modules/core/src/test/scala/doobie/util/UpdateLogSuite.scala
jatcwang: Ouch yes we probably need to. One of our docuemented examples also uses IOLocal to pass additional context to LogHandler.
In a perfect world, I do feel like `IO.cancelable` implementation should pass on the IOLocal though..

armanbilge:
> Now I see that there is cases when `onCancel` works and `cancelable` is not working, and vice versa (even if our effect is running on a blocked thread).

Sorry, what is an example of a case where `cancelable` runs but `onCancel` does not run (this is the "vice versa")? That seems problematic.

modules/core/src/test/scala/doobie/util/UpdateLogSuite.scala
armanbilge: > In a perfect world, I do feel like `IO.cancelable` implementation should pass on the IOLocal though..

In general, you cannot make these kinds of assumptions when using `IOLocal` with combinators. Further reading:
- https://github.com/typelevel/cats-effect/issues/3100
- https://github.com/typelevel/cats-effect/issues/3385

modules/core/src/main/scala/doobie/hi/connection.scala
```diff
@@ -175,7 +175,11 @@ object connection {
     }
 
     createLogged
-      .bracket(ps => IFC.embed(ps, prepLogged *> execAndProcessLogged))(ps => IFC.embed(ps, IFPS.close))
+      .bracket(ps =>
+        WeakAsyncConnectionIO.cancelable(
+          IFC.embed(ps, prepLogged *> execAndProcessLogged),
+          IFC.embed(ps, IFPS.close)
+        ))(IFC.embed(_, IFPS.close) )
```
armanbilge, line 182: this looks right 👍

TalkingFoxMid:
> > Now I see that there is cases when `onCancel` works and `cancelable` is not working, and vice versa (even if our effect is running on a blocked thread).
> 
> Sorry, what is an example of a case where `cancelable` runs but `onCancel` does not run (this is the "vice versa")? That seems problematic.

Sorry for confusion. There is exactly one case I mean:



> > Now I see that there is cases when `onCancel` works and `cancelable` is not working, and vice versa (even if our effect is running on a blocked thread).
> 
> Sorry, what is an example of a case where `cancelable` runs but `onCancel` does not run (this is the "vice versa")? That seems problematic.

I mean exactly one case - cancelation on blocked thread. It prints nothing even after ten seconds.

```
object TestCancelable extends IOApp.Simple {

  def task: IO[Unit] = IO.blocking(Thread.sleep(10000))
    .onCancel(IO(println("onCancel run")))

  override def run: IO[Unit] = for {
    fiber <- task.start
    _ <- IO.sleep(1.seconds)
    _ <- fiber.cancel
    _ <- IO.never[Nothing]
  } yield ()
}
```

So my point is to guarantee at-least-once cancelation we must use `cancelable` and `onCancel` both. Am I missing something?

TalkingFoxMid:
> > Now I see that there is cases when `onCancel` works and `cancelable` is not working, and vice versa (even if our effect is running on a blocked thread).
> 
> Sorry, what is an example of a case where `cancelable` runs but `onCancel` does not run (this is the "vice versa")? That seems problematic.

Sorry for confusion. I mean that if we want finalizer to run instantly after we cancel a fiber - we must use `cancelable` and `onCancel` together. So there is two possible branches: if the fiber is on a blocked thread then `cancelable` will run instantly but if the fiber continues running `runLoop` then `onCancel` will run finalizer instantly

modules/core/src/test/scala/doobie/util/UpdateLogSuite.scala
TalkingFoxMid: I think there are a lot of problems with maintaining this feature:
1) It is not very functional. I think the most appropriate way to do this - use writer monads, and I suppose that is possible in current setup.
2) It restricts ours capabilities to rely on CE3 functional combinators.

To be honest, I think that IO should be used exclusively in the internal code of an application (or library). The API of a functional library should not provide (or at least guarantee that it will work fine) the use of this tool to users

TalkingFoxMid: @jatcwang 

Please give a reference to place in the documentation where IOLocal is used. IOLocal can be used to pass context to LogHandler. But It can't be used to modify this context by logHandler (in tests loghandler modifies IOLocal context). I've found this example in doc:

```scala mdoc
import cats.effect.{IOLocal, Ref}
import doobie.util.log.Success

def users = List.range(0, 4).map(n => s"user-$n")

def program: IO[List[String]] =
  for {
    // define an IOLocal where we store the user which caused the query to be run  
    currentUser <- IOLocal("")
    // store all successful sql here, for all users
    successLogsRef <- Ref[IO].of(List.empty[String])
    xa = Transactor.fromDriverManager[IO](
        driver = "org.h2.Driver",
        url = "jdbc:h2:mem:queryspec;DB_CLOSE_DELAY=-1",
        user = "sa", 
        password = "",
        logHandler = Some(new LogHandler[IO] {
          def run(logEvent: LogEvent): IO[Unit] =
            currentUser.get.flatMap(user => successLogsRef.update(logs => s"sql for user $user: '${logEvent.sql}'" :: logs))
        })
      )
    // run a bunch of queries
    _ <- users.parTraverse(user =>
      for {
        _ <- currentUser.set(user)
        _ <- sql"select 1".query[Int].unique.transact(xa)
      } yield ()
    )
    // return collected log messages
    logs <- successLogsRef.get
  } yield logs

program.unsafeRunSync().sorted
```

It will work because child fibers inherit context of parents. So if we set up context in parrent fiber it can be accessed in loghandler. But if we modify context in loghandler this modification can't be propagated to a parent fiber. In tests there is modification of context in child fiber

TalkingFoxMid: Consider these two examples:

```
import cats.effect.{IO, IOApp, IOLocal}

object IOCancelTest extends IOApp.Simple {

  def childTask(local: IOLocal[Int]): IO[Unit] =
    for {
      _ <- local.get.flatTap(IO.println) // 42
    } yield ()

  override def run: IO[Unit] = for {
    local <- IOLocal(42)
    _ <- childTask(local).cancelable(IO(println("some finalizer")))
  } yield ()
}
```

```
import cats.effect.{IO, IOApp, IOLocal}

object IOCancelTest2 extends IOApp.Simple {

  def childTask(local: IOLocal[Int]): IO[Unit] =
    for {
      _ <- local.set(64)
    } yield ()

  override def run: IO[Unit] = for {
    local <- IOLocal(42)
    _ <- childTask(local).cancelable(IO(println("some finalizer")))
    _ <- local.get.flatTap(IO.println) // 42
  } yield ()
}
```

satorg:
@TalkingFoxMid ,
> ... if we want finalizer to run instantly after we cancel a fiber - we must use cancelable and onCancel together. So there is two possible branches ...

I feel there's still a confusion. `onCancel` and `cancelable` are very different combinators that serve different purposes. We cannot talk about them in terms of "either ... or ... ".

- `onCancel` simply lets us know when (and whether) cancelation occurred. Its `fin` effect gets executed _after_ the cancellation happened. Therefore, `onCancel` _does not_ affect the behavior of the effect it is applied to.
- `cancelable` allows us to convert otherwise a "non-cancelable" effect into a "cancelable" one (it does not matter whether it is blocking or not). The `cancelable`'s `fin` effect it the key for the conversion – it should contain an action that interrupts the target "non-cancelable" effect. Therefore, `cancelable` _does_ affect the behavior of the effect it is applied to.

armanbilge:
Yes, and to add on to that a bit: I would not say that `cancelable` runs a "finalizer", it runs an "interrupt" for the primary task. If you want to run a finalizer, you must use a combinator such as `guarantee` or `onCancel`.

satorg:
@armanbilge , 
> I would not say that cancelable runs a "finalizer", it runs an "interrupt" for the primary task.

I guess the confusion might come from the fact, "finalizer" is the name used in the docs for `cancelable`:

https://github.com/typelevel/cats-effect/blob/caee0e86f815b641af61f19becd4eac27a769f11/core/shared/src/main/scala/cats/effect/IO.scala#L400-L442

> Given an effect which might be [[uncancelable]] and a finalizer, produce an effect which can be canceled by running the finalizer. 

Also, its parameter name is `fin` similar to one used for `onCancel`.

modules/core/src/test/scala/doobie/util/UpdateLogSuite.scala
jatcwang: Yep you've found the example in the docs :+1: 

I think trying to modify IOLocal within a ConnectionIO should be quite rare. Since this is a general issue with IOLocal propagation it's not something we can solve here. 

Thanks for updating the tests to use Ref :+1:

jatcwang (approved):
This LGTM thanks @TalkingFoxMid! Very much appreciate the reviews too @armanbilge @satorg

PR #2085, opened 2024-08-27
Introduces attemptSomeSql into ApplicativeErrorOps

prascuna:
Introduces an `attemptSomeSql` catcher to partial match on a specific `SQLException`.
The use case for it is to match both the stage, but also log the exception itself:

```scala
        .attemptSomeSql {
          case e: PSQLException if SqlState(e.getSQLState) == sqlstate.class23.UNIQUE_VIOLATION =>
            logger.error("Database Unique Violation", e)
            DbCreateError.UniqueViolation
          case e: PSQLException if SqlState(e.getSQLState) == sqlstate.class23.CHECK_VIOLATION  =>
            logger.error("Database Check Violation", e)
            DbCreateError.CheckViolationOnCreate
        }

```

For some reason `attemptSomeSql` catcher is mentioned in the docs but not implemented.

Also, I've changed the docker-compose mysql image to not rely on the debian version, as that's not available for arm64 devs

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
 
 
   mysql:
-    image: mysql:8.0-debian
+    image: mysql:8.0
     environment:
       MYSQL_ROOT_PASSWORD: password
       MYSQL_DATABASE: world
--- a/modules/core/src/main/scala/doobie/syntax/applicativeerror.scala
+++ b/modules/core/src/main/scala/doobie/syntax/applicativeerror.scala
@@ -11,6 +11,7 @@ import java.sql.SQLException
 
 class ApplicativeErrorOps[M[_], A](self: M[A])(implicit ev: ApplicativeError[M, Throwable]) {
   def attemptSql: M[Either[SQLException, A]] = C.attemptSql(self)
+  def attemptSomeSql[B](f: PartialFunction[SQLException, B]): M[Either[B, A]] = C.attemptSomeSql(self)(f)
   def attemptSqlState: M[Either[SqlState, A]] = C.attemptSqlState(self)
   def attemptSomeSqlState[B](f: PartialFunction[SqlState, B]): M[Either[B, A]] = C.attemptSomeSqlState(self)(f)
   def exceptSql(handler: SQLException => M[A]): M[A] = C.exceptSql(self)(handler)
--- a/modules/core/src/main/scala/doobie/util/catchsql.scala
+++ b/modules/core/src/main/scala/doobie/util/catchsql.scala
@@ -22,6 +22,12 @@ object catchsql {
       case e: SQLException => e.asLeft
     }
 
+  /** Like `attempt` but catches only the defined `SQLException`. */
+  def attemptSomeSql[M[_], A, B](ma: M[A])(f: PartialFunction[SQLException, B])(implicit AE: ApplicativeError[M, Throwable]): M[Either[B, A]] =
+    ma.map(_.asRight[B]).recoverWith { case e: SQLException =>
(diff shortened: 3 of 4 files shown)


satorg:
Thank you for the contribution!

Just a note though: since `SQLException` is a `Throwable`, it does not make a lot of sense to restrict `PartialFunction` to that type only. For example, a combinator with the following signature
```scala
def attemptYOUNAMEIT[M[_], A, B](ma: M[A])(f: PartialFunction[Throwable, B]): M[Either[B, A]] = ???
```
would do here just as well:
```scala
.attemptYOUNAMEIT {
  case e: PSQLException if SqlState(e.getSQLState) == sqlstate.class23.UNIQUE_VIOLATION =>
    logger.error("Database Unique Violation", e)
    DbCreateError.UniqueViolation
  case e: PSQLException if SqlState(e.getSQLState) == sqlstate.class23.CHECK_VIOLATION  =>
    logger.error("Database Check Violation", e)
    DbCreateError.CheckViolationOnCreate
}
```
Apparently, the `attemptYOUNAMEIT` combinator is not restricted to Doobie/JDBC domain only – it can work with any other exception type.

Therefore, I would suggest to propose that kind of combinator straight to the [Cats](https://github.com/typelevel/cats) library so that everyone besides Doobie could benefit from that. Then, Doobie users can get it automatically from Cats.

jatcwang:
Thanks for the contribution @prascuna! I believe cat's [attemptNarrow](https://github.com/typelevel/cats/blob/main/core/src/main/scala/cats/syntax/applicativeError.scala#L80) does exactly this so probably don't need to be added to Doobie?

Will take the docker related changes though :smile:

prascuna:
Hey folks, thank you for your feedback!

@satorg you're right, there's no need to restrict it to just `SQLException`, I could give a go raising a PR into Cats 



> I believe cat's [attemptNarrow](https://github.com/typelevel/cats/blob/main/core/src/main/scala/cats/syntax/applicativeError.scala#L80) does exactly this so probably don't need to be added to Doobie?
@jatcwang it's not quite the same, as I'm not just interested in narrowing the type, but also the value (namely, the sql state).  


If I were to implement it with `attemptNarrow` without using `EitherT` would look pretty ugly:

```scala
        .attemptNarrow[PSQLException]
        .flatMap[Either[DbCreateError, MyType]] {
          case Left(e)      =>
            e match {
              case e: PSQLException if SqlState(e.getSQLState) == sqlstate.class23.UNIQUE_VIOLATION =>
                logger.error("Database Unique Violation", e)
                (DbCreateError.UniqueViolation: DbCreateError).asLeft[MyType].pure[ConnectionIO]
              case e: PSQLException if SqlState(e.getSQLState) == sqlstate.class23.CHECK_VIOLATION  =>
                logger.error("Database Check Violation", e)
                (DbCreateError.CheckViolationOnCreate: DbCreateError).asLeft[MyType].pure[ConnectionIO]
              case e                                                                                => 
                ApplicativeError[ConnectionIO, Throwable].raiseError(e)
            }
          case Right(value) => value.asRight[DbCreateError].pure[ConnectionIO]
        }

```


Shall we close this PR then? (and I can raise another one for Docker?)

jatcwang:
@prascuna I see what you mean. The best I can come up with is 

```
            .attemptT
            .leftSemiflatMap {
              case e: SQLException if ...=> e.pure
              case e => e.raiseError
            }
            .value
```
which does use EitherT and it isn't obvious what `leftSemiflatMap` does at first glance (Seems like type inference works well here though).

Do agree with @satorg that this seems useful enough to try to put into cats. Let's close and reopen if cats maintainers doesn't want to add it :)

PR #2088, opened 2024-09-01
Query cancellation for streaming queries

jatcwang:
`Stream.bracket` does not allow cancellation in the `acquire` step which is why we're using `bracketFull` instead


satorg (approved):
Great, thanks!

PR #2089, opened 2024-09-02
Update microsite link

bartma11:
I noticed that links in the README.md stopped working. Once I saw the discussion about moving the repository into the typelevel org, I figured out why. I replaced the old link with `https://typelevel.org/doobie/`. Since it would be more annoying to open an issue than it was to fix this, I created this PR straight away. If this is not okay then feel free to ignore it and sorry for the trouble.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 **doobie** is a pure functional JDBC layer for Scala.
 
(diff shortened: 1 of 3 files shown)


jatcwang (approved):
Hey thank you this is quick! :smile: 
We do have a redirect PR setup so once it's merged the old links should work, but linking to the canonical site as you did here is preferred of course!

PR #2128, opened 2024-11-03
Proposal: `IN` and `NOT IN` fragment builders for product types with arbitrary arities

satorg:
Current `in`, `notIn`, `inOpt`, etc fragment builders only allow 1 or 2 values per matching row.
The proposed builders take `Read` instead of `Get` and therefore allow products of any arity.
Moreover, the current 2-arity builders accept values packed as `Tuple2` only. The proposed builders accept tuples as well as case classes, which can come in handy in some cases.

#### Why `inValues` and `notInValues` names?

In fact, neither the existing nor the proposed builders allow to build full-featured `IN`/`NOT IN` expression. Both allow just a subset of features – matching against a static set of values. Besides, we cannot have `in` function overloads for both kinds of builders. Therefore, if we want to keep the existing builders, we'll have to choose another names for the new ones.

Names `inValues` and `notInValues` look pretty reasonable, but I'm totally open to change them to something better.

### DIFF
--- a/modules/core/src/main/scala/doobie/util/fragments.scala
+++ b/modules/core/src/main/scala/doobie/util/fragments.scala
@@ -6,7 +6,6 @@ package doobie
 package util
 
 import cats.Foldable
-import cats.Functor
 import cats.Reducible
 import cats.data.NonEmptyList
 import cats.syntax.all.*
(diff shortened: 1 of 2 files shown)


satorg:
Hmm.. CI has failed, but doesn't seem related to the PR to me 🤔

modules/core/src/main/scala/doobie/util/fragments.scala
```diff
@@ -45,6 +47,45 @@ object fragments {
     NonEmptyList.fromFoldable(fs).map(nel => notIn(f, nel))
   }
 
+  @inline
+  private def inNotInValues[F[_]: Reducible: Functor, A](
+      fwop: Fragment,
+      fs: F[A],
+      fallback: Fragment
+  )(implicit A: util.Write[A]): Fragment = {
+    if (A.length == 0) fallback
```
jatcwang, line 56: Thanks this looks good! Small suggestion - I don't think we need/should handle 0-arity products.

Reason:
- We no longer derive instances for case objects so it's really difficult to accidentally pass in a 0-arity product 
- Which means the only other possible case is if user is passing `List[Unit]`, which never makes sense and it's better to generate a bad SQL that fails than silently fallback to TRUE/FALSE.

WDYT?

satorg: Agree, I don't like it either. I'll update the PR, thx!

modules/core/src/main/scala/doobie/util/fragments.scala
```diff
@@ -45,6 +47,45 @@ object fragments {
...
+
+  /** Returns `f IN (fs0, fs1, ...)`.
+    * @param f
+    *   left-hand expression.
+    * @param fs
+    *   values of `Product` type to compare to the left-hand expression.
+    * @return
+    *   the `IN` subquery expression or `FALSE` if `fs` is a 0-arity product.
+    */
+  def inValues[F[_]: Reducible: Functor, A: util.Write](f: Fragment, fs: F[A]): Fragment =
```
jatcwang, line 75: I'm also wondering whether we can replace `in` with this. It should be source compatible

satorg: Do you mean, remove the existing `in`/`notIn` whatsoever and rename the proposed `inValues`/`notInValues` to `in`/`notIn`, correct?

jatcwang: Yes that's right

satorg:
Hi @jatcwang ,
I've updated the PR:
1. removed checks for 0 arity;
2. removed all previous versions of `in/inOut/notIn/notInOpt` and replaced them with corresponding methods that take `Write` instead of `Put` (all of them);
3. fixed/updated/enhanced corresponding tests.

jatcwang:
Looks great tyvm!

Issue #2104, opened 2024-09-19
1.0.0-RC6: `import doobie.implicits._` is not source compatible.


notxcain:
It looks like the imported auto-derivation takes precedence over explicitly defined instances available in scope when `Option` is involved.

Reproducer:

[Scastie](https://scastie.scala-lang.org/gvGT8vxVQPWUhjduDKkH1A)

```scala
case class Person(
  id: Int,
  name: String,
  address: Option[Address]
)

case class Address(
  city: String,
  street: String,
  zip: Int
)

trait Codecs {
  import doobie._
  
  // Ignore the correctness
  implicit val addressGet: Get[Address] = 
    Get[String].map(_ => Address("", "", 1))
}

import doobie._
import doobie.implicits._

object Dao extends Codecs {
  val autoderived: Read[Person] = Read[Person]
}

assert(Dao.autoderived.length == 3)

// Dao.autoderived.length is 5

```

notxcain:
Got even shorter [reproducer](https://scastie.scala-lang.org/BERFfSQYQ7SHtwixSaeDhQ). The issue is with the `Option`.

jatcwang:
Thank you for the report & repro @notxcain!
Someone else have the same issue too https://discord.com/channels/632277896739946517/632727524434247691/1286285929690169415

I think I know what needs to be done to fix this 🔨

sideeffffect:
Hello @jatcwang , at our company, we're loving users of Doobie. We were considering to update to RC6, but it seems like it's better to wait for RC7. Do you think RC7 will be released sometime this or next week?

jatcwang:
Hey @sideeffffect I'm knee deep in implicit prioritization land right now trying to sort it out 😄
Yes I think waiting for RC7 is the better idea unless you're not dependent on any custom `Meta/Read/Write` instances.

I'm currently trying a variation of Circe's `Exported` to fix this. Can't promise a timeline though sorry - These things either work or don't. Will push some changes for ideas/review if I'm stuck.

jatcwang:
So..the good news is the general approach works. However, while testing I discovered that because doobie has always relied on auto-derivation. You can not simply take a `Read/Write[A]` instance and produce `Read/Write[Option[A]]`. The latter are have to be completely derived from scratch.
That's something we can change but involves changing the implementation of Write / Read.. :hammer:

jsoref:
[...]

PR #2136, opened 2024-11-13
Rework Read/Write & Fix derivation to use custom instances

### LINKED ISSUE #1054: implicit Read[Option[A]] - compilation times
Using `scalac-profiling`, I was able to generate a flamegraph to profile the compilation of my scala project. With great success, I was able to dramatically reduce compilation times by creating implicit `Read`s for my scala composite case classes.

However, it seems I'm not able to do this for `Option` types.

If I have some composite type `A`, the `Read[Option[A]]` is automatically generated by Doobie. Which is cool, except that it seems there's no way for me to write `Read[Option[A]]` myself to avoid the extra compilation process.

Even when I've created the `Read[A]` myself, it seemingly doesn't effect the compilation times of generating `Read[Option[A]]`.

Is it possible to manually write the `Read[Option[A]]`, if so, how?

To summarize, my flamegraph says that:
```
doobie.util.Read[Option[com.panono.db.DBAlbum]] (expanded macros 0)  (2,546,205 Î¼s, 5.65%)
```
even when there exists a `Read[DBAlbum]`.

jatcwang:
Fix both semiauto and automatic derivation to use custom defined Read/Write instances (e.g. in companion objects).

A complete rework of Read and Write is unfortunately necessary because with the previous implementation, we cannot simply obtain a `Read[Option[A]]` given a `Read[A]` - we'd need to derive `Option[A]` from scratch by resolving `Read[Option[X]]` instances for each of `A`'s columns.

After the rework, both `Read` and `Write` are now sealed traits, each with 3 subclasses:
- Single: Wrapper over a `Get/Put`
- SingleOpt: Wrapper over a `Get/Put`, but is nullable i.e. `Read[Option[A]]`, `Write[Option[A]]`
- Composite: A composite of `Read/Write` instances

Apart from enabling proper semiauto and automatic derivation, the rework also brings these benefits:

- Derivation rules are much simpler (which also translates to faster compile times). In particular, given a `Read/Write[A]` we can trivially derive `Read/Write[Option[A]]`.
- We now correctly handle optionality for `Option[CaseClassWithOptionalAndNonOptionalFields]`. More tests will be added for this in a follow up PR to demonstrate

Other notes:
- Semiauto and Auto derivation of unary product type (e.g. 1 element case class) are removed due to it causing auto derivation to pick the wrong path. It seems unnecessary since Write/Read derivation will yield the same behaviour anyway?

Fixes #1054, #2104


modules/core/src/main/scala/doobie/util/read.scala
```diff
@@ -42,67 +44,120 @@ and similarly with Get:
...
+
+    override def unsafeGet(rs: ResultSet, startIdx: Int): A = {
+      import scala.collection.mutable
+      val accum = mutable.ArrayBuffer.empty[Any]
+      var idx = startIdx
+      readInstances.foreach { rr =>
+        accum += rr.unsafeGet(rs, idx)
+        idx += rr.length
+      }
+      construct(accum.toList)
```
guymers, line 126: If `accum` is being turned into a list is it more efficient to use `ListBuffer` instead of `ArrayBuffer`?

modules/core/src/main/scala/doobie/util/read.scala
```diff
@@ -42,67 +44,120 @@ and similarly with Get:
...
+
+    override def toOpt: Read[Option[A]] = SingleOpt(get)
+
+    override def gets: List[(Get[?], NullabilityKnown)] = List(get -> NoNulls)
+
+    override val length: Int = 1
+
+  }
+
+  /** Simple instance wrapping a Put. i.e. single column nullable value */
```
guymers, line 100: "Put" -> "Get"?

modules/core/src/main/scala/doobie/util/read.scala
```diff
@@ -42,67 +44,120 @@ and similarly with Get:
...
+  implicit val optionUnit: Read[Option[Unit]] = Read.Composite(Nil, _ => Some(()))
 
-  implicit val optionUnit: Read[Option[Unit]] =
-    Read(Nil, (_, _) => Some(()))
+  implicit def fromReadOption[A](implicit read: Read[A]): Read[Option[A]] = read.toOpt
 
-  implicit def fromGet[A](implicit ev: Get[A]): Read[A] =
-    new Read(List((ev, NoNulls)), ev.unsafeGetNonNullable) {}
+  /** Simple instance wrapping a Put. i.e. single column non-null value */
+  final case class Single[A](get: Get[A]) extends Read[A] {
```
guymers, line 88: Is there any need for this to be a `case class` instead of a `class`? It has no concept of equality and its not data.

modules/core/src/test/scala/doobie/util/ReadSuite.scala
```diff
@@ -121,7 +151,7 @@ class ReadSuite extends munit.FunSuite with ReadSuitePlatform {
     // This result doesn't seem ideal, because we should know that Int isn't
     // nullable, so the correct result is Some((1, None, 3, None))
     // But with how things are wired at the moment this isn't possible
```
guymers, line 153: Could remove this comment now

guymers (approved):

modules/core/src/main/scala/doobie/util/read.scala
jatcwang: Nice - that seems to give a 20% boost in the benchmarks :)

modules/core/src/main/scala/doobie/util/read.scala
jatcwang: Fixed thanks

modules/core/src/main/scala/doobie/util/read.scala
jatcwang: Thanks I've changed this for for both Read & Write

modules/core/src/test/scala/doobie/util/ReadSuite.scala
jatcwang: Removed

build.sbt
```diff
@@ -98,9 +99,12 @@ lazy val compilerFlags = Seq(
...
   ),
-//  Test / scalacOptions --= Seq(
-//    "-Xfatal-warnings"
-//  ),
+  // Disable warning when @nowarn annotation isn't suppressing a warning
+  // to simplify cross-building
+  // because 2.12 @nowarn doesn't actually do anything.. https://github.com/scala/bug/issues/12313
+  scalacOptions ++= Seq(
+    "-Wconf:cat=unused-nowarn:s"
+  ),
```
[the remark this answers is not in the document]
jatcwang: Thanks I didn't know about the scalac-compat. I gave it a go but unfortunately somehow while compiling in Scala 3 it thinks `@nowarn213` is unused :/ 

```
[warn] -- [E198] Unused Symbol Warning: /home/jacob/proj/contrib/doobie/modules/core/src/test/scala/doobie/util/ReadSuite.scala:14:45 
[warn] 14 |import org.typelevel.scalaccompat.annotation.nowarn213
[warn]    |                                             ^^^^^^^^^
[warn]    |                                             unused import
```

I was being used in https://github.com/typelevel/doobie/blob/25dfdd3855c95dc1290dba4924b1ecff0063a776/modules/core/src/test/scala/doobie/util/ReadSuite.scala#L122 (I changed `nowarn` to `nowarn213`)

modules/core/src/main/scala/doobie/util/read.scala
```diff
@@ -42,67 +43,133 @@ and similarly with Get:
...
+  class Transform[A, From](underlyingRead: Read[From], f: From => A) extends Read[A] {
+    override def unsafeGet(rs: ResultSet, startIdx: Int): A = f(underlyingRead.unsafeGet(rs, startIdx))
+    override def gets: List[(Get[?], NullabilityKnown)] = underlyingRead.gets
+    override def toOpt: Read[Option[A]] =
+      new Transform[Option[A], Option[From]](underlyingRead.toOpt, opt => opt.map(f))
+    override lazy val length: Int = underlyingRead.length
+  }
+
+  /** A Read instance consists of multiple underlying Read instances */
+  class Composite[A, S0, S1](read0: Read[S0], read1: Read[S1], f: (S0, S1) => A) extends Read[A] {
```
jatcwang, line 127: After some benchmarking, I have reworked `Composite` to be structured more like `HList`s rather than `List[Read[Any]]`. 
With `List[Read[Any]]` my understanding is that JVM cannot effectively perform inlining because due to the dynamic length/type nature of the list.
But with a HList-like structure (e.g. `Composite(Read[Int], Composite(Read[String], Read[Int]))`) the call graph and types are "constant" which I think made the JIT happier

Using benchmark from #2139, the current implementation has basically no penalty between Read[A] and Read[Option[A]], whereas the previous implementation has around a 50% performance penalty.

```
[info] Benchmark                                Mode  Cnt        Score        Error  Units
[info] LargeRow.autoDerivedComplex             thrpt    5   956584.142 ± 105909.280  ops/s
[info] LargeRow.autoDerivedComplexOpt          thrpt    5   925193.390 ± 240656.559  ops/s
[info] LargeRow.rawJdbcComplex                 thrpt    5  1390634.422 ± 195647.819  ops/s
[info] LargeRow.rawJdbcTuple                   thrpt    5  1341330.033 ± 248449.784  ops/s
[info] LargeRow.semiautoDerivedComplex         thrpt    5   957341.195 ± 155499.630  ops/s
[info] LargeRow.semiautoDerivedComplexOpt      thrpt    5   917308.756 ± 181313.503  ops/s
[info] LargeRow.tuple                          thrpt    5  1068742.195 ± 208005.075  ops/s
[info] LargeRow.tupleOpt                       thrpt    5  1072509.030 ± 177644.605  ops/s
```

PR #2137, opened 2024-11-13
Prefetch a chunk of result for stream operation

wb14123:
Resolves https://github.com/typelevel/doobie/issues/2132.
    
Use fs2 Stream's `prefetchN` to buffer query results so that slow downstream operations don't slow down transaction commit.
    
The buffered result size is `chunkSize` (equals to `fetchSize`) by default. If want a different buffer size, can use `transactNoPrefetc` to get a stream without `prefetchN` and append any prefetch operations wanted.

### DIFF
--- a/modules/core/src/main/scala/doobie/syntax/stream.scala
+++ b/modules/core/src/main/scala/doobie/syntax/stream.scala
@@ -8,16 +8,24 @@ import doobie.util.compat.=:=
 import doobie.util.transactor.Transactor
 import doobie.free.connection.ConnectionIO
 import cats.data.Kleisli
+import cats.effect.Concurrent
 import cats.effect.kernel.{Async, MonadCancelThrow}
 import fs2.{Pipe, Stream}
 
 class StreamOps[F[_], A](fa: Stream[F, A]) {
-  def transact[M[_]: MonadCancelThrow](xa: Transactor[M])(implicit
+  def transactNoPrefetch[M[_]: MonadCancelThrow](xa: Transactor[M])(implicit
       ev: Stream[F, A] =:= Stream[ConnectionIO, A]
   ): Stream[M, A] = xa.transP.apply(fa)
+
+  def transact[M[_]: Concurrent](xa: Transactor[M])(implicit
+      ev: Stream[F, A] =:= Stream[ConnectionIO, A]
+  ): Stream[M, A] = transactNoPrefetch(xa).prefetchN(1)
+
 }
 class KleisliStreamOps[A, B](fa: Stream[Kleisli[ConnectionIO, A, *], B]) {
-  def transact[M[_]: MonadCancelThrow](xa: Transactor[M]): Stream[Kleisli[M, A, *], B] = xa.transPK[A].apply(fa)
+  def transactNoPrefetch[M[_]: MonadCancelThrow](xa: Transactor[M]): Stream[Kleisli[M, A, *], B] =
+    xa.transPK[A].apply(fa)
+  def transact[M[_]: Concurrent](xa: Transactor[M]): Stream[Kleisli[M, A, *], B] = transactNoPrefetch(xa).prefetchN(1)
 }
 class PipeOps[F[_], A, B](inner: Pipe[F, A, B]) {
   def transact[M[_]: Async](xa: Transactor[M])(implicit ev: Pipe[F, A, B] =:= Pipe[ConnectionIO, A, B]): Pipe[M, A, B] =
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/doobie/util/transactor.scala
```diff
@@ -186,6 +188,19 @@ object transactor {
...
+        def apply[T](s: Stream[ConnectionIO, T]) = {
+          fs2.Stream.eval(Queue.bounded[M, Option[T]](bufferSize)).flatMap { buffer =>
+            val res = Stream.resource(connect(kernel)).flatMap { c =>
+              Stream.resource(strategy.resource).flatMap(_ => s).translate(run(c))
+                .evalMap(x => buffer.offer(Some(x))) ++ Stream.emit(None).evalMap(buffer.offer)
+            }
+            Stream.fromQueueNoneTerminated(buffer).concurrently(res)
+          }
+        }
+      }
```
jatcwang, line 202: Perhaps we don't need to add a new method here, since users will most likely be calling `someStream.transact`

wb14123: Yeah agree. Where do you think is the best place to put the `bufferSize` config?

modules/hikari/src/test/scala/doobie/postgres/PGConcurrentSuite.scala
```diff
@@ -62,4 +62,34 @@ class PGConcurrentSuite extends munit.FunSuite {
...
+    val xa = Transactor.fromDataSource[IO](
+      dataSource,
+      ExecutionContext.fromExecutor(Executors.newFixedThreadPool(32))
+    )
+
+    val count = 100
+    val insert = for {
+      _ <- Stream.eval(sql"CREATE TABLE if not exists stream_cancel_test (i text)".update.run.transact(xa))
+      _ <- Stream.eval(sql"truncate table stream_cancel_test".update.run.transact(xa))
+      _ <- Stream.eval(sql"INSERT INTO stream_cancel_test values ('1')".update.run.transact(xa)).repeatN(count)
```
jatcwang, line 75:
```suggestion
      _ <- Stream.eval(sql"INSERT INTO stream_cancel_test select 1 from generate_series(1, $count)".update.run.transact(xa))
```

modules/hikari/src/test/scala/doobie/postgres/PGConcurrentSuite.scala
```diff
@@ -62,4 +62,34 @@ class PGConcurrentSuite extends munit.FunSuite {
...
+
+    val streamLargerBuffer = fr"select * from stream_cancel_test".query[Int].stream.transactBuffer(xa, 1024)
+      .evalMap(_ => fr"select 1".query[Int].unique.transact(xa))
+      .compile.count
+
+    assertEquals(streamLargerBuffer.unsafeRunSync(), count.toLong)
+
+    // if buffer is less than result set, it will be still block new connection since the result set is not drained
+    // use sleep to test the result set can be drained
+    val streamSmallerBuffer = fr"select * from stream_cancel_test".query[Int].stream.transactBuffer(xa, 50)
```
jatcwang, line 88: Hm sorry I can't quite understand how this is testing that the connection is closed before the stream is completed :thinking: 

An alternative way I can think of to test this is to use some sort of atomic boolean and set it to true when the transactor finishes the transaction, and then you can assert that this was set to true before you process the last chunk of the stream.
```
  import doobie.util.transactor.Strategy
  import doobie.FC
  import java.util.concurrent.atomic.AtomicBoolean
  val hasClosed = new AtomicBoolean(false)
  xa.copy(
    strategy0 = Strategy.default.copy(
      always = Strategy.default.always.flatMap(_ => FC.delay(hasClosed.set(true)))
    )
  )
```

modules/core/src/main/scala/doobie/util/transactor.scala
jatcwang: Maybe something like this in `StreamOps`?

```
  def transactPrefetchN[M[_]: MonadCancelThrow: Concurrent](xa: Transactor[M], numChunks: Int)(implicit
      ev: Stream[F, A] =:= Stream[ConnectionIO, A],
  ): Stream[M, A] = xa.transP.prefetchN(numChunks)
```

modules/hikari/src/test/scala/doobie/postgres/PGConcurrentSuite.scala
wb14123: thanks for suggestion! fixed

modules/hikari/src/test/scala/doobie/postgres/PGConcurrentSuite.scala
wb14123: Fixed the test to assert if it's closed

wb14123:
Hi @jatcwang just follow up on this PR. Please let me know if the updated code addressed your comments or not.

jatcwang (commented):
Thanks for your patience :)

Issue #809, opened 2018-12-11
Unclear error message when passing null values to updateMany


LeonardMeyer:
I was trying to do an `updateMany()` and have a nullable FK in my table. So I just put `null` inside my tuple... And here we are I guess ?
```
java.lang.RuntimeException: oops, null
	at scala.sys.package$.error(package.scala:27)
	at doobie.util.Put.unsafeSetNonNullable(put.scala:36)
	at doobie.util.Write$$anonfun$fromPut$2.apply(write.scala:64)
	at doobie.util.Write$$anonfun$fromPut$2.apply(write.scala:64)
	at doobie.util.LowerPriorityWrite$$anonfun$product$5.apply(write.scala:99)
	at doobie.util.LowerPriorityWrite$$anonfun$product$5.apply(write.scala:99)
	at doobie.util.LowerPriorityWrite$$anonfun$product$5.apply(write.scala:99)
	at doobie.util.LowerPriorityWrite$$anonfun$product$5.apply(write.scala:99)
	at doobie.util.LowerPriorityWrite$$anonfun$product$5.apply(write.scala:99)
	at doobie.util.LowerPriorityWrite$$anonfun$product$5.apply(write.scala:99)
	at doobie.util.LowerPriorityWrite$$anonfun$product$5.apply(write.scala:99)
	at doobie.util.LowerPriorityWrite$$anonfun$product$5.apply(write.scala:99)
	at doobie.util.LowerPriorityWrite$$anonfun$product$5.apply(write.scala:99)
	at doobie.util.LowerPriorityWrite$$anonfun$product$5.apply(write.scala:99)
	at doobie.util.LowerPriorityWrite$$anonfun$generic$2.apply(write.scala:110)
	at doobie.util.LowerPriorityWrite$$anonfun$generic$2.apply(write.scala:110)
	at doobie.util.Write$$anonfun$set$1.apply(write.scala:24)
	at doobie.util.Write$$anonfun$set$1.apply(write.scala:24)
	at doobie.free.KleisliInterpreter$$anonfun$primitive$1$$anonfun$apply$1.apply(kleisliinterpreter.scala:114)
	at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:87)
	at cats.effect.internals.IORunLoop$.startCancelable(IORunLoop.scala:41)
	at cats.effect.internals.IOBracket$BracketStart.run(IOBracket.scala:86)
[...]

PR #2141, opened 2024-11-19

ulfryk:


modules/core/src/main/scala/doobie/util/put.scala
```diff
@@ -56,25 +56,37 @@ sealed abstract class Put[A](
...
   def unsafeSetNullable(ps: PreparedStatement, n: Int, oa: Option[A]): Unit =
     oa match {
-      case Some(a) => unsafeSetNonNullable(ps, n, a)
+      case Some(a) => unsafeSetWhatShouldNotBeNull(ps, n, a, expectedOptionalParamAt(n))
       case None    => unsafeSetNull(ps, n)
     }
 
+  private def unsafeSetWhatShouldNotBeNull(ps: PreparedStatement, n: Int, a: A, message: String): Unit =
+    if (a == null) sys.error(message)
+    else put.fi.apply(ps, n, (put.k(a)))
```
jatcwang, line 69: In this case we're creating a new string for every value we set, which is probably not ideal for performance.
We probably want to only construct the message in the `if (a == null)` branch. To simplify things, I think it's probably not necessary to differentiate between `null` and `Option(null)` (though I appreciate the thoughtfulness behind it :slightly_smiling_face: )

ulfryk: @jatcwang I have updated the code. It was the `n` param info that caused creating new string for every value we set. Now both strings are persisted and we keep the `null` vs `Option(null)` distinction :)

ulfryk: Or do you think super simple would be even better? Like this --> https://github.com/typelevel/doobie/pull/2142

ulfryk: Or with message constructed only in the `if (a == null)` branch and full information --> #2143

PR #2142, opened 2024-11-20

ulfryk:
Same as #2141 but in the simplest possible manner.

### DIFF
--- a/modules/core/src/main/scala/doobie/util/put.scala
+++ b/modules/core/src/main/scala/doobie/util/put.scala
@@ -56,7 +56,7 @@ sealed abstract class Put[A](
     ) {}
 
   def unsafeSetNonNullable(ps: PreparedStatement, n: Int, a: A): Unit =
-    if (a == null) sys.error("oops, null")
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/doobie/util/put.scala
```diff
@@ -56,7 +56,7 @@ sealed abstract class Put[A](
     ) {}
 
   def unsafeSetNonNullable(ps: PreparedStatement, n: Int, a: A): Unit =
-    if (a == null) sys.error("oops, null")
+    if (a == null) sys.error("Expected non-nullable param. Use Option to describe nullable values.")
```
jatcwang, line 59: Thanks nice and simple :) 
We can put `n` into the error message to provide an even friendlier error message. Since this is all under the `if` clause we won't pay the cost of constructing the string until the null case is encountered.

ulfryk: done :)

jatcwang:
Thanks a lot! :smile:

PR #2143, opened 2024-11-20

ulfryk:
Same as #2141 but with more information.

The error message string is created only in `if (a == null)` branch ( to solve issue spotted here https://github.com/typelevel/doobie/pull/2141#discussion_r1848922349 )

### DIFF
--- a/modules/core/src/main/scala/doobie/util/put.scala
+++ b/modules/core/src/main/scala/doobie/util/put.scala
@@ -56,24 +56,36 @@ sealed abstract class Put[A](
     ) {}
 
   def unsafeSetNonNullable(ps: PreparedStatement, n: Int, a: A): Unit =
-    if (a == null) sys.error(s"Expected non-nullable param at $n. Use Option to describe nullable values.")
-    else put.fi.apply(ps, n, (put.k(a)))
+    unsafeSetWhatShouldNotBeNull(ps, n, a, expectedNonNullableParam)
 
   def unsafeSetNullable(ps: PreparedStatement, n: Int, oa: Option[A]): Unit =
     oa match {
-      case Some(a) => unsafeSetNonNullable(ps, n, a)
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/doobie/util/put.scala
```diff
@@ -56,25 +56,37 @@ sealed abstract class Put[A](
...
+    unsafeSetWhatShouldNotBeNull(ps, n, a, expectedNonNullableParam)
 
   def unsafeSetNullable(ps: PreparedStatement, n: Int, oa: Option[A]): Unit =
     oa match {
-      case Some(a) => unsafeSetNonNullable(ps, n, a)
+      case Some(a) => unsafeSetWhatShouldNotBeNull(ps, n, a, expectedOptionalParam)
       case None    => unsafeSetNull(ps, n)
     }
 
+  private def unsafeSetWhatShouldNotBeNull(ps: PreparedStatement, n: Int, a: A, message: Int => String): Unit =
```
jatcwang, line 67: I believe this will still allocate the anonymous function (unless JIT is able to elide the allocation?)

ulfryk: That's actually a good point, but is it so if it's an axisting method passed as an argument? How is it compiled to bytecode?

jatcwang: https://godbolt.org/z/5sz68x4cE

When I use an anonymous function it seems to extract it into a static method, so there are probably no allocation either way in this case. 
I know next to nothing about JVM bytecode though so :smile:

ulfryk: :D

So it seems (but not without a doubt) that it is not affecting performance too much.

Your call then @jatcwang - should we go with it for sake of Developer Experience or is the simpler (already merged) version enough 🙌

jatcwang: Will keep it simple for now as I feel the merged version already have enough information for people to find the source of null :)

PR #2154, opened 2024-11-28
Feature: Allow customizing individual steps of query execution

ulfryk:
@jatcwang a promised PoC so we can make some decisions early on :)


modules/core/src/main/scala/doobie/hi/connection.scala
```diff
@@ -543,4 +577,24 @@ object connection {
...
+      prep: PreparedStatementIO[Unit],
+      exec: PreparedStatementIO[A]
+  )
+
+  final case class PreparedExecutionStream(
+      create: ConnectionIO[PreparedStatement],
+      prep: PreparedStatementIO[Unit],
+      exec: PreparedStatementIO[ResultSet],
+      chunkSize: Int
+  )
```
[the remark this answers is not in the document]
jatcwang: If we remove `chunkSize` then this will be the same as `PreparedExecutionWithoutResultSet[ResultSet]`. That does look weird though, so I think we can rename `PreparedExecutionWithoutResultSet` to something like `PreparedExecutionWithoutProcessStep`? :smile:

modules/core/src/main/scala/doobie/hi/connection.scala
```diff
@@ -226,6 +248,18 @@ object connection {
     } yield ele
   }
 
+  def stream[A: Read](
+      prepared: PreparedExecutionStream,
+      loggingInfo: LoggingInfo
```
jatcwang, line 253: I'm not sure these methods are really needed? Especially since the current API does not really allow users to work with (execute) a PreparedExecutionStream directly

modules/core/src/main/scala/doobie/util/query.scala
```diff
@@ -252,6 +264,8 @@ object query {
 
   }
 
+  type PreparedExecutionUpdate[A] = PreparedExecutionWithResultSet[A] => PreparedExecutionWithResultSet[A]
```
jatcwang, line 267: IMO I don't feel we don't need a type alias for this? My don't use type aliases too often as I find it makes APIs harder to understand as it requires additional jump-to-definition to understand what a function is doing.

jatcwang:
I agree with the overall direction you're going with this. I think it will provide most if not all of the customizations users will need :+1:

ulfryk:
@jatcwang

Methods added to Query class:
- `toAlteringExecution` (as `to`)
- `toMapAlteringExecution` (as `toMap`)
- `accumulateAlteringExecution` (as `accumulate`)
- `uniqueAlteringExecution` (as `unique`)
- `optionAlteringExecution` (as `option`)
- `nelAlteringExecution` (as `nel`)

Methods added to Update class:
- `runAlteringExecution` (as `run`)
- `updateManyAlteringExecution` (as `updateMany`)
- `withUniqueGeneratedKeysAlteringExecution` (as `withUniqueGeneratedKeys`)

So basically all operations returning `ConnectionIO`.
As suggested in https://github.com/typelevel/doobie/pull/2154#discussion_r1864641661 I have skipped ones yielding Stream.

---

Now some questions:
- should we use this name pattern - `…AlteringExecution` suffix?
- should we add similar methods to the `Query0` and `Update0` traits?

jatcwang:
Thanks this looks good! Sorry for the late reply. 

I think `AlteringExection` is good! One other possibility I can think of is `Customized` e.g. `toMapCustomized`.

Yes we'll need to add these to Query0 and Update0 too

PR #2166, opened 2025-01-14
Propose a fragment concat operator that ensures a whitespace in between

satorg:
This PR proposes a new operator `+~+` for fragments concatenation that makes sure that there is at least 1 whitespace between the fragments being concatenated. I realize, this proposal is going to be a controversial one. However I decided to give it a shot.

### Motivation

I've been working with Doobie for quite a while and use its fragment API to create pretty sophisticated queries. Some of those queries are created dynamically, i.e. various parts of the SQL are concatenated depending on some external conditions. With time I started getting a feeling that the whole `fr`/`fr0`/`const`/`const0` paradigm might not be the great idea at all. When I create a fragment that may or may not be concatenated with something else later, I don't really want to worry about whether I have to end it with a whitespace or not. I _do_ cary though about the whitespace at the moment when the concatenation happens, if it is the case. So this is where the idea of the proposed operator comes from: instead of keeping track of previous fragments endings, we can simply use `fr0` and `const0` _everywhere_ along with either `++` or `+~+` operators, depending on a particular case. For example:
```scala
// Consider it returns something like `SELECT ... FROM foo JOIN bar ON whatever`
def baseSelectSQL: Fragment = ???

def conditionalSelectSQL(one: Option[Int], two: Option[Int]) =
  baseSelectSQL +~+
    andOpt(
      one.map(a => fr0"one = $one"),
      two.map(b => fr0"two = $b)
    ).fold(Fragment.empty)(fr0"WHERE" +~+ _)
```
In other words, with `+~+` operator we should never worry about adding explicit whitespaces anywhere anymore – the operator will take care of it automatically. Thereby the code becomes clearer and SQL logs (if employed) become less cluttered.

### About the Naming

My initial intent was to give the operator this name: `+_+`. Unfortunately, Scala parser has its own business with underscores in operator-alike names and such a name cannot be used without backticks. So I ended up with something that is relatively simple and self-explanatory.

However, if there is a chance that the PR can be accepted, I'm generally open to any other (better) name, so feel free to suggest other options please.

### DIFF
--- a/modules/core/src/main/scala/doobie/util/fragment.scala
+++ b/modules/core/src/main/scala/doobie/util/fragment.scala
@@ -57,6 +57,23 @@ object fragment {
     def ++(fb: Fragment): Fragment =
       new Fragment(sql + fb.sql, elems ++ fb.elems, pos orElse fb.pos)
 
+    /** Concatenate this fragment with another, yielding a larger fragment and making sure there is at least one
+      * whitespace between the source fragments.
+      */
+    def +~+(that: Fragment): Fragment = {
+      val res =
+        if (sql.isEmpty) that.sql
+        else if (that.sql.isEmpty) sql
+        else if (sql.last.isWhitespace || that.sql.head.isWhitespace) sql + that.sql
+        else sql + " " + that.sql
+
+      new Fragment(
+        res,
+        elems ++ that.elems,
+        pos orElse that.pos
+      )
+    }
+
     def stripMargin(marginChar: Char): Fragment =
       new Fragment(sql.stripMargin(marginChar), elems, pos)
 
--- a/modules/core/src/test/scala/doobie/util/FragmentSuite.scala
+++ b/modules/core/src/test/scala/doobie/util/FragmentSuite.scala
@@ -4,9 +4,10 @@
 
 package doobie.util
 
-import cats.syntax.all.*
 import cats.effect.IO
+import cats.syntax.all.*
 import doobie.*
+import doobie.Fragment.const0
 import doobie.implicits.*
 import doobie.testutils.VoidExtensions
 import munit.CatsEffectSuite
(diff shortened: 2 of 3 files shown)


jatcwang:
Thanks for the PR! I agree that this is very useful.

modules/core/src/main/scala/doobie/util/fragment.scala
```diff
@@ -57,6 +57,23 @@ object fragment {
...
       new Fragment(sql + fb.sql, elems ++ fb.elems, pos orElse fb.pos)
 
+    /** Concatenate this fragment with another, yielding a larger fragment and making sure there is at least one
+      * whitespace between the source fragments.
+      */
+    def +~+(that: Fragment): Fragment = {
+      val res =
+        if (sql.isEmpty) that.sql
+        else if (that.sql.isEmpty) sql
+        else if (sql.last.isWhitespace || that.sql.head.isWhitespace) sql + that.sql
```
jatcwang, line 67: I'm wondering whether we want to perform this check. Given the context it's used, I don't think extra spaces should matter?

(While working on fragments.scala, I did consider doing similar checks to generate nicer looking SQL but ultimately decided against it)

satorg: Sorry, I'm not sure I fully understand your question, do you mean to remove all the "if-s" and leave only the `sql + " " + that.sql` unconditionally?

My initial intent was that the checks are the main point of the new operator – it tries not to interfere with the spaces which a user might have already added and steps in only if they didn't, but at least one whitespace is still required. For example:
```scala
const0("SELECT a, b, c\nFROM the_table\n") +~+
  (if (we_need_cond) fr0("WHERE a = $a") else (Fragment.empty))
```
I.e. it should keep a new line before `WHERE` without more whitespaces added. So basically yes, it allows to format SQL nicely, when necessary. I personally do it for the sake of more readable logs, in particular.

But I realize it is not exactly a big deal, so let me know if it is actually a concern please.

jatcwang: Yeah that's what I meant.

I was trying to avoid the constant check on the last character of the string. Upon further digging I think it's negligible though. If you're often looking at logged DB queries (which I'm not) then the additional readability is beneficial, I agree :+1:

jatcwang: Would you be able to take a look at `src/main/mdoc/docs/08-Fragments.md` and see where the docs can be updated to reflect this improvement? :)

satorg: @jatcwang , I've added some examples with `+~+` to the docs, as you suggested. Let me know if it works. Thanks!

PR #2195, opened 2025-02-11
Allow constructing Read/Write from a list of underlying instances

jatcwang:


modules/core/src/main/scala/doobie/util/read.scala
```diff
@@ -145,6 +146,30 @@ object Read extends LowerPriority1Read {
     override lazy val length: Int = read0.length + read1.length
   }
 
+  /** A Composite made up of a list of underlying Read instances. This class but is intended to provide a simpler
+    * interface for other libraries that uses Doobie. This isn't used by Doobie itself for its derived instances.
+    *
+    * For large number of columns, this class may be more performant than chain of Read.Composite.
+    */
+  class CompositeOfInstances[A: ClassTag](readInstances: Array[Read[A]]) extends Read[Array[A]] {
```
jatcwang, line 154: @milessabin Switched back to using Array instead of ArraySeq to avoid dependency on scala-collection-compat

PR #2212, opened 2025-03-10
Migrated NotifySuite tests

PraverBajaj:
I have successfully migrated all the tests, and they are now failing for incorrect values as well. This confirms that the migration is working as intended and performing correctly, which was the primary objective.

### DIFF
--- a/modules/postgres/src/test/scala/doobie/postgres/NotifySuite.scala
+++ b/modules/postgres/src/test/scala/doobie/postgres/NotifySuite.scala
@@ -15,24 +15,24 @@ class NotifySuite extends munit.CatsEffectSuite {
   import PostgresTestTransactor.xa
 
   // Listen on the given channel, notify on another connection
-def listen[A](channel: String, notify: ConnectionIO[A]): IO[List[PGNotification]] = 
-  WeakAsync.liftIO[ConnectionIO].use { liftIO =>
-    (for {
-      _ <- PHC.pgListen(channel)
-      _ <- commit
-      _ <- delay { Thread.sleep(50) }
-      _ <- liftIO.liftIO(notify.transact(xa))
-      _ <- commit
-      _ <- delay { Thread.sleep(50) }
-      notifications <- PHC.pgGetNotifications
-    } yield notifications).transact(xa)
-  }
+  def listen[A](channel: String, notify: ConnectionIO[A]): IO[List[PGNotification]] =
+    WeakAsync.liftIO[ConnectionIO].use { liftIO =>
+      (for {
+        _ <- PHC.pgListen(channel)
+        _ <- commit
+        _ <- delay { Thread.sleep(50) }
+        _ <- liftIO.liftIO(notify.transact(xa))
+        _ <- commit
+        _ <- delay { Thread.sleep(50) }
+        notifications <- PHC.pgGetNotifications
+      } yield notifications).transact(xa)
+    }
 
   test("LISTEN/NOTIFY should allow cross-connection notification") {
     val channel = "cha" + System.nanoTime.toString
     val notify = PHC.pgNotify(channel)
     val test = listen(channel, notify).map(_.length)
-    test.assertEquals(1)  
+    test.assertEquals(1)
   }
 
   test("LISTEN/NOTIFY should allow cross-connection notification with parameter") {
@@ -40,7 +40,7 @@ def listen[A](channel: String, notify: ConnectionIO[A]): IO[List[PGNotification]
     val messages = List("foo", "bar", "baz", "qux")
     val notify = messages.traverse(PHC.pgNotify(channel, _))
     val test = listen(channel, notify).map(_.map(_.getParameter))
-    test.assertEquals(messages) 
+    test.assertEquals(messages)
   }
 
   test("LISTEN/NOTIFY should collapse identical notifications") {
--- a/modules/postgres/src/test/scala/doobie/postgres/NotifySuite.scala
+++ b/modules/postgres/src/test/scala/doobie/postgres/NotifySuite.scala
@@ -29,20 +29,23 @@ class NotifySuite extends munit.CatsEffectSuite {
   test("LISTEN/NOTIFY should allow cross-connection notification") {
     val channel = "cha" + System.nanoTime.toString
     val notify = PHC.pgNotify(channel)
-    listen(channel, notify).map(_.length).assertEquals(1)  
+    val test = listen(channel, notify).map(_.length)
+    test.assertEquals(1)  
   }
 
   test("LISTEN/NOTIFY should allow cross-connection notification with parameter") {
     val channel = "chb" + System.nanoTime.toString
     val messages = List("foo", "bar", "baz", "qux")
     val notify = messages.traverse(PHC.pgNotify(channel, _))
-    listen(channel, notify).map(_.map(_.getParameter)).assertEquals(messages) 
+    val test = listen(channel, notify).map(_.map(_.getParameter))
+    test.assertEquals(messages) 
   }
 
   test("LISTEN/NOTIFY should collapse identical notifications") {
     val channel = "chc" + System.nanoTime.toString
     val messages = List("foo", "bar", "bar", "baz", "qux", "foo")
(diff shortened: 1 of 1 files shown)


modules/postgres/src/test/scala/doobie/postgres/NotifySuite.scala
```diff
@@ -4,46 +4,45 @@
...
+    _ <- delay { Thread.sleep(50) }
+    notifications <- PHC.pgGetNotifications
+  } yield notifications).transact(xa)
 
   test("LISTEN/NOTIFY should allow cross-connection notification") {
     val channel = "cha" + System.nanoTime.toString
     val notify = PHC.pgNotify(channel)
-    val test = listen(channel, notify).map(_.length)
-    assertEquals(test.unsafeRunSync(), 1)
+    listen(channel, notify).map(_.length).assertEquals(1)
```
armanbilge, line 32: Thanks for the PR! To keep the changes minimal and easy to review, do you think we can keep the `val test =` line unchanged?

modules/postgres/src/test/scala/doobie/postgres/NotifySuite.scala
```diff
@@ -4,46 +4,45 @@
...
-      PHC.pgGetNotifications).transact(xa)
+  (for {
+    _ <- PHC.pgListen(channel)
+    _ <- commit
+    _ <- delay { Thread.sleep(50) }
+    _ <- notify
+    _ <- commit
+    _ <- delay { Thread.sleep(50) }
+    notifications <- PHC.pgGetNotifications
+  } yield notifications).transact(xa)
```
armanbilge, line 27: I don't think that this rewrite is equivalent, we lost the `transact(xa)` on `notify`.

modules/postgres/src/test/scala/doobie/postgres/NotifySuite.scala
PraverBajaj: Sure, I can revert that change to keep the diff minimal. My intention was to leverage CatsEffectSuite's built-in support for IO assertions, making the code more idiomatic. However, I understand the importance of keeping changes easy to review, so I'll restore the val test = line as it was. Let me update the PR accordingly.

modules/postgres/src/test/scala/doobie/postgres/NotifySuite.scala
PraverBajaj: You're right—we lost transact(xa) on notify, which changes the semantics. My goal was to remove unsafeRunSync(), so I'll update the code to explicitly use .transact(xa) on notify while keeping the refactored structure. Thanks for catching that!

modules/postgres/src/test/scala/doobie/postgres/NotifySuite.scala
PraverBajaj: PR Updated

modules/postgres/src/test/scala/doobie/postgres/NotifySuite.scala
PraverBajaj: 1. The problem is that notify.transact(xa) transforms notify, which is of type ConnectionIO[A], into IO[A], but listen expects a ConnectionIO[A].
2. notify.transact(xa).unsafeRunSync() was forcing execution in IO, but we need to keep everything in ConnectionIO.
3. I placed notify inside the for comprehension to ensure it executes within the ConnectionIO transaction.
4. .transact(xa) is called only once at the end, keeping the entire sequence in ConnectionIO before converting it to IO.
5. If I call notify.transact(xa) inside the for comprehension, it converts it to IO[A], breaking the transaction sequencing within ConnectionIO.
6. Calling .transact(xa) only once at the end ensures that the entire process (LISTEN → NOTIFY → GET NOTIFICATIONS) runs within a single database transaction.

This is what I thought while executing and this is the learning I got from the errors I got while doing the same.

I might be wrong but according to me it could be a problem :-

- If notify includes operations that must be committed separately before pgGetNotifications executes.

- If notify depends on external IO side effects outside ConnectionIO.

PraverBajaj: However, an alternative approach is to use the fiber method, where we call notify.transact(xa).start. This runs notify asynchronously in the background while the main transaction proceeds. By doing this, we avoid blocking, ensure the notification executes correctly, and maintain transaction integrity. Finally, we use fiber.joinWithNever to synchronize and clean up, ensuring notify completes properly without interfering with the main logic.As shown in the picture .

It runs notify in the background without disrupting the main transaction

If there is any other approach, please guide me with that. I will be very grateful to explore alternative solutions.

armanbilge: > 2\. notify.transact(xa).unsafeRunSync() was forcing execution in IO, but we need to keep everything in ConnectionIO.

Take a look at `LiftIO` to convert `IO` back into `ConnectionIO`.

https://github.com/typelevel/doobie/blob/2bdd9a80d68ee2e5f8608eada83abc03abf33888/modules/free/src/main/scala/doobie/WeakAsync.scala#L55

PraverBajaj: Okay, that’s interesting. I implemented this, but the test failed earlier. Let me take a deeper look into it.
Maybe my implementation was incorrect. I'll investigate further.

modules/postgres/src/test/scala/doobie/postgres/NotifySuite.scala
```diff
@@ -4,46 +4,50 @@
...
-    (PHC.pgListen(channel) *> commit *>
-      delay { Thread.sleep(50) } *>
-      Sync[ConnectionIO].delay(notify.transact(xa).unsafeRunSync()) *>
-      delay { Thread.sleep(50) } *>
-      PHC.pgGetNotifications).transact(xa)
+    WeakAsync.liftIO[ConnectionIO].use { liftIO =>
+      (for {
+        _ <- PHC.pgListen(channel)
+        _ <- commit
+        _ <- delay { Thread.sleep(50) }
```
armanbilge, line 23: I wonder if we can also rewrite these:
```suggestion
        _ <- liftIO(IO.sleep(50.millis))
```

PraverBajaj: Idk, but when I am rewriting them as specified, it is giving me errors.

PraverBajaj: In Cats Effect 3, LiftIO has been replaced with Sync.
found this on the web

PraverBajaj: It is not possible to do that.

armanbilge: > In Cats Effect 3, LiftIO has been replaced with Sync.
> found this on the web

That's not correct ...

PraverBajaj: > What are the errors?

PraverBajaj: > What are the errors?

Screenshot 2025-03-11 at 4 56 23 AM

armanbilge: Sorry, I wrote it too hastily ... it should be `liftIO.liftIO(...)`, like the other ones.

PraverBajaj: "Yes, that's correct! I didn't notice it either."

PraverBajaj: It worked ,
Updating the PR.

PraverBajaj: Got it . I used 
import scala.concurrent.duration._
it should be 
import scala.concurrent.duration.*

jatcwang (approved):

jatcwang:
Thanks!

Issue #2147, opened 2024-11-22
Convert to use munit-cats-effect


jatcwang:
https://typelevel.org/munit-cats-effect/

We currently write a lot of unsafeRunSync in our test code which can be avoided with the above library

mzuehlke:
Has this been done in scope of #2181 ?

jatcwang:
@mzuehlke Only a few files have been converted so this isn't done yet

![Image](https://github.com/user-attachments/assets/73f59d4d-686a-40c0-b17f-7bac146d145d)

mzuehlke:
Thanks

PR #2213, opened 2025-03-12
 Fix Scala 3 migration issues in Hikari module

Sanskriti0805:
- Fixed incorrect usage of `liftF` in transaction handling
- Replaced deprecated `join` with `joinWithNever`
- Added missing import for `ExecutionContexts`
- Renamed class from `824` to `Issue824` for better readability


jatcwang:
Most of the files in doobie are cross compiled between Scala 2 and 3, so we shouldn't / cannot make changes to the file that make it only compile in one Scala version.

This also doesn't do anything to resolve #2147 which is what I think you want to work on.

All in all this seems like a bunch of code change made by AI (apologies if I'm mistaken). I would recommend not using it and try to really understand the changes you are making. This is is critical for your own learning and career.

Sanskriti0805:
"Thanks for the feedback, @jatcwang! I misunderstood the issue earlier and made changes that were incompatible with Scala 2. i will look deeper into it again

PR #2215, opened 2025-03-13
 converted to munits cat effect

Sanskriti0805:
Fixed the assertion method in the test acc to munits cat effect library


modules/hikari/src/test/scala/doobie/hikari/issue/824.scala
```diff
@@ -10,16 +10,16 @@ import com.zaxxer.hikari.HikariDataSource
...
 
-class `824` extends munit.FunSuite {
-
-  import cats.effect.unsafe.implicits.global
+class `824` extends CatsEffectSuite {
 
   val transactor: Resource[IO, HikariTransactor[IO]] =
     for {
-      ce <- ExecutionContexts.fixedThreadPool[IO](16) // our connect EC
+      ce <- ExecutionContexts.fixedThreadPool[IO](16)  // Fiper implicit resolution
```
satorg, line 22: "Fiper" – did you mean "Fiber"?

modules/hikari/src/test/scala/doobie/hikari/issue/824.scala
```diff
@@ -59,12 +59,11 @@ class `824` extends munit.FunSuite {
 
     } flatMap { case (n, ds) =>
       // One final report to show that all connections are disposed
-      report(ds) *> IO((n, ds.getHikariPoolMXBean.getTotalConnections))
-
+      report(ds) *> IO.pure(n -> ds.getHikariPoolMXBean.getTotalConnections)
```
satorg, line 62: I wouldn't recommend this change – since `ds.getHikariPoolMXBean.getTotalConnections` is impure operation, it is correct to keep it inside `IO`.

modules/hikari/src/test/scala/doobie/hikari/issue/824.scala
```diff
@@ -59,12 +59,11 @@ class `824` extends munit.FunSuite {
...
     }
 
   test("HikariTransactor should close connections logically within `use` block and physically afterward.") {
-    assertEquals(prog.unsafeRunSync(), (0, 0))
+    prog.map(result => assertEquals(result, (0, 0))) // Fixed: Replaced assertIOEquals with assertEquals inside map
```
satorg, line 66: This one will do better:
```suggestion
    prog.assertEquals(result, (0, 0))
```
It is not only more concise and readable,
but also it is generally not recommended to run impure operations inside `map`.

modules/hikari/src/test/scala/doobie/hikari/issue/824.scala
Sanskriti0805: yes yess

Sanskriti0805:
@satorg i have reviewed the changes and force pushed them , can you take a look ?

satorg:
Thank you @Sanskriti0805 !

A couple of comments:
1. What do you mean by "Fiber implicit resolution" exactly? I'm not sure I understand this comment, could you elaborate please.
2. The project code base requires proper formatting. Otherwise it won't build on CI. The command `sbt +scalafmtAll` usually does the trick.

modules/hikari/src/test/scala/doobie/hikari/issue/824.scala
Sanskriti0805: i did this but then its an error as result is not defined anywhere in the code , prog is an IO[(Int, Int)], so we need to extract the result from it.

Sanskriti0805:
> Thank you @Sanskriti0805 !
> 
> A couple of comments:
> 
> 1. What do you mean by "Fiber implicit resolution" exactly? I'm not sure I understand this comment, could you elaborate please.
> 2. The project code base requires proper formatting. Otherwise it won't build on CI. The command `sbt +scalafmtAll` usually does the trick.
  by fiber implicit resolution i actually meant that Cats Effect automatically chooses the right way to run small and lightweight tasks (called fibers) in the background.

modules/hikari/src/test/scala/doobie/hikari/issue/824.scala
```diff
@@ -60,11 +60,11 @@ class `824` extends munit.FunSuite {
     } flatMap { case (n, ds) =>
       // One final report to show that all connections are disposed
       report(ds) *> IO((n, ds.getHikariPoolMXBean.getTotalConnections))
-
     }
 
   test("HikariTransactor should close connections logically within `use` block and physically afterward.") {
-    assertEquals(prog.unsafeRunSync(), (0, 0))
+    prog.assertEquals(result, (0, 0))
```
satorg, line 66:
```suggestion
    prog.assertEquals((0, 0))
```

satorg: @Sanskriti0805 ,
Actually, it was not necessary to pull another request just to get a "cleaner version" of this one. If you'd like to get rid of many not-so-perfect commits, you could simply [squash](https://www.geeksforgeeks.org/use-of-git-squash-commits/) them into a single shiny final one. Moreover, creating multiple PRs while working on the same task pollutes PR history in the repository quite a bit and rather creates confusion than helps to maintain clearness.

satorg:
@Sanskriti0805 , what is the reason of attaching `sbt-launcher.jar` to the PR?

modules/hikari/src/test/scala/doobie/hikari/issue/824.scala
Sanskriti0805: @satorg "Got it! Thanks for the guidance. I'll make sure to squash commits next time instead of opening a new PR to keep the history clean. Appreciate the feedback!"

PR #2224, opened 2025-03-20
Migrated test suites to MunitCatsEffect

### LINKED ISSUE #2147: Convert to use munit-cats-effect

sharmaakshay177:
Migrated tests mentioned in comments of https://github.com/typelevel/doobie/issues/2147 to use munit-cats-effect.

Closes #2147


modules/postgres/src/test/scala/doobie/postgres/TextSuite.scala
```diff
@@ -12,10 +12,9 @@ import doobie.postgres.implicits.*
 import fs2.*
 import org.scalacheck.{Gen, Test}
 import org.scalacheck.Arbitrary.arbitrary
-import org.scalacheck.Prop.forAll
+import org.scalacheck.effect.PropF.forAllF
 
-class TextSuite extends munit.ScalaCheckSuite {
-  import cats.effect.unsafe.implicits.global
+class TextSuite extends munit.CatsEffectSuite with munit.ScalaCheckSuite {
```
satorg, line 17: Oh, here is a catch: if you simply combine `munit.CatsEffectSuite` with `munit.ScalaCheckSuite`, then the tests will compile successfully but won't work and will always succeed:
```scala
class NonWorkingSuite extends munit.CatsEffectSuite with munit.ScalaCheckSuite {
  test("always succeeds no matter what") {
    PropF.forAllF { (a: Boolean) =>
      IO(a).assertEquals(!a)
    }
  }
}
```
It is an unfortunate consequence of the way how MUnit treats return values from test cases.
In order to make it working correctly, you have to use `munit.ScalaCheckEffectSuite` instead:
```scala
class WorkingSuite extends munit.CatsEffectSuite with munit.ScalaCheckEffectSuite {
  test("always fails as expected") {
    PropF.forAllF { (a: Boolean) =>
      IO(a).assertEquals(!a)
    }
  }
}
```

sharmaakshay177: Thank you, @satorg for the explanation and the review. 
Made changes to use `munit.ScalaCheckEffectSuite` instead of `munit.ScalaCheckSuite`

modules/postgres/src/test/scala/doobie/postgres/ReadErrorSuite.scala
```diff
@@ -29,18 +28,17 @@ class ReadErrorSuite extends munit.FunSuite {
...
-    assertEquals(r, Left(InvalidEnum[MyScalaEnum.Value]("invalid")))
+    sql"select 'invalid' :: myenum".query[MyScalaEnum.Value].unique.transact(xa).attempt.assertEquals(Left(
+      InvalidEnum[MyScalaEnum.Value]("invalid")))
   }
 
   test("pgJavaEnum") {
-    val r = sql"select 'invalid' :: myenum".query[MyJavaEnum].unique.transact(xa).attempt.unsafeRunSync()
-    assertEquals(r, Left(InvalidEnum[MyJavaEnum]("invalid")))
+    sql"select 'invalid' :: myenum".query[MyJavaEnum].unique.transact(xa).attempt.assertEquals(Left(
+      InvalidEnum[MyJavaEnum]("invalid")))
```
satorg, line 41: Not a requirement, but I think these test expressions would be a bit more readable if executions and assertions parts were kept on separate lines like this:
```suggestion
    sql"select 'invalid' :: myenum".query[MyJavaEnum].unique.transact(xa).attempt
      .assertEquals(Left(InvalidEnum[MyJavaEnum]("invalid")))
```
This basically could be applied to all the test cases in the PR.

Although, you don't have to, just a suggestion. Feel free to ignore)

modules/refined/src/test/scala/doobie/refined/RefinedSuite.scala
```diff
@@ -54,15 +52,15 @@ class RefinedSuite extends munit.FunSuite {
   }
 
   test("Query should return a refined type when conversion is possible") {
-    sql"select 123".query[PositiveInt].unique.transact(xa).void.unsafeRunSync()
+    sql"select 123".query[PositiveInt].unique.transact(xa).void.assert
```
satorg, line 55: Not sure, why the original test doesn't contain any value checks besides that there's no failure.
I think adding this check shouldn't hurt, should it?
```suggestion
    sql"select 123".query[PositiveInt].unique.transact(xa)
      .assertEquals(PositiveInt(123))
```

modules/refined/src/test/scala/doobie/refined/RefinedSuite.scala
```diff
@@ -54,15 +52,15 @@ class RefinedSuite extends munit.FunSuite {
...
   }
 
   test("Query should return an Option of a refined type when query returns null-value") {
-    sql"select NULL".query[Option[PositiveInt]].unique.transact(xa).void.unsafeRunSync()
+    sql"select NULL".query[Option[PositiveInt]].unique.transact(xa).void.assert
```
satorg, line 59: The same concern here:
```suggestion
    sql"select NULL".query[Option[PositiveInt]].unique.transact(xa)
      .assertEquals(None)
```

modules/refined/src/test/scala/doobie/refined/RefinedSuite.scala
```diff
@@ -54,15 +52,15 @@ class RefinedSuite extends munit.FunSuite {
...
   }
 
   test("Query should return an Option of a refined type when query returns a value and converion is possible") {
-    sql"select NULL".query[Option[PositiveInt]].unique.transact(xa).void.unsafeRunSync()
+    sql"select NULL".query[Option[PositiveInt]].unique.transact(xa).void.assert
```
satorg, line 63: This test case does not correspond to the test title, I guess it should look like this:
```suggestion
    sql"select 123".query[Option[PositiveInt]].unique.transact(xa)
      .assertEquals(Some(PositiveInt(123))
```
Otherwise there's no difference with the previous one.

satorg: Also there's a typo in the title, must be "conver***s***ion", I guess.

modules/refined/src/test/scala/doobie/refined/RefinedSuite.scala
```diff
@@ -80,26 +78,26 @@ class RefinedSuite extends munit.FunSuite {
...
       _ <- sql"INSERT INTO TEST VALUES ($v)".update.run
     } yield ()
-    queryRes.transact(xa).unsafeRunSync()
+    queryRes.transact(xa).assert
   }
 
   test("Query should throw an SecondaryValidationFailed if value does not fit the refinement-type ") {
     secondaryValidationFailedCaught_?(
-      sql"select -1".query[PositiveInt].unique.transact(xa).void.unsafeRunSync()
+      sql"select -1".query[PositiveInt].unique.transact(xa).void.assert
```
satorg, line 86: `.assert` seems redundant here and below – the exception is going to be intercepted inside `secondaryValidationFailedCaught_?`.

modules/refined/src/test/scala/doobie/refined/RefinedSuite.scala
```diff
@@ -80,26 +78,26 @@ class RefinedSuite extends munit.FunSuite {
...
     secondaryValidationFailedCaught_?(
-      sql"select -1, 1".query[PointInQuadrant1].unique.transact(xa).void.unsafeRunSync()
+      sql"select -1, 1".query[PointInQuadrant1].unique.transact(xa).void.assert
     )
   }
 
-  private def secondaryValidationFailedCaught_?(query: => Unit) =
-    intercept[SecondaryValidationFailed[?]](query)
+  private def secondaryValidationFailedCaught_?(query: => IO[Unit]) =
+    query.intercept[SecondaryValidationFailed[?]]
```
satorg, line 101: By-name parameter is redundant with effects:
```suggestion
  private def secondaryValidationFailedCaught_?(query: IO[Unit]) =
    query.intercept[SecondaryValidationFailed[?]]
```
To be honest, I'm not sure this function is necessary at all. With the effect assertsions it would probably be clearer to write tests directly like this:
```scala
  test("Query should throw an SecondaryValidationFailed if object does not fit the refinement-type ") {
    sql"select -1, 1".query[PointInQuadrant1].unique.transact(xa)
      .intercept[SecondaryValidationFailed[?]]
  }
```
In my opinion, it looks so much better than all this clutter with `secondaryValidationFailedCaught_?`, doesn't it?

modules/postgres/src/test/scala/doobie/postgres/ReadErrorSuite.scala
sharmaakshay177: Hi @satorg, Sounds good to me will apply this.

modules/refined/src/test/scala/doobie/refined/RefinedSuite.scala
sharmaakshay177: Sounds good. Added this check.

modules/refined/src/test/scala/doobie/refined/RefinedSuite.scala
sharmaakshay177: Resolved Both

modules/refined/src/test/scala/doobie/refined/RefinedSuite.scala
sharmaakshay177: Indeed it looks much better without this `secondaryValidationFailedCaught_?`

modules/refined/src/test/scala/doobie/refined/RefinedSuite.scala
```diff
@@ -54,15 +52,18 @@ class RefinedSuite extends munit.FunSuite {
   }
 
   test("Query should return a refined type when conversion is possible") {
-    sql"select 123".query[PositiveInt].unique.transact(xa).void.unsafeRunSync()
+    sql"select 123".query[PositiveInt].unique.transact(xa)
+      .assertEquals(Refined.unsafeApply[Int, Positive](123))
```
satorg, line 56: Ah, that's right: there's no macro support in Refined for Scala3, unfortunately.

Just in case, there's another form for create refined values in runtime that you may find less cumbersome:
```suggestion
      .assertEquals(refineV[Positive].unsafeFrom(123))
```
But that one you are using works too 👍

modules/postgres/src/test/scala/doobie/postgres/TypesSuite.scala
```diff
@@ -55,13 +54,16 @@ class TypesSuite extends munit.ScalaCheckSuite {
 
   def testInOut[A](col: String, a: A)(implicit m: Get[A], p: Put[A]) = {
     test(s"Mapping for $col as ${m.typeStack} - write+read $col as ${m.typeStack}") {
-      assertEquals(inOut(col, a).transact(xa).attempt.unsafeRunSync(), Right(a))
+      inOut(col, a).transact(xa).attempt
+        .assertEquals(Right(a))
```
satorg, line 58: Looks like the reason why `.attempt` was used here in the current version in the first place is that it was a literal translation from Specs2 tests.

Feel free to keep it this way here too, but just in case, I'd suppose that without `attempt` it would work here and below just as fine:
```suggestion
      inOut(col, a).transact(xa)
        .assertEquals(a)
```

satorg (approved):
Looks great, thanks!

satorg (approved):

PR #2225, opened 2025-03-20
Optimize compile time by removing use of anonymous classes for WriteComposite

ZDevelop94:
Addressing the issue https://github.com/typelevel/doobie/issues/1976

By adding a companion object for `Write.Composite[A]` with an instance method it should enable users of `Write.Composite[A]` to avoid use of the anonymous class and avoiding automatic derivation.

### DIFF
--- a/modules/core/src/main/scala/doobie/util/write.scala
+++ b/modules/core/src/main/scala/doobie/util/write.scala
@@ -91,7 +91,7 @@ object Write extends LowerPriority1Write {
     }
 
   implicit val unitWrite: Write[Unit] =
-    new Composite[Unit](Nil, _ => List.empty)
+    Composite.instance[Unit](Nil, _ => List.empty)
 
   /** Simple instance wrapping a Put. i.e. single column non-null value */
   class Single[A](put: Put[A]) extends Write[A] {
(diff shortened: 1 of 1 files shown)


modules/core/src/main/scala/doobie/util/write.scala
```diff
@@ -124,7 +124,8 @@ object Write extends LowerPriority1Write {
 
     override def toList(a: Option[A]): List[Any] = List(a)
 
-    override def toOpt: Write[Option[Option[A]]] = new Composite[Option[Option[A]]](List(this), x => List(x.flatten))
```
jatcwang, line 127: Oops. These are actually fine as they are just creating an instance of `Composite`. 

e.g.
```
class Foo(i: Int)

new Foo(1) // fine, creates an instance of foo

new Foo(2) {} // anonymous subclass of Foo

new Foo(1) {
  def abc(): Unit = ...
}  // less fine, this is an anonymous subclass
```

Some instances I can find in the codebase where anonymous subclasses are created are `Get/Put/Query/Query0/Update/Update0`. 
`Get/Put` shouldn't be too hard to fix, I'd leave the others as changing them is a bit more tricky.

ZDevelop94: Reverted

PR #2231, opened 2025-03-26
Mark `class Derived` as `final`

satorg:
Extendable value classes are suspicious. In most cases it is an oversight.
Extending value classes can lead to unexpected performance consequences.


satorg:
UPD.: on Scala3 a better alternative to value classes is opaque types. Since Read/Write derivation is completely separate on Scala2 versus Scala3, I decided to give it a shot and made `Derived` an opaque type on Scala3.

Let me know if works or there are concern/objections.

jatcwang (approved):

jatcwang:
Thanks!

Issue #1976, opened 2024-01-07
Optimize compile time by removing anonymous subclass


jatcwang:
https://oyvindberg.github.io/typo/blog/the-cost-of-implicits/#avoiding-anonymous-classes

Will be cool to verify the impact too. Here's his benchmark code
https://github.com/oyvindberg/typo/blob/53539fa8205327828958748dbd00b78e13bf6813/typo-scripts/src/scala/scripts/CompileBenchmark.scala

armanbilge:
> Will be cool to verify the impact too.

This will not only optimize compile time, but it will reduce the size of jars and potentially have some runtime performance benefits as well.

PR #2232, opened 2025-03-27
Rid of anonymous `Get`s and `Put`s

satorg:
A part of #1976.

### DIFF
--- a/modules/core/src/main/scala/doobie/util/write.scala
+++ b/modules/core/src/main/scala/doobie/util/write.scala
@@ -66,8 +66,10 @@ sealed trait Write[A] {
 
   def toFragment(a: A, sql: String = List.fill(length)("?").mkString(",")): Fragment = {
     val elems: List[Elem] = (puts zip toList(a)).map {
-      case ((p: Put[a], NoNulls), a)  => Elem.Arg(a.asInstanceOf[a], p)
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/doobie/util/write.scala
```diff
@@ -66,8 +66,10 @@ sealed trait Write[A] {
 
   def toFragment(a: A, sql: String = List.fill(length)("?").mkString(",")): Fragment = {
     val elems: List[Elem] = (puts zip toList(a)).map {
-      case ((p: Put[a], NoNulls), a)  => Elem.Arg(a.asInstanceOf[a], p)
-      case ((p: Put[a], Nullable), a) => Elem.Opt(a.asInstanceOf[Option[a]], p)
+      case ((p, nullab), x) => nullab match {
+          case NoNulls  => p match { case px: Put[x] => Elem.Arg(x.asInstanceOf[x], px) }
+          case Nullable => p match { case px: Put[x] => Elem.Opt(x.asInstanceOf[Option[x]], px) }
+        }
```
satorg, line 72: For some reason, concrete Put makes the original match dedected as non-exhaustive on both Scala2-s. Scala3 looks fine.

However, this change can be for good regardless: it makes the extraction of values out of the nested tuples occuring exactly once.

jatcwang:
Cheers!

PR #2242, opened 2025-04-05
Scala 3 derivation - Switch to Mirror-based derivation

### LINKED ISSUE #2214: Derivation of "large" case classes doesn't work in 1.0.0-RC8
Hi,

I noticed that case classes with more than 12 members don't compile.
Example:

```
//> using dep "org.tpolecat::doobie-core::1.0.0-RC8"

import doobie.*
import doobie.implicits.*

@main
def main() = {}

case class Test1(
  p1: String,
  p2: String,
  p3: String,
  p4: String,
  p5: String,
  p6: String,
  p7: String,
  p8: String,
  p9: String,
  p10: String,
  p11: String,
  p12: String,
  p13: String,
  p14: String,
  p15: String,
  p16: String,
  p17: String,
) derives Read
```
does not compile.

With 12 members the case class derivation works without any problem.
With 13 members I get the following warning:
```
Diagnostics:
1. Implicit search problem too large.
   an implicit search was terminated with failure after trying 50000 expressions.
   The root candidate for the search was:
   
     method derived in trait MkReadPlatform  for  doobie.util.Derived[doobie.util.MkRead[Test1]]}
   
   You can change the behavior by setting the `-Ximplicit-search-limit` value.
   Smaller values cause the search to fail faster.
   Larger values might make a very large search problem succeed.
    [168]
```

With more than 13 members, the compiler throws an error:

```
Diagnostics:
1. Cannot derive Read instance. Please check that each field in the case class has a Read instance or can derive one.
   I found:
   
       doobie.implicits.derived[Test1, A](
         Test1.$asInstanceOf[
           
             scala.deriving.Mirror.Product{
               type MirroredMonoType = Test1; type MirroredType = Test1;
                 type MirroredLabel = ("Test1" : String);
                 type MirroredElemTypes = (String, String, String, String, String,
                   String, String, String, String, String, String, String, String,
                   String);
                 type MirroredElemLabels = (("p1" : String), ("p2" : String),
                   ("p3" : String), ("p4" : String), ("p5" : String), ("p6" : String),
                   ("p7" : String), ("p8" : String), ("p9" : String),
                   ("p10" : String), ("p11" : String), ("p12" : String),
                   ("p13" : String), ("p14" : String))
             }
           
         ],
         <:<.refl[
           
             (String, String, String, String, String, String, String, String, String,
               String, String, String, String, String)
           
         ],
         /* missing */
           summon[
             doobie.util.shapeless.OrElse[
               doobie.util.Read[(String, String, String, String, String, String,
                 String, String, String, String, String, String, String, String)],
               doobie.util.Derived[
                 doobie.util.MkRead[(String, String, String, String, String, String,
                   String, String, String, String, String, String, String, String)]
               ]
             ]
           ]
         ,
       ???)
   
   But no implicit values were found that match type doobie.util.shapeless.OrElse[
     doobie.util.Read[(String, String, String, String, String, String, String,
       String, String, String, String, String, String, String)],
     doobie.util.Derived[
       doobie.util.MkRead[(String, String, String, String, String, String, String,
         String, String, String, String, String, String, String)]
     ]
   ].
   
   Note: method secondary in class OrElse0 was not considered because it was not imported with `import given`. [172]
```

Can you please check what the problem is?

Thank you so much
If I can help, please tell me how.

jatcwang:
Switch Scala 3 derivation to be based on macros instead of implicit prioritization (MkWrite etc). For Scala 3 this improves compile times significantly, as the `MkWrite/MkRead`-based derivation seems to be evaluated differently by Scala 3 ("Implicit search problem too large").

- Move MkWrite/MkRead to Scala 2, as they're no longer needed in Scala 3 derivation implementation
- Remove fromPutOption/fromGetOption for both Scala 2 and 3 as it's somewhat redundant whe we can get the same result via `Put[A] => Read[A]`, `Read[A] => Read[Option[A]]`.

Resolves #2214 #2241


modules/core/src/main/scala-2/doobie/util/WritePlatform.scala
```diff
@@ -4,10 +4,34 @@
...
+
+trait LowerPriority1Write extends LowerPriority2Write {
+
+  implicit def optionalFromWrite[A](implicit write: Write[A]): Write[Option[A]] =
+    write.toOpt
+}
+
+trait LowerPriority2Write extends LowerPriority3Write {
+  implicit def fromPut[A](implicit put: Put[A]): Write[A] =
+    new Write.Single(put)
```
jatcwang, line 30: @satorg I removed fromPutOption / fromGetOption :)

PR #2245, opened 2025-04-10
Add syntactic sugar transactRaw to ConnectionIOOps 

mertant:
`ConnectionIOOps` already has syntax for calling `.transact(xa)` on a query, which applies `xa.trans`, but there is no equivalent for `xa.rawTrans`.

See the `doobie.util.transactor.Transactor#rawTrans` method which already exists here:

[modules/core/src/main/scala/doobie/util/transactor.scala#L147-L158](https://github.com/typelevel/doobie/blob/23623fe78930c50b99e128e4c5d1c8afef1152d9/modules/core/src/main/scala/doobie/util/transactor.scala#L147-L158)

`xa.rawTrans` is useful when one does not want to use the `Strategy`, namely when we want to keep the `autoCommit` setting of the underlying DB connection instead of having Doobie override it to `false`, which it does by default. In other words, [as the Doobie documentation puts it](https://typelevel.org/doobie/docs/14-Managing-Connections.html#about-transactors), "This can be useful in cases **where transactional handling is unsupported or undesired**".

Here, we add `transactRaw` as a new syntax in `ConnectionIOOps`, paralleling the existing `transact`, for the sake of convenience.

See the discussion here:
- https://github.com/typelevel/doobie/issues/1244#issuecomment-2757961204

To keep feature parity with `ConnectionIOOps`, we also add `OptionTConnectionIOOps`, `EitherTConnectionIOOps`, `KleisliConnectionIOOps`) to keep their symmetry with `ConnectionIOOps`. However, note that these don't have unit tests. Also, could be made a bit more DRYDRY.

### DIFF
--- a/modules/core/src/test/scala/doobie/util/ConnectionIOSuite.scala
+++ b/modules/core/src/test/scala/doobie/util/ConnectionIOSuite.scala
@@ -24,11 +24,18 @@ class ConnectionIOSuite extends munit.CatsEffectSuite {
   test("Semigroup ConnectionIO") {
     val prg = Applicative[ConnectionIO].pure(List(1, 2, 3)) `combine` Applicative[ConnectionIO].pure(List(4, 5, 6))
     prg.transact(xa).assertEquals(List(1, 2, 3, 4, 5, 6))
+  }
+
+  test("Semigroup ConnectionIO raw") {
+    val prg = Applicative[ConnectionIO].pure(List(1, 2, 3)) `combine` Applicative[ConnectionIO].pure(List(4, 5, 6))
     prg.transactRaw(xa).assertEquals(List(1, 2, 3, 4, 5, 6))
   }
 
   test("Monoid ConnectionIO") {
     Monoid[ConnectionIO[List[Int]]].empty.transact(xa).assertEquals(Nil)
+  }
+
+  test("Monoid ConnectionIO raw") {
     Monoid[ConnectionIO[List[Int]]].empty.transactRaw(xa).assertEquals(Nil)
   }
 


satorg:
Just to clarify, using `transactRaw` would be somewhat equivalent to
```
xa.copy(strategy0 = Strategy.void).transact
```
is it correct?

mertant:
>would be somewhat equivalent to ... `Strategy.void`

Yes, I believe that's the intent - this PR does not add the implementation `rawTrans`, only a more convenient way to call it (just like we already have for the non-raw equivalent).

See the `doobie.util.transactor.Transactor#rawTrans` method which already exists here:
https://github.com/typelevel/doobie/blob/23623fe78930c50b99e128e4c5d1c8afef1152d9/modules/core/src/main/scala/doobie/util/transactor.scala#L147-L158

See also this documentation
https://github.com/typelevel/doobie/blob/23623fe78930c50b99e128e4c5d1c8afef1152d9/modules/docs/src/main/mdoc/docs/14-Managing-Connections.md?plain=1#L27-L30

mertant:
As for `xa.copy(strategy0 = Strategy.void)`, for performance reasons I would only execute the `.copy` once per transactor, not once for every `.transact` call, but yeah, I think it's basically just as you said.

With the lack of convenience for raw transactions currently, that is basically what we ended up doing in our project for now:
```
  /** A transactor with Doobie's default Strategy, which has a hidden gotcha:
    * it will set autoCommit=false if that's not already case on the Hikari connection from the pool!
    */
  val xa = ...

  /** A transactor with a "raw" Strategy, which is more like what one would expect, i.e.
    * it keeps autoCommit=true as we have set it on our Hikari pool!
    */
  val xaRaw = xa.copy(strategy0 = Strategy.void)
```

satorg:
Looks reasonable to me, thanks for the clarification.
Do you think it would make sense to add similar methods to `StreamOps`?

PR #2300, opened 2025-08-27
Add instancies for Array[Json] on Postgres

albertoadami:
Related to https://github.com/typelevel/doobie/issues/2299

Working on multiple projects using `JSONB[]` column, will not work out of the box with postgres-doobie.
In my opinion it will be helpful to have this functionality directly on the doobie library.

This pr adds contains the following changes:

- Add `Get[Array[Json]]` and `Put[Array[Json]]` in `JsonbInstances`
- Add tests to covered the new functionality.


jatcwang:
Thanks for the contribution @albertoadami! May I request that we test that it works for reading and writing as well? In the same test files there's `testInOut` to generate the test cases.

modules/postgres-circe/src/main/scala/doobie/postgres/circe/Instances.scala
```diff
@@ -34,6 +34,20 @@ object Instances {
       ).temap(a =>
         parse(a.getValue).leftMap(_.show))
 
+    implicit val arrayJsonGet: Get[Array[Json]] =
+      Get.Advanced.array[String](NonEmptyList.of("jsonb[]", "_jsonb"))
+        .map(_.map(jsonStr => parse(jsonStr).fold(throw _, identity)))
```
satorg, line 39: I'd suggest using `Ge.temap` instead of `map` and thereby avoid `throw` inside `map` context.

jatcwang (commented):
Thank you!

PR #2366, opened 2026-01-31
Update build.sbt: keep Scala 2.13 and Scala 3 in otel4s module

iRevive:
That's a bit of an oversight from my side. I forgot that doobie supports Scala 2.12, while otel4s is available only for Scala 2.13 and 3.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -219,5 +219,5 @@ jobs:
       - name: Submit Dependencies
         uses: scalacenter/sbt-dependency-submission@v2
         with:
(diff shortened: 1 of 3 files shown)


jatcwang:
Seems like we're still hitting some issues on the cross build side. It's a known sbt 1.x cross building issue.
`core` module supports 2.12, but `otel4s` module does not. However when we do `+update` (or `++ 2.12`) it'll still forcibly try to resolve otel4s by using 2.13 version of `core`, leading to the error.

I'll try to convert the project to sbt-projectmatrix

PR #2367, opened 2026-01-31
Setup sbt-projectmatrix

jatcwang:
Update FreeGen2 code generation for otel4s new `trace` methods

sbt-projectmatrix is needed because otel4s module only builds for 2.13/3, which won't
work in vanilla sbt cross building when we cross build against 2.12 too.

Also:
small test adjustment in TracedTransactorSuite


iRevive:
Would it make sense to start using https://github.com/typelevel/sbt-typelevel? I can take a look if that's feasible.

PR #2436, opened 2026-05-26
Adjust docs for noticeable fragment constructors according current real behavior

XriXis:
`or`, `and`, `where`, and their combinations wrap operands with parentheses, while doc contains old examples (from RC-5) with unwrapped operands. 

Made doc consistent with behavior (examples now show parentheses in constructors, where they will appear)


jatcwang:
Awesome thanks for the fix!

Issue #1728, opened 2022-08-08
Road to 1.0


jatcwang:
# RC3:
- [ ] Clear outstanding PRs
- [ ] Remove LogHandler, add label

# RC4:
Typelevel org migration
- [ ] Scalafix rule to help migrate packages 
- [ ] New docs location
- [ ] docs redirect

# Maybe?
- [ ] Typelevel SBT plugin

armanbilge:
> * Typelevel SBT plugin

The build is already using the plugin :)

https://github.com/tpolecat/doobie/blob/8a8ca1a40bb0709a300d6fdf65ad71331b078d7d/project/plugins.sbt#L4

george-wilson-rea:
Does #1798 get us closer to checking off "Clear outstanding PRs"?
I would really like a 1.0.0-RC3 release - is there anything I can do to help make that happen? The description for this issue mentions something about `LogHandler` - is there a separate issue describing that work?

armanbilge:
@george-wilson-rea note that snapshots are published for every merge to main. The latest is `1.0-b90576a-SNAPSHOT`, for example:

https://s01.oss.sonatype.org/content/repositories/snapshots/org/tpolecat/doobie-core_3/1.0-b90576a-SNAPSHOT/doobie-core_3-1.0-b90576a-SNAPSHOT.pom

jatcwang:
@george-wilson-rea I have created the issue describing my thoughts on what needs to be done around LogHandler/Query/Update.

Jasper-M:
Looks like the `HikariTransactor` doesn't have `LogHandlerM` support yet. And maybe there are others?

george-wilson-rea:
Could you please update this roadmap? I think most of these checkboxes can be checked. What is left before 1.0?

satorg:
I dare to suggest that this my PR could be useful for the migration to the sbt-typelevel plugin, because it has scalafmt enabled by default: #1989.

SuperFola:
Hello, thanks for the library, it's really useful!

I was wondering, what's blocking from releasing 1.0.0? Is the code package change to org.typelevel.doobie still in the works? It's been in RC land for a few years, it should be quite stable now right?

jatcwang:
Hi @SuperFola yes I've been chipping away at it! It's mostly done https://github.com/typelevel/doobie/pull/2423. Scalafix rule is done too which should help people migrate.

felix-hedenstrom:
[...]

PR #2480, opened 2026-07-20
Return Long from Update run/updateMany via executeLargeUpdate

### LINKED ISSUE #2479: Update run returns Int instead of Long
`Update#run` returns an Int which is a problem when more than 2**32 rows were affected.

https://github.com/typelevel/doobie/blob/19503f050d2f911942048148bc8da69b4e1fad14/modules/core/src/main/scala/doobie/util/update.scala#L85

It should be Long (using `executeLargeUpdate`).
This would be a breaking change, so ~~maybe we need `runLarge` method~~ it would be good to get it before 1.0 (https://github.com/typelevel/doobie/issues/1728).

mwisnicki:
## Summary

`Update#run`, `Update0#run`, `updateMany`, and their `*AlteringExecution` variants now return `ConnectionIO[Long]` instead of `ConnectionIO[Int]`, using JDBC's `executeLargeUpdate` / `executeLargeBatch`. This avoids overflow when more than 2^31 rows are affected.

Fixes #2479

## Details

- `modules/core/.../util/update.scala` — switched `executeUpdate` → `executeLargeUpdate` and `executeBatch` → `executeLargeBatch`; return types and the `PreparedExecutionWithoutProcessStep` type parameters changed from `Int` to `Long`. The free algebra already provided `executeLargeUpdate`/`executeLargeBatch`.
- Updated downstream call sites that hard-coded `Int` (tests, examples, and mdoc docs).

## Breaking change

This is a **source- and binary-breaking** change to the `Update`/`Update0` execution API. Per the issue discussion, targeting it for 1.0 (#1728) rather than adding a separate `runLarge` method.

## Testing

- `core/test` — all passing (H2 in-memory)
- `postgres/test` — all passing (exercises real `executeLargeUpdate`/`executeLargeBatch`)
- `mysql/test` — all passing

Generated with Claude Code

### DIFF
--- a/modules/otel4s/src/test/scala/doobie/otel4s/TracedTransactorSuite.scala
+++ b/modules/otel4s/src/test/scala/doobie/otel4s/TracedTransactorSuite.scala
@@ -170,18 +170,18 @@ class TracedTransactorSuite extends munit.CatsEffectSuite {
   testkitTest("record db.operation.batch.size for batch operations") { testkit =>
     val expected = expectedSpans(
       span(
-        name = "executeUpdate",
+        name = "executeLargeUpdate",
         attributes = Attributes(
           DbAttributes.DbQueryText("CREATE LOCAL TEMPORARY TABLE TEST_BATCH (int_value INT)"),
-          DbAttributes.DbOperationName("executeUpdate")
+          DbAttributes.DbOperationName("executeLargeUpdate")
         )
       ),
       span(
-        name = "executeBatch",
+        name = "executeLargeBatch",
         attributes = Attributes(
           DbAttributes.DbQueryText("insert into TEST_BATCH (int_value) values (?)"),
           DbAttributes.DbOperationBatchSize(3L),
-          DbAttributes.DbOperationName("executeBatch")
+          DbAttributes.DbOperationName("executeLargeBatch")
         )
       )
     )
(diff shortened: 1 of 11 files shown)


mwisnicki:
I've reviewed all the changes

modules/otel4s/src/test/scala/doobie/otel4s/TracedTransactorSuite.scala
```diff
@@ -170,18 +170,18 @@ class TracedTransactorSuite extends munit.CatsEffectSuite {
...
         )
       ),
       span(
-        name = "executeBatch",
+        name = "executeLargeBatch",
         attributes = Attributes(
           DbAttributes.DbQueryText("insert into TEST_BATCH (int_value) values (?)"),
           DbAttributes.DbOperationBatchSize(3L),
-          DbAttributes.DbOperationName("executeBatch")
+          DbAttributes.DbOperationName("executeLargeBatch")
```
mwisnicki, line 184: Unrelated to this PR but `db.operation.name` [should](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/) be a name of SQL operation like`INSERT` and not JDBC method name.

mwisnicki:
Also I expected MiMa to fail but looks like the MiMa step is not really checking [anything](https://github.com/typelevel/doobie/actions/runs/29718651147/job/88276745030?pr=2480#step:11:29)?

jatcwang:
@mwisnicki Yes MiMa is disabled until we reach 1.0 which is why there's not failure here

mwisnicki:
Is there anything preventing merging this PR?

